PR #6, opened 2023-08-24
apply policy check to txpool and block verification

### LINKED ISSUE #4: Apply policies in txpool
# Rationale

Policy contract defines the minimum allowed gas price and transaction blacklist, and they should be applied and executed by transaction pool.

# Implementation

The traditional txpool is now called Legacy Pool and Ethereum is going to have another subpool called Blob Pool for [EIP-4844](https://eips.ethereum.org/EIPS/eip-4844). 

They handle different types of transactions but share the same validation logic when we execute `pool.Add()`. The internal `validateTx` calls `txpool.ValidateTransactionWithState` to check whether a transaction is valid according to consensus rules.

The limit of minimum gas price and transaction blacklist will be applied here, the same as what Neo does.

- [x] Update method `ValidateTransactionWithState()`;
- [x] Add related methods to load policy address from `genesis.json`.

txhsl:
Close #4.

This PR is going to add a policy check to Geth transaction pool. Whenever there is a new transaction trying to be added into `legacypool` or `blobpool`, the method `ValidateTransactionWithState` will read a `minGasPrice` and a `blocked` status from current blockchain state database. The method will not use EVM but compute the index key and access the contract storage directly.

The `PolicyAddress` is going to be a field of `ChainConfig`, and be set to zero address by default. So if there is no policy contract or the field is missing in configuration, then the value of `minGasPrice` and `blocked` will be both zero and all transactions will pass the check.


ZhangTao1596 (approved):

nicolegys:
Is `minGasPrice` an addtion to the existing tx gas price check mechanism? The gasprice must be high enough to cover the requirement of the calling pool and block producer, and also not less than `minGasPrice`?

txhsl:
> Is `minGasPrice` an addtion to the existing tx gas price check mechanism? The gasprice must be high enough to cover the requirement of the calling pool and block producer, and also not less than `minGasPrice`?

Yes, it is a policy of the lowest gas price that a transaction sender should provide. And this policy will be added to block verification as well in the future, when Bane's DBFT is built up.

core/txpool/validation.go
```diff
@@ -207,6 +210,23 @@ func ValidateTransactionWithState(tx *types.Transaction, signer types.Signer, op
...
+	// Apply policy minimum gas price
+	var minGasPrice = opts.State.GetState(opts.PolicyAddress, common.BytesToHash([]byte{0}))
+	if tx.GasPrice().Cmp(minGasPrice.Big()) < 0 {
+		return fmt.Errorf("%w: policy needed %v, tx permitted %v", ErrUnderpriced, minGasPrice.Big(), tx.GasPrice())
+	}
+	// Apply policy blacklist
+	var blocked = opts.State.GetState(opts.PolicyAddress, crypto.Keccak256Hash(common.LeftPadBytes(from.Bytes(), 32), common.LeftPadBytes([]byte{1}, 32)))
+	if blocked != (common.Hash{}) {
+		return fmt.Errorf("%w: sender %s", ErrBlockedSender, from.Hex())
+	}
```
nicolegys:
```suggestion
	// then minGasPrice is slot 1, blacklist is slot 2
	// Apply policy minimum gas price
	var minGasPrice = opts.State.GetState(opts.PolicyAddress, common.BytesToHash([]byte{1}))
	if tx.GasPrice().Cmp(minGasPrice.Big()) < 0 {
		return fmt.Errorf("%w: policy needed %v, tx permitted %v", ErrUnderpriced, minGasPrice.Big(), tx.GasPrice())
	}
	// Apply policy blacklist
	var blocked = opts.State.GetState(opts.PolicyAddress, crypto.Keccak256Hash(common.LeftPadBytes(from.Bytes(), 32), common.LeftPadBytes([]byte{2}, 32)))
	if blocked != (common.Hash{}) {
		return fmt.Errorf("%w: sender %s", ErrBlockedSender, from.Hex())
	}
```
`Policy` contract inherits from `GovernanceVote` contract, slot 0 is `voteMap`.

nicolegys:
merge?

AnnaShaleva:
@txhsl, we have a dedicated issue for PrepareRequest verification, could you please write a couple of words about minGasPrice requirement to https://github.com/bane-labs/go-ethereum/issues/53? This constraint then will likely be implemented as a part of PrepareRequest verification. Or at list it won't hang in the air and we'll transfer it to a separate issue if needed.

nicolegys:
Can we move on?

Now we have,
1. `base fee` and `priority fee` limit by orginal Geth source code, which `priority fee` is set by every CN node and now it will be sent to `coinbase`.
2. `minGasPrice` limit by policy contract.

What the newest fee rules?
https://github.com/bane-labs/go-ethereum/issues/53#issuecomment-1863774265

txhsl:
> Can we move on?
> 
> Now we have,
> 
> 1. `base fee` and `priority fee` limit by orginal Geth source code, which `priority fee` is set by every CN node and now it will be sent to `coinbase`.
> 2. `minGasPrice` limit by policy contract.
> 
> What the newest fee rules? [#53 (comment)](https://github.com/bane-labs/go-ethereum/issues/53#issuecomment-1863774265)

As far as I remember, we may going to add the `base fee` configuration into Policy contract, this PR will update after related contract is updated.

I'm free to do it now, but some research is necessary as well, especially when the `priority fee` is set by every CN but equally shared to everyone. It's different from Geth.

AnnaShaleva:
I'd like to review this PR as far, but let's agree on the final version of #7 firstly and merge it. After this we can rebase this PR onto #7.

txhsl:
Sure.

params/config.go
```diff
@@ -290,7 +298,8 @@ var NetworkNames = map[string]string{
 // that any network, identified by its genesis block, can have its own
 // set of configuration options.
 type ChainConfig struct {
-	ChainID *big.Int `json:"chainId"` // chainId identifies the current chain and is used for replay protection
+	ChainID       *big.Int       `json:"chainId"`                 // chainId identifies the current chain and is used for replay protection
+	PolicyAddress common.Address `json:"policyAddress,omitempty"` // Policy contract address
```
AnnaShaleva: It seems that Policy hash (proxy) is going to be fixed for all networks. And thus, we may add this hash directly to `systemcontracts` package: https://github.com/bane-labs/go-ethereum/blob/0aed4be8286331e4ce7965c7203a3dcc3279182b/core/systemcontracts/contracts.go#L10

core/txpool/validation.go
```diff
@@ -207,6 +210,17 @@ func ValidateTransactionWithState(tx *types.Transaction, signer types.Signer, op
...
 		return fmt.Errorf("%w: next nonce %v, tx nonce %v", core.ErrNonceTooLow, next, tx.Nonce())
 	}
+	// Ensure the transaction is allowed by policy
+	// Apply policy minimum gas price
+	var minGasPrice = opts.State.GetState(opts.PolicyAddress, common.BytesToHash([]byte{1}))
+	if tx.GasPrice().Cmp(minGasPrice.Big()) < 0 {
+		return fmt.Errorf("%w: policy needed %v, tx permitted %v", ErrUnderpriced, minGasPrice.Big(), tx.GasPrice())
+	}
+	// Apply policy blacklist
+	var blocked = opts.State.GetState(opts.PolicyAddress, crypto.Keccak256Hash(common.LeftPadBytes(from.Bytes(), 32), common.LeftPadBytes([]byte{2}, 32)))
```
AnnaShaleva: I'm not sure it's a good way to directly get storage items from state. Have you consider the "honest" test invocation via Policy API? Something like https://github.com/bane-labs/go-ethereum/blob/0aed4be8286331e4ce7965c7203a3dcc3279182b/consensus/dbft/dbft.go#L1608

core/txpool/validation.go
```diff
@@ -207,6 +210,17 @@ func ValidateTransactionWithState(tx *types.Transaction, signer types.Signer, op
...
+	if blocked != (common.Hash{}) {
+		return fmt.Errorf("%w: sender %s", ErrBlockedSender, from.Hex())
+	}
```
AnnaShaleva, line 214: I think that this logic should be added to state transition validation (the code that is responsible for tx processing during incoming block verification). It's good that we have it in pool, but it's also necessary to have it for incoming blocks verification.

AnnaShaleva (commented):
Let's rebase it to the fresh master.

core/txpool/validation.go
txhsl: Sure.

core/txpool/validation.go
txhsl: I know this way may look better, but here in transaction validation, we are using `ValidationOptionsWithState` to validate tx with statedb directly.

And also, I'm a bit worried about the code performance when request increases.

AnnaShaleva: Well, can we at least move `common.LeftPadBytes([]byte{2}, 32)` or `[]byte{2}` to a separate constant/variable with some meaningful name? We need to keep it in sync with the Policy contract implementation.

core/txpool/validation.go
txhsl: Done.

core/txpool/validation.go
txhsl: I've moved them to 

https://github.com/bane-labs/go-ethereum/blob/a326b1ef0775518c7d6ba9ef9cc1e30eb395103e/core/systemcontracts/contracts.go#L19-L20

and provide two methods to get the hash 

https://github.com/bane-labs/go-ethereum/blob/a326b1ef0775518c7d6ba9ef9cc1e30eb395103e/core/systemcontracts/contracts.go#L47-L53

txhsl:
@chenquanyu Need your review about policy check in block verification, related to #138.

core/state_transition.go
```diff
@@ -293,6 +294,19 @@ func (st *StateTransition) preCheck() error {
...
+		if msg.GasPrice.Cmp(minGasPrice.Big()) < 0 {
+			return fmt.Errorf("%w: address %v, gasprice %v", ErrUnderpriced,
+				msg.From.Hex(), msg.GasPrice)
+		}
+		// Apply policy blacklist
+		var blocked = st.state.GetState(systemcontracts.PolicyProxyHash, systemcontracts.GetBlackListStateHash(msg.From))
+		if blocked != (common.Hash{}) {
+			return fmt.Errorf("%w: address %v", ErrBlockedSender,
+				msg.From.Hex())
+		}
```
AnnaShaleva: Good, it's exactly the place where this code should be.

However, I think we need to come to an agreement about the following: unlike in NeoN3, here we check blocked accounts *after previous transaction is executed* (in N3 we check it once for all in-block transactions before the block processing). And this new way LGTM, we react quickly on newly-blocked accounts. The reason I started this conversation is in `minGasPrice`. In N3 policy changes like BaseExecFee factor or storage FeePerByte start to be applied from the next block (and technically it's correct, because users construct new transactions based on the previous block state). And here if we increase `minGasPrice` in the middle of the block processing, then subsequent transactions may fail due to underpriced GAS whereas these transactions were valid in the start of the block processing.

That's why I'd suggest to capture `minGasPrice` in the start of block execution and then reuse this value within all in-block transactions processing.

AnnaShaleva: @roman-khimov, need you thoughts on this.

core/systemcontracts/contracts.go
```diff
@@ -26,4 +38,16 @@ func init() {
...
+		panic(fmt.Errorf("failed to decode Policy contract ABI: %w", err))
+	}
+}
+
+func GetMinGasPriceStateHash() common.Hash {
+	return common.BytesToHash([]byte{minGasPriceSlotIndex})
+}
+
+func GetBlackListStateHash(addr common.Address) common.Hash {
+	return crypto.Keccak256Hash(common.LeftPadBytes(addr.Bytes(), 32), common.LeftPadBytes([]byte{blackListSlotIndex}, 32))
```
AnnaShaleva: Exported methods should have comments.

core/state_transition.go
roman-khimov: These parameters don't change often, so it's not a huge problem. At the same time you're right in that we can solve it this way, shouldn't be very hard to do that.

AnnaShaleva: >  shouldn't be very hard to do that.

I've reviewed the existing code, so there are a lot of `ApplyMessage` usages and some of them are tricky enough to change. I'm afraid we may destabilize node by these changes.

>These parameters don't change often, so it's not a huge problem.

So let's go with the current solution, since this policy setting is really not a hot path.

AnnaShaleva (commented):
The code itself LGTM, let's squash the commits since it's a separate logical block of changes and wait for #7 to be merged. We need to test this PR against the updated master after #7 merge.

chenquanyu (approved):

core/txpool/validation.go
```diff
@@ -200,6 +201,17 @@ func ValidateTransactionWithState(tx *types.Transaction, signer types.Signer, op
 	if next > tx.Nonce() {
 		return fmt.Errorf("%w: next nonce %v, tx nonce %v", core.ErrNonceTooLow, next, tx.Nonce())
 	}
+	// Ensure the transaction is allowed by policy
+	// Apply policy minimum gas price
+	var minGasPrice = opts.State.GetState(systemcontracts.PolicyProxyHash, systemcontracts.GetMinGasPriceStateHash())
+	if tx.GasPrice().Cmp(minGasPrice.Big()) < 0 {
+		return fmt.Errorf("%w: policy needed %v, tx permitted %v", ErrUnderpriced, minGasPrice.Big(), tx.GasPrice())
+	}
```
AnnaShaleva: After some [digging in this direction](https://github.com/bane-labs/go-ethereum/issues/137#issuecomment-2046960695) I have two issues:
1. Are you sure that it's correct to use `tx.GasPrice()`? What is the intention of Policy's minGASPrice restriction? I'm asking because txpool state-independent check includes verification against `tx.GasTipCap()`, and there are transaction types for those this value differs from `tx.GasPrice()` (e.g. dynamic fee transactions):
https://github.com/bane-labs/go-ethereum/blob/6cd26f231675ad8c229ce4fc8644d2f3d29c3d6a/core/txpool/validation.go#L106-L110
2. This value of Policy contract may conflict with miner's minGasPrce setting configured via `--miner.gasprice` or via `miner.SetGasPrice` API. So I'm not sure whether should we do something to unify these two settings.

txhsl: For point 1, dynamic fee transactions imported two new fields `GasFeeCap` and `GasTipCap`, and mixed the usage of method `gasPrice()`. While a legacy transaction has the field `GasPrice`, a dynamic fee transaction leaves it as empty when being sent, but the field will be replaced when being added to mempool. 

You may refer to their definitions [LegacyTx](https://github.com/bane-labs/go-ethereum/blob/policy-txpool/core/types/tx_legacy.go) and [DynamicFeeTx](https://github.com/bane-labs/go-ethereum/blob/policy-txpool/core/types/tx_dynamic_fee.go), and have a look at the implementation of `gasPrice()`, they return different things. The `gasPrice()` of LegacyTx is `GasPrice`, but the `gasPrice()` of DynamicFeeTx is `GasFeeCap`. But anyway, it is not the `GasTipCap`.

The intention of Policy's minGASPrice restriction is, well, simply just giving governance something to config, at the beginning of our NGD's discussion. But I think it can prevent `gasprice`, or I mean the `basefee` to drop to nearly 0 when the blocks are always empty, as it is defined in [EIP-1559](https://github.com/bane-labs/go-ethereum/blob/policy-txpool/consensus/misc/eip1559/eip1559.go).

Also, for point 2, I think they are not conflict. Because the `--miner.gasprice` in Ethereum is the minimum `gasprice` that a miner wants to accept, it gives a setting for the miner to grab more benefit when packing. But the `minGASPrice` of Policy is a setting for the network to refuse transactions with a low gas price.

When `--miner.gasprice` is higher than `minGASPrice`, nothing happens, but when the former is lower, then some transactions may get refused by policy. This setting will keep the fee above a specific level, and affect the fee history to help users of e.g. Metamask to send "good" transactions.

AnnaShaleva: A clear explanation, good to go.

eth/api_backend.go
```diff
@@ -363,7 +365,21 @@ func (b *EthAPIBackend) SyncProgress() ethereum.SyncProgress {
...
+	}
+	header := b.CurrentHeader()
+	root := types.EmptyRootHash
+	if header != nil {
+		root = header.Root
+	}
+	stateDb, err := b.eth.BlockChain().StateAt(root)
+	if err != nil {
+		return nil, err
+	}
```
AnnaShaleva: Just use `StateAndHeaderByNumber` with `rpc.LatestBlockNumber`.

core/txpool/validation.go
```diff
@@ -200,6 +201,17 @@ func ValidateTransactionWithState(tx *types.Transaction, signer types.Signer, op
 	if next > tx.Nonce() {
 		return fmt.Errorf("%w: next nonce %v, tx nonce %v", core.ErrNonceTooLow, next, tx.Nonce())
 	}
+	// Ensure the transaction is allowed by policy
+	// Apply policy minimum gas tip cap
+	var minGasTipCap = opts.State.GetState(systemcontracts.PolicyProxyHash, systemcontracts.GetMinGasTipCapStateHash())
+	if tx.GasTipCap().Cmp(minGasTipCap.Big()) < 0 {
+		return fmt.Errorf("%w: policy needed %v, tip permitted %v", ErrUnderpriced, minGasTipCap.Big(), tx.GasTipCap())
+	}
```
nicolegys, line 209: So to pass the policy check, for dynamicTx should be `maxPriorityFeePerGas >= policy.minGasTipCap`, for legacyTx should be `gasPrice >= policy.minGasTipCap` ?

eth/api_backend.go
```diff
@@ -363,7 +365,16 @@ func (b *EthAPIBackend) SyncProgress() ethereum.SyncProgress {
...
+	suggestTipCap, err := b.gpo.SuggestTipCap(ctx)
+	if err != nil {
+		return nil, err
+	}
+	stateDb, err := b.eth.BlockChain().StateAt(b.CurrentHeader().Root)
+	if err != nil {
+		return nil, err
+	}
+	minGasTipCap := stateDb.GetState(systemcontracts.PolicyProxyHash, systemcontracts.GetMinGasTipCapStateHash()).Big()
+	return cmath.BigMax(suggestTipCap, minGasTipCap), nil
```
nicolegys: I know it's necessary but this change makes the value of `eth_gasPrice` inaccurate, e.g.
When baseFee is `7 wei`, miner.gasPrice is `1 Gwei`(just like in T2), policy_minGasTipCap is `1.000000005 Gwei`, `eth_gasPrice` will return `policy_minGasTipCap+baseFee` which is `1.000000012` Gwei as suggested gasPrice, while `1.000000007 Gwei` is enough.
The root cause is, miners still rely on their own miner_gasPrice setting and network baseFee to decide whether to package a tx.

txhsl: The policy check in mempool and consensus will filter them out, but yes, I'm worried about the capability with `eth_feehistory` and related tools e.g. Metamask. In fact, it seems not working well.

@chenquanyu is going to set fixed `basefee` for burning, then only `gastip` can be shared as governance reward. If we don't use an "inaccurate" price, then governance reward can be always 0 with empty blocks.

core/txpool/validation.go
txhsl: > So to pass the policy check, for dynamicTx should be `maxPriorityFeePerGas >= policy.minGasTipCap`, for legacyTx should be `gasPrice >= policy.minGasTipCap` ?

Yes, the same handling as `--miner.gasprice` thing.

eth/api_backend.go
nicolegys: > @chenquanyu is going to set fixed basefee for burning

0.0 What do you mean by `set fixed basefee for burning`? Burning `block.basefee` like ethereum or burning a fixed value?

> I'm worried about the capability with eth_feehistory and related tools e.g. Metamask.

What's wrong with `eth_feehistory`? I see it just returns `baseFeePerGas` and `gasUsedRatio`,  and seems OK.

> If we don't use an "inaccurate" price, then governance reward can be always 0 with empty blocks.

Why? Empty blocks means no tx, no tx so no governance reward, it's OK. If you mean that empty blocks makes baseFee drop to the lowest value nearly 0, we still have miner settings to ensure governance reward.

But anyway, policy check prevent malicious txs with low fees from entering the pool. So by the way, I wonder if we still need `--txpool.nolocals` for t3 seeds, because node doesn't store transactions.rlp with this flag, txs in the txpool will be lost if the node restarts.

txhsl: > Burning `block.basefee` like ethereum or burning a fixed value?

Still burn `block.basefee`, but maybe not based on the same rule.

> What's wrong with `eth_feehistory`?

`eth_feehistory` works well. But I tested it with Metamask, which set 1.5 Gwei as default when 5 blocks in `feehistory` are all empty, so when I improved `minGasTipCap` above that, this default setting got RPC errors.

eth/api_backend.go
txhsl: Cool.

eth/api_backend.go
nicolegys: > `eth_feehistory` works well. But I tested it with Metamask, which set 1.5 Gwei as default when 5 blocks in `feehistory` are all empty, so when I improved `minGasTipCap` above that, this default setting got RPC errors.

What RPC error have you encountered?


I can set the feeCap and tipCap to the lowest value, and the tx succeed.



Even can set them lower than the network required. Although Metamask shows warning, tx can be sent to node txpool successfully, and whether it's packaged or not is a matter of miner.

I think the issue is that Metamask has its own way of calculating fees, users can't get appropriate fee value.

txhsl: The `minGasPrice` in #7 genesis file is extremely low, try to change [this line](https://github.com/bane-labs/go-ethereum/blob/75bc45d241dfe9c6843fb5a6337edad127a196f7/config/genesis_testnet.json#L72) and you may get RPC error.

AnnaShaleva (approved):

eth/api_backend.go
AnnaShaleva: Ref. #167.

core/state_transition.go
```diff
@@ -293,6 +294,19 @@ func (st *StateTransition) preCheck() error {
 			return fmt.Errorf("%w: address %v, codehash: %s", ErrSenderNoEOA,
 				msg.From.Hex(), codeHash)
 		}
+		// Ensure the transaction is allowed by policy
+		// Apply policy minimum gas tip cap
+		var minGasTipCap = st.state.GetState(systemcontracts.PolicyProxyHash, systemcontracts.GetMinGasTipCapStateHash())
```
dusmart, line 299: I have a question here. As this will fetch the latest state of the chain. Does this affect transactions that follows after a `MinGasTip`-changing tx that appeared in the same block?

txhsl: It will, so I think you mean, it will lead to a case that some transactions accepted by txpool will be rejected by consensus.

txhsl: Check #166 if necessary, there is the latest code.

PR #7, opened 2023-09-01
Add governance and policy contracts

### LINKED ISSUE #2: Governance contract
# Rationale

We need a predeployed governance contract to manage the consensus nodes list.

# Implementation

We will maintain a Phase list to record consensus changes, each Phase has startHeight and the previous Phase's start height.
We can find any Phase with a give block height.
Consensus node account can propose a draft to change current consensus nodes list.
Users can get draft list and vote to support the specified draft with unique draftId.

The current interface of Governance contract:

```
interface IGovernance {
    struct Draft {
        uint id;
        uint startHeight;
        address[] miners;
    }

    struct Phase {
        uint startHeight;
        address[] miners;
        uint preHeight;
    }

    // propose draft, contains start height and consensus list
    function propose(
        uint startHeight,
        address[] memory miners
    ) external returns (bool);

    // get Draft list, contians unique id
    function getDraftList() external view returns (Draft[] memory);

    // vote draft with gas
    // when the vote condition meets: 1. convert draft to phase; 2.clean draft list
    function vote(uint256 draftId) external returns (bool);

    // get current consensus phase
    function getCurrentPhase() external view returns (Phase memory);

    // get consensus phase by start height
    function getPhaseByHeight(uint height) external view returns (Phase memory);
}

```

### LINKED ISSUE #3: Policy contract
# Rationale

Policy contract is a configuration contract that manages several global policies for every node, e.g. minimum allowed gas price, transaction blacklist. Policy contract can be only written by the consensus, but should be read and applied in runtime by everyone.

# Implementation

Policy contract is going to be a pre-deployed Solidity smart contract, which has the following functionality.
- [x] Declare the lowest gas price for a transaction to be allowed in transaction pool;
- [x] Declare a blacklist for some addresses that are not allowed in transaction pool;
- [x] Check and only allow the consensus (governance) to update the configurations and upgrade the contract.

### LINKED ISSUE #5: Deploy Governance contract and Policy contract in genesis block
# Rationale

We are going to develop Governance contract and Policy contract as Solidity contracts instead of precompile contracts, but they are both related with node executions (e.g. consensus and txpool). So these two contract are supposed to be deployed in genesis block.

# Implementation

- [x] Pre deploy these two contracts with a configuration file.

chenquanyu:
This PR is a code implementation of #2, #3, #5. Close #2, Close #3, Close #5. 

This PR is going to add the `Governance` contract to manage consensus list, `Policy` contract to manage policy parameters. We will have the first Phase of consensus in genesis block and the users can vote for new Phases with GAS in `Governance` contract. The `GovReward` contract will keep the reward GAS for governance voters. For governance requirements, please refer to https://docs.google.com/document/d/11iDJGpCDNrKmPQN8TOhmEqBlWZAOstH15j3Afc2xoQo/edit#heading=h.idvp9uir3xt8.

The `GovernanceVote` contract is an abstract contract for Policy and other contracts (like `GovProxyAdmin`) who need a 1/2 consensus vote to execute some function.

We will use upgradable contracts for Governance, Policy and GovReward, `ERC1967Proxy` is used as proxy and `GovProxyAdmin` is used as the admin of upgrade.

For the implemention of #5, we will need following contracts in genesis alloc: 
1. GovProxyAdmin 0x1212000000000000000000000000000000000000
2. Governance Proxy 0x1212000000000000000000000000000000000001
3. Governance Implemention 0x1212100000000000000000000000000000000001
4. Policy Proxy 0x1212000000000000000000000000000000000002
5. Policy implemention 0x1212100000000000000000000000000000000002
6. Governance Reward 0x1212000000000000000000000000000000000003
7. Governance Reward implemention 0x1212100000000000000000000000000000000003


txhsl (approved):

contracts/solidity/Governance.sol
```diff
@@ -0,0 +1,226 @@
...
+        draftVoteMap[msg.sender] = draftId;
+        draftVoteList[draftId].add(msg.sender);
+        emit Vote(msg.sender, draftId);
+
+        // check vote balance
+        uint votedBalance = getVoteBalance(draftId);
+        if (votedBalance >= VOTE_TARGET_AMOUNT) {
+            Draft memory draft = draftMap[draftId];
+            Phase memory phase = Phase({
+                startHeight: draft.startHeight,
```
ZhangTao1596: Check draft.startHeight > block.number here?

contracts/solidity/Policy.sol
```diff
@@ -0,0 +1,43 @@
...
+
+contract Policy is GovernanceVote{
+    using EnumerableSet for EnumerableSet.AddressSet;   
+    uint public  minGasPrice ;
+    mapping(address=>bool) public isBlackListed;
+    EnumerableSet.AddressSet  blackList;
+    bool private initialized;
+   
+
+    function initialize()public{
```
ZhangTao1596: format

contracts/solidity/Policy.sol
```diff
@@ -0,0 +1,43 @@
...
+        require(!initialized, "Contract instance has already been initialized");
+        minGasPrice = 10000000;       
+        initialized = true;
+	}
```
ZhangTao1596: format

contracts/solidity/Policy.sol
```diff
@@ -0,0 +1,43 @@
...
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.9;
+import "@openzeppelin/contracts/utils/structs/EnumerableSet.sol";
+import "./GovernanceVote.sol";
+
+contract Policy is GovernanceVote{
+    using EnumerableSet for EnumerableSet.AddressSet;   
+    uint public  minGasPrice ;
+    mapping(address=>bool) public isBlackListed;
+    EnumerableSet.AddressSet  blackList;
```
txhsl: Do we need the `blackList` and `getBlackList()` since we already have the map `isBlackListed`? I don't find them in Neo's Policy contract.

Tobeyw: #16

contracts/solidity/Policy.sol
Tobeyw: #17

contracts/solidity/Policy.sol
chenquanyu, line 1: add events for policy contract

contracts/solidity/Governance.sol
chenquanyu: fixed in commit a6fdfc5767bdc56ca421fd83c26f8f1df607b484

cloud8little (commented):
https://github.com/bane-labs/go-ethereum/pull/7/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R36
GOBIN -> GETHBIN

genesis.json
```diff
@@ -0,0 +1,62 @@
...
...05b838110156112d85780820151818401526020810190506112bd565b60008484015250505050565b60006112ef8261129e565b6112f981856112a9565b93506113098185602086016112ba565b61131281610cc9565b840191505092915050565b600060608201905061133260008301866111e9565b61133f6020830185610b4c565b818103604083015261135181846112e4565b9050949350505050565b60006040820190506113706000830185610b4c565b818103602083015261138281846112e4565b90509392505050565b7ff851a44000000000000000000000000000000000000000000000000000000000600082015250565b60006113c1600483610e74565b91506113cc8261138b565b600482019050919050565b60006113e2826113b4565b9150819050919050565b6000819050919050565b6113ff816113ec565b82525050565b600060608201905061141a6000830186610b4c565b61142760208301856113f6565b61143460408301846113f6565b949350505050565b600061144782610f55565b915061145283610f55565b925082820261146081610f55565b9150828204841483151761147757611476611160565b5b5092915050565b7f4e487b7100000000000000000000000000000000000000000000000000000000600052601260045260246000fd5b60006114b882610f55565b91506114c383610f55565b9250826114d3576114d261147e565b5b82820490509291505056fea2646970667358221220bccdf24785bb5bdc3329d771b90100daba80b99f240bdedcb7b4f20228fa226664736f6c63430008120033"
```
nicolegys: The `code` should be runtime bytecode.

nicolegys (changes requested):

contracts/solidity/GovProxyAdmin.sol
```diff
@@ -0,0 +1,116 @@
...
+     *
+     * - This contract must be the current admin of `proxy`.
+     */
+    function changeProxyAdmin(
+        ITransparentUpgradeableProxy proxy,
+        address newAdmin
+    )
+        public
+        virtual
+        needVote(
```
nicolegys: `{GovProxyAdmin}` is the admin of `{Governance Proxy}`, and it cannot fallback to `{Governance Proxy}`. However, `needVote` calls `{Governance.getCurrentPhase}`, so the execution will be reverted.

chenquanyu: Yes, ProxyAdmin can not call fallback methods in Proxy. I will add a middle contract(GovReward.sol) to read governance voters.

contracts/solidity/Governance.sol
```diff
@@ -0,0 +1,220 @@
...
+            id: endDraftId,
+            miners: miners,
+            startHeight: startHeight
+        });
+        emit Propose(msg.sender, endDraftId, startHeight, miners);
+    }
+
+    function getDraftList() external view override returns (Draft[] memory) {
+        uint256 count = endDraftId - startDraftId + 1;
+        Draft[] memory drafts = new Draft[](count);
```
nicolegys: Call to `getDraftList` errored when there was no draft
image
Because in the initial state, startDraftId=1, endDraftId=0

chenquanyu: Fixed by changing the order, please help to check the new logic of voting too.

contracts/solidity/GovProxyAdmin.sol
chenquanyu: > `{GovProxyAdmin}` is the admin of `{Governance Proxy}`, and it cannot fallback to `{Governance Proxy}`. However, `needVote` calls `{Governance.getCurrentPhase}`, so the execution will be reverted.
Please verify this with new genesis.json file.

contracts/solidity/Governance.sol
```diff
@@ -0,0 +1,318 @@
...
+        draftValueMap[draftId] -= voteValue;
+        // only update rewardShareMap in current phase
+        if (draftId >= startDraftId) {
+            delete rewardShareMap[draftId][msg.sender];
+        }
+        safeTransferETH(msg.sender, voteValue);
+        emit RevokeVote(msg.sender, draftId, voteValue);
+    }
+
+    function getRewardAmount(
```
nicolegys: The contract updates `lastStartHeight` and `rewardRecord` when vote passes, not when new phase starts, and there is a transition time between vote-pass and phase-start. 
`getRewardAmount` function returns different results when a voter calls it before and after a new draft passes, the latter amount is smaller.
So if someone had withdrawed his reward at `phase N - no draft passes`, then tries to withdraw again at `phase N - new draft passes`, there will be an underflow error.

chenquanyu: Fixed by changing the current phase to last phase. 
`Phase memory current = phaseMap[lastStartHeight]`

nicolegys:
From the user's point of view, there is no function for users to get their voted draftIds. If someone has voted for several proposals in the past phases, he must walk through the previous transactions to find the draftIds and then get his ethers back (Unless he wrote them down somewhere by himself).

contracts/solidity/Governance.sol
nicolegys: > Fixed by changing the current phase to last phase. `Phase memory current = phaseMap[lastStartHeight]`

I'll test it again.

contracts/solidity/Governance.sol
```diff
@@ -0,0 +1,318 @@
...
+        reward -= withdrawedReward[addr];
+        return reward;
+    }
+
+    function withdrawReward() external override nonReentrant {
+        uint reward = getRewardAmount(msg.sender);
+        if (reward > 0) {
+            withdrawedReward[msg.sender] += reward;
+            totalWithdrawedReward += reward;
+            IGovReward(govReward).withdraw(msg.sender, reward);
```
nicolegys:
```suggestion
            IGovReward(govReward).withdraw(msg.sender, reward);
            emit WithdrawReward(msg.sender, reward);
```

privnet/node2/genesis_privnet.json
```diff
@@ -23,7 +23,51 @@
...
+      "storage": {
+        "0x0000000000000000000000000000000000000000000000000000000000000000": "0x989680",
```
nicolegys:
```suggestion
        "0x0000000000000000000000000000000000000000000000000000000000000001": "0x989680",
```

genesis.json
```diff
@@ -0,0 +1,67 @@
...
+      "storage": {
+        "0x0000000000000000000000000000000000000000000000000000000000000000": "0x989680",
```
nicolegys:
```suggestion
        "0x0000000000000000000000000000000000000000000000000000000000000001": "0x989680",
```

privnet/node1/genesis_privnet.json
```diff
@@ -23,7 +23,51 @@
...
+      "storage": {
+        "0x0000000000000000000000000000000000000000000000000000000000000000": "0x989680",
```
nicolegys:
```suggestion
        "0x0000000000000000000000000000000000000000000000000000000000000001": "0x989680",
```

nicolegys:
merge?

chenquanyu:
> From the user's point of view, there is no function for users to get their voted draftIds. If someone has voted for several proposals in the past phases, he must walk through the previous transactions to find the draftIds and then get his ethers back (Unless he wrote them down somewhere by himself).

May be will need a UI for users to vote, and will list all the voted draft from event log.

Makefile
```diff
@@ -15,22 +15,27 @@ FOUR_DIR = $(MAIN_DIR)/four
 NODE1 = node1
 NODE1_PORT = 30306
 NODE1_RPC_PORT = 8552
+NODE1_HTTP_PORT = 8545
```
AnnaShaleva: Can't we reuse `NODEX_RPC_PORT` variable to set `--http.port`? This Makefile is already complicated.

contracts/solidity/Governance.sol
```diff
@@ -0,0 +1,319 @@
...
+        _status = _ENTERED;
+
+        _;
+
+        // By storing the original value once again, a refund is triggered (see
+        // https://eips.ethereum.org/EIPS/eip-2200)
+        _status = _NOT_ENTERED;
+    }
+
+    receive() external payable {}
```
AnnaShaleva: Why do we need an empty-arguments receive? According to https://docs.soliditylang.org/en/latest/contracts.html#receive-ether-function, this function will be executed on a call to the contract with empty calldata. I don't think we should allow such transfers to the contract. But maybe I'm missing something.

contracts/solidity/Governance.sol
```diff
@@ -0,0 +1,319 @@
...
+
+    modifier onlyConsensus() {
+        require(isMiner(msg.sender), "Not Consensus");
```
AnnaShaleva: Could we make error messages more verbose? Maybe at least something like "Sender is not a consensus member" for this case? It's a user-facing thing.

contracts/solidity/Governance.sol
```diff
@@ -0,0 +1,319 @@
...
+        }
+        return false;
+    }
+
+    function propose(
+        uint startHeight,
+        address[] memory miners
+    ) external override onlyConsensus {
+        require(startHeight > block.number, "invalid startHeight");
+        require(miners.length > 0, "invalid miners lenght");
```
AnnaShaleva: What's the number of validators `Draft` can be created for? Don't we restrict it in any way?

contracts/solidity/Governance.sol
```diff
@@ -0,0 +1,319 @@
...
+        // check miners order
+        if (miners.length > 1) {
+            for (uint i = 0; i < miners.length - 1; i++) {
+                require(miners[i] < miners[i + 1], "invalid miners order");
+            }
+        }
+
+        require(
+            block.number > lastStartHeight,
+            "propose should after last phase active"
```
AnnaShaleva: `Propose should be called after last phase start` or `Last phase is not active yet, proposing is prohibited`?

privnet/four/genesis_template.json
```diff
@@ -32,6 +32,48 @@
     "node2": { "balance": "100000000000000000000" },
     "node3": { "balance": "100000000000000000000" },
     "node4": { "balance": "100000000000000000000" },
-    "node5": { "balance": "100000000000000000000" }
+    "node5": { "balance": "100000000000000000000" },
+    "0x1212000000000000000000000000000000000000": {
+      "balance": "0",
```
AnnaShaleva: We need to update not only template files, but also all `genesis_privnet.json` files.

genesis.json
```diff
@@ -0,0 +1,76 @@
+{
```
AnnaShaleva: Why do we need this new `genesis.json` file?

privnet/four/genesis_template.json
AnnaShaleva: And please, rebase onto current bane-main. There are seven-nodes privnet configurations available.

contracts/solidity/Governance.sol
```diff
@@ -0,0 +1,319 @@
...
+    // GovReward contract
+    address public constant govReward =
+        0x1212000000000000000000000000000000000003;
+
+    // the last Phase's start height, default 1
+    uint public lastStartHeight;
+    // draft list start id, default 1
+    uint public startDraftId;
+    // draft list end id, default 0
+    uint public endDraftId;
```
AnnaShaleva: I see that you set the default values in the genesis allocations. How do you define the DB prefixes for these values? (`0x0000000000000000000000000000000000000000000000000000000000000000`, etc.)

Are you sure that these prefixes won't change after contract version update? We probably need to add some comment about it to the contract's code, because we need to keep these values in sync with genesis configuration.

contracts/solidity/Governance.sol
```diff
@@ -0,0 +1,319 @@
...
+        });
+        emit Propose(msg.sender, endDraftId, startHeight, miners);
+    }
+
+    function getDraftList() external view override returns (Draft[] memory) {
+        uint count = endDraftId + 1 - startDraftId;
+        Draft[] memory drafts = new Draft[](count);
+        for (uint i = 0; i < count; i++) {
+            drafts[i] = draftMap[startDraftId + i];
+        }
```
AnnaShaleva: A style comment: how about initializing `startDraftId` with `0`, `endDraftId` with `0`, and starting drafts enumeration from `0`? Then you won't have problems with indexes and here it will be possible to use `uint count = endDraftId - startDraftId` and the following:
```
        for (uint i = 0; i < endDraftId; i++) {
            drafts[i] = draftMap[i];
        }
```

contracts/solidity/Governance.sol
```diff
@@ -0,0 +1,319 @@
...
+        return drafts;
+    }
+
+    function vote(
+        uint256 draftId,
```
AnnaShaleva: Why `draftId` has `uint256` type here and `uint` type in the struct declaration?

contracts/solidity/Governance.sol
```diff
@@ -0,0 +1,319 @@
...
+        );
+        require(
+            draftMap[draftId].startHeight > block.number,
+            "invalid draft start height"
+        );
+        require(amount >= MIN_VOTE_AMOUNT, "insufficient amount");
+        require(msg.value >= amount, "insufficient value");
+
+        // add new record
+        draftVoterMap[draftId][msg.sender] += amount;
```
AnnaShaleva: But what if `msg.value` is *greater* than `amount`? Shouldn't we add `msg.value` instead of `amount`?

contracts/solidity/Governance.sol
```diff
@@ -0,0 +1,319 @@
...
+        rewardShareMap[draftId][msg.sender] += amount;
+        draftValueMap[draftId] += amount;
+        emit Vote(msg.sender, draftId, amount);
+
+        // check vote balance
+        uint votedBalance = draftValueMap[draftId];
+        if (votedBalance >= VOTE_TARGET_AMOUNT) {
```
AnnaShaleva: Currently we don't have a way to customize `VOTE_TARGET_AMOUNT`, it's hard-coded. However, I'd suggest to consider the way to customize it (some setter with restricted set of callers). What do you think?

contracts/solidity/Governance.sol
```diff
@@ -0,0 +1,319 @@
...
+                voteAmount: votedBalance
+            });
+            phaseMap[draft.startHeight] = phase;
+            lastStartHeight = draft.startHeight;
+            startDraftId = endDraftId + 1;
+
+            // record current reward balance
+            rewardRecord[draft.startHeight] =
+                govReward.balance +
+                totalWithdrawedReward;
```
AnnaShaleva: OK, I see that we can track only voters reward balance, not the CNs miners reward that is transferred to the governance contract (also docs warns about it in the "Warning" section: https://docs.soliditylang.org/en/latest/contracts.html#receive-ether-function). But what if we store voter's reward balance in the contract and then the diff between `govReward.balance` and the actual value would be the CNs reward.

contracts/solidity/Governance.sol
AnnaShaleva: OK, I see there's refund in the end.

contracts/solidity/Governance.sol
AnnaShaleva: Also, I think we need to be careful with ` govReward.balance` usage here anyway. We should prohibit no-data transfers to this contract.

contracts/solidity/Governance.sol
```diff
@@ -0,0 +1,319 @@
...
+    // the min balance for voting
+    uint public constant MIN_VOTE_AMOUNT = 1 ether;
+    // the balance target for a vote to pass
+    uint public constant VOTE_TARGET_AMOUNT = 3000000 ether;
+    // GovReward contract
+    address public constant govReward =
+        0x1212000000000000000000000000000000000003;
+
+    // the last Phase's start height, default 1
+    uint public lastStartHeight;
```
AnnaShaleva: Do we have a Phase duration constraint in the contract?

contracts/solidity/Governance.sol
```diff
@@ -0,0 +1,319 @@
...
+    function vote(uint draftId, uint value) external payable;
+
+    // revoke vote
+    function revokeVote(uint draftId) external;
+
+    // get current consensus phase
+    function getCurrentPhase() external view returns (Phase memory);
+
+    // get consensus phase by start height
+    function getPhaseByHeight(uint height) external view returns (Phase memory);
```
AnnaShaleva: Correct me if I'm wrong, please. AFAIU, this method returns the current `Phase` which contains `address[] miners` which is the list of current network validators. This list remains constant throughout the whole Phase (is it true?). This method will be used by dBFT engine to retrieve current block validators that should sign the subsequent block. However, as you know, dBFT should include NextConsensus address in the block which is a hashed flattened list of validators for the next block. In case of middle-epoch block next block validators match the current block validators, but for the last block in the epoch we need an ability to retrieve validators list for the next block. based on the current contract storage state. Currently I don't see the functionality for this in the contract.

chenquanyu: `getPhaseByHeight` can be used to get `NextConsensus ` by passing the next block height. `getCurrentPhase` is used to get current Phase.

contracts/solidity/Governance.sol
chenquanyu: No, the Phase duration depends on voting process.

contracts/solidity/Governance.sol
chenquanyu: The prefixes are fixed if the type and order are fixed. They are set in genesis file, please refer to this article https://medium.com/@flores.eugenio03/exploring-the-storage-layout-in-solidity-and-how-to-access-state-variables-bf2cbc6f8018. We can also get the prefixes from Remix storage after initialing them in `constructor`.

contracts/solidity/Governance.sol
chenquanyu: Yes, we only need `vote()` to be payable for now.

contracts/solidity/Governance.sol
chenquanyu: Now we don't restrict it, I don't know how many we should restrict.

contracts/solidity/Governance.sol
chenquanyu: `0` is the default value of `uint` in solidity, sometimes the same as null, so I avoid using it in not null value.

contracts/solidity/Governance.sol
chenquanyu: uint and uint256 are the same type, will unify it.

contracts/solidity/Governance.sol
chenquanyu: If it can be set by validators, maybe we can set it in policy contract.

Makefile
chenquanyu: We don't need http open in dev mode, will remove it.

genesis.json
chenquanyu: Will remove it.

contracts/solidity/Governance.sol
chenquanyu: > But what if we store voter's reward balance in the contract and then the diff between govReward.balance and the actual value would be the CNs reward.

All the transaction fee reward will be sent to `govReward` contract, the governance contract only keeps the Gas for voting.

privnet/four/genesis_template.json
AnnaShaleva: You're right, this tremendous  set of genesis config files is a mess. Will be improved after #112 merge.

contracts/solidity/Governance.sol
AnnaShaleva: Then let's remove it.

contracts/solidity/Governance.sol
AnnaShaleva: Good,thanks! Then we need to add a comment to this block of fields requiring genesis storage initialisation. The comment should contain a note about the fact that we should preserve the order of these fields and we should keep these values in sync with genesis configuration, it's important.

contracts/solidity/Governance.sol
AnnaShaleva: Do you have an updated version of code? I see your reactions, but the code isn't changed, so probably you forgot to push new commits?

contracts/solidity/Governance.sol
AnnaShaleva: But according to the "NeoX Governance Discussion":
>During each sidechain election period (which lasts one epoch, approximately 7 days), a total of 7 nodes are elected as consensus nodes, with 1 node becoming the block-producing node (Speaker) and 6 witness nodes (DBFT).

Do we have our Governance proposal updated? If https://docs.google.com/document/d/11iDJGpCDNrKmPQN8TOhmEqBlWZAOstH15j3Afc2xoQo/edit?mode=html is a latest relevant version then we need to modify contract code. If it's an outdated link, then please, provide the relevant one.

To restrict epoch duration I'd propose to bind to the number of blocks (not to the time period). So given 15-seconds blocks we need to use  `(7 * 24 * 60 * 60) / 15 = 40320` blocks as one epoch and Governance contract should be aware of it.

contracts/solidity/Governance.sol
AnnaShaleva: Good, we need to test it during #14 implementation.

contracts/solidity/Governance.sol
AnnaShaleva: According to the https://docs.google.com/document/d/11iDJGpCDNrKmPQN8TOhmEqBlWZAOstH15j3Afc2xoQo/edit?mode=html we need 7 validators for every epoch. This value is immutable within a network scope, and thus, can be moved to a variable those state can be setup via genesis allocations.

contracts/solidity/Governance.sol
```diff
@@ -0,0 +1,319 @@
...
+    function withdrawERC20(address to, address token, uint256 amount) external;
+
+    function withdraw(address to, uint256 amount) external;
+}
+
+contract Governance is IGovernance {
+    // the min balance for voting
+    uint public constant MIN_VOTE_AMOUNT = 1 ether;
+    // the balance target for a vote to pass
+    uint public constant VOTE_TARGET_AMOUNT = 3000000 ether;
```
AnnaShaleva: This value is not fixed in the [proposal](https://docs.google.com/document/d/11iDJGpCDNrKmPQN8TOhmEqBlWZAOstH15j3Afc2xoQo/edit?mode=html). We either need to make it customizable by validators via restricted setters or we need to agree on this in Governance.

contracts/solidity/Governance.sol
```diff
@@ -0,0 +1,319 @@
...
+    function propose(
+        uint startHeight,
+        address[] memory miners
+    ) external override onlyConsensus {
+        require(startHeight > block.number, "invalid startHeight");
+        require(miners.length > 0, "invalid miners lenght");
+        // check miners order
+        if (miners.length > 1) {
+            for (uint i = 0; i < miners.length - 1; i++) {
+                require(miners[i] < miners[i + 1], "invalid miners order");
```
AnnaShaleva: According to [proposal](https://docs.google.com/document/d/11iDJGpCDNrKmPQN8TOhmEqBlWZAOstH15j3Afc2xoQo/edit?mode=html) for every proposed miner it should be true that:
> Stakes a deposit of 1000 GAS in the NeoX governance interface, which will be returned upon normal node exit (with a tentative locking period of one month equivalent to the same number of blocks).

I think this check should be presented on the contract level.

contracts/solidity/Governance.sol
```diff
@@ -0,0 +1,319 @@
...
+        }
+    }
+
+    function safeTransferETH(address to, uint256 value) internal {
+        (bool success, ) = to.call{value: value}(new bytes(0));
+        require(success, "safeTransferETH: ETH transfer failed");
+    }
+
+    function revokeVote(uint draftId) external override nonReentrant {
+        require(draftId <= endDraftId, "invalid draftId");
```
AnnaShaleva: Shouldn't we require `draftId == endDraftId`? `revokeVote` is allowed for the latest in-progress draft, right?

Because according to the [proposal](https://docs.google.com/document/d/11iDJGpCDNrKmPQN8TOhmEqBlWZAOstH15j3Afc2xoQo/edit?mode=html):
> It is necessary to re-vote during each epoch (TBD, re-vote is easier for calculate the fee reward),

So after every epoch all previous votes should be automatically revoked. But this code is currently not implemented in the contract.

And to be honest, I'd propose not to perform automatic votes revoking after every epoch, because it might be complicated for the users to perform voting every week. It may decrease motivation to vote. And I think we can safely handle voters rewards without automatic revoking in the end of every epoch. @roman-khimov @txhsl @chenquanyu and everyone interested, need your opinion here.

AnnaShaleva (commented):
Oh, the PR was updated, so I'm leaving existing comments here for now and continue review once my review is re-requested.

@chenquanyu, please, tag me once you finish adjustments.

contracts/solidity/Governance.sol
chenquanyu: > because it might be complicated for the users to perform voting every week

User don't need to vote unless they want to change the consensus, the phase has no time limit. I think we should update the [proposal](https://docs.google.com/document/d/11iDJGpCDNrKmPQN8TOhmEqBlWZAOstH15j3Afc2xoQo/edit?mode=html) first.

AnnaShaleva: But how about initial Governance proposal? See the https://github.com/bane-labs/go-ethereum/pull/7#discussion_r1462941106.

AnnaShaleva: > I think we should update the [proposal](https://docs.google.com/document/d/11iDJGpCDNrKmPQN8TOhmEqBlWZAOstH15j3Afc2xoQo/edit?mode=html) first.

Agree, let's discuss this set of proposal-related issues on the synchonisation meeting tomorrow. We need a single view of the desired Governance model to properly review this PR.

contracts/solidity/Governance.sol
chenquanyu: We may set this to policy contract, the values in policy conract are voted by validators.

contracts/solidity/GovReward.sol
```diff
@@ -0,0 +1,70 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
```
AnnaShaleva: Do we really need the previous version of GovReward and Governance contracts? We have V2 now, and V1 is never used.

contracts/solidity/GovernanceVote.sol
```diff
@@ -0,0 +1,70 @@
...
+    // events for voting
+    event Vote(address voter, bytes32 methodKey, bytes32 paramKey);
+    event VotePass(bytes32 methodKey, bytes32 paramKey);
+
+    // governance reward contact
+    address public constant govReward =
+        0x1212000000000000000000000000000000000003;
+
+    // vote mapping, method key ->(user address -> param key)
+    mapping(bytes32 => mapping(address => bytes32)) private voteMap;
```
AnnaShaleva, line 18: I think we need some kind of expiration for voting proposals. To clear expired (too old and useless) vote requests.

contracts/solidity/GovernanceVote.sol
```diff
@@ -0,0 +1,70 @@
...
+        for (uint i; i < voters.length; i++) {
+            if (voteMap[methodKey][voters[i]] == paramKey) {
+                votedCount++;
+            }
+        }
+        return votedCount > (voters.length * 2) / 3;
+    }
+
+    modifier needVote(bytes32 methodKey, bytes32 paramKey) {
+        require(isMiner(msg.sender), "Not Miner");
```
AnnaShaleva: Consider situation: if some validator votes for some method, but not enough votes are collected and vote request is still pending. Then next epoch is coming and validators set is being recalculated and changed. Then `voteMap` contains stale votes from previous validators, whereas technically it's correct to remove old votes from previous validator set and start to collect new votes from the new epoch validators.

contracts/solidity/GovernanceVote.sol
```diff
@@ -0,0 +1,70 @@
...
+        bytes32 paramKey
+    ) internal view returns (bool isPass) {
+        address[] memory voters = IGovReward(govReward).getMiners();
+        uint votedCount;
+        for (uint i; i < voters.length; i++) {
+            if (voteMap[methodKey][voters[i]] == paramKey) {
+                votedCount++;
+            }
+        }
+        return votedCount > (voters.length * 2) / 3;
```
AnnaShaleva: Why do we use dBFT threshold here for policy? Shouldn't it be the majority-style agreement, i.e. `(N+1)/2`, like in Neo3?

genesis.json
```diff
@@ -0,0 +1,81 @@
...
```
AnnaShaleva: This `genesis.json` file should be removed. We have `/config/genesis_testnet.json` file that should include contracts for T3-specific values, and we have a set of genesis files for privnet that should also be updated with the final contracts versions.

But this may be done once all our contracts are finalized.

contracts/solidity/Policy.sol
```diff
@@ -0,0 +1,47 @@
...
+
+    //  cancel blacklist
+    function removeBlackList(
+        address _addr
+    )
+        external
+        needVote(keccak256("removeBlackList"), keccak256(abi.encode(_addr)))
+    {
+        require(isBlackListed[_addr], "Policy: Blacklist does not exist");
+        isBlackListed[_addr] = false;
```
AnnaShaleva: Just remove from `isBlackListed`? No need to store it in the map.

contracts/solidity/GovReward.sol
chenquanyu: I will keep only V2 as the GovReward and Governance contracts.

contracts/solidity/GovernanceVote.sol
chenquanyu: Mapping is not traversable and it will cost more gas to find the expired keys.

contracts/solidity/GovernanceVote.sol
chenquanyu: We calculate the votes with current consensus list, I think the vote from the same candidate at pre epoch should keep active in this epoch.

contracts/solidity/Policy.sol
chenquanyu: They are nearly the same in solidity, please refer to https://ethereum.stackexchange.com/questions/44433/delete-an-entry-in-a-mapping. But I think `delete` is more readable.

genesis.json
chenquanyu: Will remove it once our contracts are finalized.

contracts/solidity/GovernanceVote.sol
chenquanyu: I think both are OK, will change it if you think `(N+1)/2` is better.

genesis.json
chenquanyu: I rollbacked the changes in genesis templates, will update them once contracts are finalized.

contracts/solidity/GovernanceVote.sol
AnnaShaleva: But we have OnPersist, we may call *any* contract during OnPersist.

contracts/solidity/Governance.sol
```diff
@@ -0,0 +1,361 @@
...
+        for (uint i = 0; i < length; i++) {
+            epochStartGasPerVote[candidates[i]][
+                currentEpochStartHeight / epochDuration
+            ] = candidateGasPerVote[candidates[i]];
+        }
+
+        // compute and update consensus
+        currentEpochStartHeight = block.number;
+        if (length < consensusSize || totalVotes < voteTargetAmount) {
+            currentConsensus = standByValidators;
```
AnnaShaleva, line 289: Good, it should work.

nicolegys: So for epoch change, we expect to choose the standByValidators as consensus whenever there are not enough candidates/votes?

AnnaShaleva: Yes, should be like this. Like in N3.

AnnaShaleva (commented):
Seems that we have the final T3 contracts implementation here. Before we merge this PR, please, consider the following:
1. Let's squash commits into some logical blocks, because we need to keep our git history clear. For example, in commit 1 we may add all system contracts implementations, and in the commit 2 we may update all genesis files.
2. Ensure that contract hashes are correct  and relevant everywhere in the contracts and match the hashes described in https://github.com/bane-labs/go-ethereum/pull/7#issue-1877141548. I'll add the rest of them to `systemcontracts` package later.
3. Since contract implementation is final for T3, then we need to compile all contracts and update both T3 `genesis_testnet.json` file (`./config/genesis_testnet.json`) and privnet genesis files (`./privnet/*/genesis_privnet.json` and `./privnet/*/genesis_template.json`). Please, be careful with storage items since all genesis configuration files have different set of validators and different number of CNs.
4. `genesis.json` file should be removed.
5. Add `Close #5`, `Close #2`, `Close #3` to the PR message for proper GH issue links.

contracts/solidity/GovernanceVote.sol
chenquanyu: This is an abstract contract, I'm not sure what should be expired vote requests. I think this should be hanlded by the contract that inherits this contract if needed.

contracts/solidity/Governance.sol
```diff
@@ -115,11 +115,13 @@ contract Governance is IGovernance, ReentrancyGuard, UUPSUpgradeable {
         address[] memory validators = currentConsensus;
         uint length = validators.length;
         for (uint i = 0; i < length; i++) {
-            candidateGasPerVote[validators[i]] +=
-                (msg.value * shareRateOf[validators[i]] * scaleFactor) /
-                consensusSize /
-                1000 /
-                receivedVotes[validators[i]];
+            if (receivedVotes[validators[i]] != 0) {
```
AnnaShaleva: I think it's not quite correct to reward only user-chosen validators (but you may disagree, it's just my thought). In the start of the network we'll use stand by validators for some time before new candidates receive enough votes, and there will be some transactions in the network during this period of time. Should we reward stand by validators for that? I think yes, because they do their job, they accept blocks.

It's just about the fact that stand by validators don't share their reward with voters because they don't have voters. But in general, I think we should reward stand by validators if they are the current consensus.

txhsl: I think they are rewarded in this code, these lines of modification are checking whether a candidate has received any votes from voters, if not, then `candidateGasPerVote` will not be updated.
This `if` case only works for voter reward sharing, and the below down `_safeTransferETH` still works to transfer governance reward to block validators, regardless of if they are stand-by validators.

txhsl: Let me know if I'm wrong here😅.

AnnaShaleva: My bad, everything is OK here.

contracts/solidity/Governance.sol
```diff
@@ -0,0 +1,375 @@
...
+        uint length = validators.length;
+        for (uint i = 0; i < length; i++) {
+            if (receivedVotes[validators[i]] != 0) {
+                candidateGasPerVote[validators[i]] +=
+                    (msg.value * shareRateOf[validators[i]] * scaleFactor) /
+                    consensusSize /
+                    1000 /
+                    receivedVotes[validators[i]];
+            }
+            _safeTransferETH(
```
nicolegys, line 146: So if one's `receivedVotes`==0, the share part of its rewards will stay in this contract forever? But this scenario doesn't seem to happen, because voters tend to get more rewards by voting for candidates with lower votes.

contracts/solidity/Governance.sol
```diff
@@ -0,0 +1,375 @@
...
+        address[] memory candidates = candidateList.values();
+        uint length = candidates.length;
+        for (uint i = 0; i < length; i++) {
+            epochStartGasPerVote[candidates[i]][
+                block.number / epochDuration
+            ] = candidateGasPerVote[candidates[i]];
+        }
+
+        // compute and update consensus
+        if (length < consensusSize || totalVotes < voteTargetAmount) {
```
nicolegys, line 288: When some candidates exit, `totalVotes` won't decrease until their voters revoke vote. So the number of valid votes in the network may not be enough, but we still choose new consensus from the current candidates.

contracts/solidity/Governance.sol
txhsl: Yes, these rewards will be locked forever, or in another word, the same as be burned. The thing is, the candidate wants to share with its voters, but nobody can receive them. This happens when a stand-by validator registers as a normal candidate, set a none-zero `shareRate`, and is selected as a consensus member, if it is reasonable.

contracts/solidity/Governance.sol
txhsl: Yes, it is an issue, but since there are few candidate exits on N3, we can fix this in the future if it is not urgent. I will create a new issue and another PR to fix it.

chenquanyu:
I have tested the `upgradeAndCall` in `GovProxyAdmin` and found the `reverted` error when calling to `UUPSUpgradeable` methods `proxiableUUID` and `upgradeToAndCall`. The root cause is `immutable __self` in `UUPSUpgradeable` is not avaliable in pre-compile contract.

AnnaShaleva:
> The root cause is immutable __self in UUPSUpgradeable is not avaliable in pre-compile contract.

Is this fact connected with the way how we deploy system contracts? I.e. some storage items are missing in genesis allocations?

chenquanyu:
> Is this fact connected with the way how we deploy system contracts? I.e. some storage items are missing in genesis allocations?

Immutable variables are evaluated once at construction time and their value is copied to all the places in the code where they are accessed. So it has nothing to do with storage, we are solving this problem by overriding these methods which use "immutable __self" variable.

contracts/solidity/Governance.sol
AnnaShaleva: Issue, please.

config/genesis_testnet.json
```diff
@@ -26,50 +26,71 @@
...
     }
   }
-}
+}
```
AnnaShaleva, line 97: Here and below no new line in the end of the file. Please, consider adding new line in the end.

config/genesis_testnet.json
```diff
@@ -26,50 +26,71 @@
...
-      "coinbase": "0x1212100000000000000000000000000000000003"
+      "coinbase": "0x1212000000000000000000000000000000000003"
     }
   },
   "difficulty": "1",
   "gasLimit": "30000000",
   "alloc": {
-    "cbbeca26e89011e32ba25610520b20741b809007": {
-      "balance": "100000000000000000000000"
+    "0x1212000000000000000000000000000000000000": {
```
AnnaShaleva, line 35: Why have you removed the initial balance from validator nodes? Is Steven OK with it?

AnnaShaleva (commented):
LGTM otherwise.

config/genesis_testnet.json
chenquanyu: Yes, we want to simulate the case for mainnet.

config/genesis_testnet.json
chenquanyu: It's formated by vs code, do we need the new line?

contracts/solidity/Governance.sol
chenquanyu: New issue #161

config/genesis_testnet.json
AnnaShaleva: OK, let's have it as is.

AnnaShaleva (approved):

AnnaShaleva:
@roman-khimov, @txhsl, @nicolegys, @ZhangTao1596 ready for the final review&merge.

contracts/solidity/Policy.sol
```diff
@@ -0,0 +1,83 @@
...
+    function setMinGasPrice(
+        uint256 _gasPrice
+    )
+        external
+        needVote(keccak256("setMinGasPrice"), keccak256(abi.encode(_gasPrice)))
+    {
+        require(_gasPrice > 0, "Policy: setMinGasPrice invalid parameter");
+        minGasPrice = _gasPrice;
+        emit SetMinGasPrice(_gasPrice);
+    }
```
AnnaShaleva: BTW, it's just came to my mind, we need a getter for `minGASPrice` so that the external NeoX users who create transactions are able to properly retrieve this value to build good transaction with proper GAS price.

AnnaShaleva: Since right now this setting is being enforced by the system.

chenquanyu: I think we should change `eth_gasPrice` interface.

AnnaShaleva: Let's have https://github.com/bane-labs/go-ethereum/issues/162 for this. But I think that getter should be added to Policy contract as far, in order to be able to get this value via "honest" contract invocation. @txhsl, @roman-khimov, what do you think?

AnnaShaleva: I'm not talking about the immediate change, we may create an issue for that and plan it to 0.2.0 to be able to properly test the updated Policy version.

chenquanyu: > But I think that getter should be added to Policy contract as far, in order to be able to get this value via "honest" contract invocation. @txhsl, @roman-khimov, what do you think?

The getter method is added by default in solidity for public fields, just call `minGasPrice()` with policy contract.

chenquanyu:
We have not yet added the function of burning fees. We can send part of fees to an empty address in the governance contract to burn and configure this burning ratio in the policy contract. Issue opened in #163.

config/genesis_testnet.json
```diff
@@ -26,50 +26,71 @@
...
+        "0x360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbc": "0x1212100000000000000000000000000000000001"
+      }
     },
-    "a51fe05b0183d01607bf48c1718d1168a1c11171": {
-      "balance": "100000000000000000000"
+    "0x1212000000000000000000000000000000000002": {
+      "balance": "0",
+      "code": "0x6080604052600a600c565b005b60186014601a565b6051565b565b6000604c7f360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbc546001600160a01b031690565b905090565b3660008037600080366000845af43d6000803e808015606f573d6000f35b3d6000fdfea2646970667358221220965379b4e29f73cf2ee1957914dc167d4751357fc5df24ccd8586bd4bf9b47e964736f6c63430008180033",
+      "storage": {
+        "0x0000000000000000000000000000000000000000000000000000000000000001": "0x989680",
```
txhsl: I think we may improve this to at least 1 Gwei = 10^9, which is `0x3b9aca00`, or above, to 1.5 Gwei. The current T2 transactions are, `Gas fees (Gwei) Base: 0.000000007 | Max: 1.500000008 | Max priority: 1.5`, which means their basefee is `7 wei`, priority fee is `1.5 Gwei`, and their gas price is `1.500000007 Gwei`.

They are all accepted with `"gasPrice": "0x59682f07", "maxFeePerGas": "0x59682f08", "maxPriorityFeePerGas": "0x59682f00"`. In theory, their `GasFeeCap` is `1.500000008 Gwei` and `GasTipCap` is `1.5 Gwei`. Just mention this issue, I'm again looking at #6.

AnnaShaleva: > I think we may improve this to at least 1 Gwei = 10^9, which is 0x3b9aca00, or above, to 1.5 Gwei. 

Maybe start with the lower value, with 1 Gwei? We can always adjust this value in Policy anyway.

AnnaShaleva (approved):
LGTM.

contracts/solidity/GovProxyAdmin.sol
```diff
@@ -0,0 +1,35 @@
...
+import "./GovernanceVote.sol";
+
+/**
+ * @dev This is an auxiliary contract meant to be assigned as the admin of a {Proxy}.
+ * Use GovernanceVote to manage upgrade
+ */
+contract GovProxyAdmin is GovernanceVote {
+    /**
+     * @dev Upgrades `proxy` to `implementation` and calls a function on the new implementation. See
+     * {TransparentUpgradeableProxy-upgradeToAndCall}.
```
mialbu:
```suggestion
     * @dev Upgrades the implementation of the proxy to `newImplementation`, and subsequently executes the function call
     * encoded in `data`.
     * 
     * See {UUPSUpgradeable-upgradeToAndCall}.
```

contracts/solidity/GovProxyAdmin.sol
```diff
@@ -0,0 +1,35 @@
...
+     *
+     * Requirements:
+     *
+     * - This contract must be the admin of `proxy`.
+     */
+    function upgradeAndCall(
+        UUPSUpgradeable proxy,
+        address implementation,
```
mialbu: Consider renaming this parameter to `newImplementation` for consistency.

contracts/solidity/GovProxyAdmin.sol
mialbu: Or

```suggestion
     * @dev Upgrades the implementation of the proxy to `implementation`, and subsequently executes the function call
     * encoded in `data`.
     * 
     * See {UUPSUpgradeable-upgradeToAndCall}.
```

mialbu (commented):
Minor documentation and consistency-related comments.

AnnaShaleva (approved):

PR #10, opened 2023-09-09
Makefile: add init, start, stop targets for privnet

### LINKED ISSUE #8: Add testing privnet config and minimal instructions
See https://geth.ethereum.org/docs/fundamentals/private-network, let's create a minimal config for Clique-enabled consensus, some well-known key (like NeoGo https://github.com/nspcc-dev/neo-go/tree/master/.docker/wallets) and instructions of how to run a "miner" node and connect another one to it. It will be extended with contracts when they're ready and it will be changed to dBFT once it's there. The idea is to have a simple way to get the network up and running for various purposes.

tatiana-nspcc:
The corrected instructions for running a private Ethereum network (privnet) have been placed in `privnet/README.md`.

Close #8


.gitignore
```diff
@@ -47,4 +47,5 @@ profile.cov
 /dashboard/assets/package-lock.json
 
 **/yarn-error.log
-logs/
\ No newline at end of file
+logs/
+/prnet/
```
AnnaShaleva: s/prnet/privnet everywhere to make it unified with NeoGo privnet setup?

Makefile
```diff
@@ -36,3 +111,43 @@ devtools:
...
+	@echo "Generate  $(GENESIS_WORK_JSON) file"
+	@cp ./genesis_base.json $(MAIN_DIR)/$(GENESIS_WORK_JSON)
+	@echo $$(date +'%y%m%d%H%M') >> $(MAIN_DIR)/networkid.txt
+	@echo "networkid is "$$(cat $(MAIN_DIR)/networkid.txt)
+	@echo "Generate bootnode"
+	$(call generate_bootnode)
+	$(call replace_chainid)
+	@echo "Create accounts"
+	$(call create_account,$(NODE1))
+	$(call create_account,$(NODE2))
```
AnnaShaleva: We don't need new accounts each time the privnet is being setup. Make it once, add these ready-to-use accounts into some repo folder like it is done for NeoGo and use them to run the privnet afterwards. If needed, you can write instructions on how to generate proper accounts, but we need solid constant accounts every privnet run. It will be useful for testing and debugging.

Makefile
```diff
@@ -36,3 +111,43 @@ devtools:
...
+
+prnet_init: prnet_clean
+	@mkdir $(MAIN_DIR)
+	@echo "Generate  $(GENESIS_WORK_JSON) file"
+	@cp ./genesis_base.json $(MAIN_DIR)/$(GENESIS_WORK_JSON)
+	@echo $$(date +'%y%m%d%H%M') >> $(MAIN_DIR)/networkid.txt
+	@echo "networkid is "$$(cat $(MAIN_DIR)/networkid.txt)
+	@echo "Generate bootnode"
+	$(call generate_bootnode)
+	$(call replace_chainid)
```
AnnaShaleva: The same comment is relevant for the chain ID and genesis file, I doubt we need new data each time the network is started. @roman-khimov?

Makefile
```diff
@@ -36,3 +111,43 @@ devtools:
 	env GOBIN= go install ./cmd/abigen
 	@type "solc" 2> /dev/null || echo 'Please install solc'
 	@type "protoc" 2> /dev/null || echo 'Please install protoc'
+
+prnet_init: prnet_clean
```
AnnaShaleva: How about `env_single`, `env_up`, `env_down` and `env_clean` to match the NeoGo pattern? See the https://github.com/nspcc-dev/neo-go/blob/2d88ed9fe5d58cd77280aa07c5b4fa95d5092abf/Makefile#L136 and related targets.

Makefile
```diff
@@ -36,3 +111,43 @@ devtools:
...
+prnet_clean: prnet_stop
+	@echo "Removing $(MAIN_DIR)"
+	@rm -rf $(MAIN_DIR)
+
+prnet_start:
+	@echo "Starting nodes..."
+	$(call run_bootnode)
+	$(call run_miner_node,$(NODE1),$(NODE1_PORT),$(NODE1_RPC_PORT),$(NODE1))
+	$(call run_node,$(NODE2),$(NODE2_PORT),$(NODE2_RPC_PORT))
+	@echo "OK! Check logs in <node_dir>/geth_node.log"
```
AnnaShaleva: So we don't use Docker for it. Are you sure we'd like to have these nodes running directly on the host machine? I'd suggest to dockerize the privnet network creation from scratch, have a separate configuration file for node2 under some `config` directory and start the single Geth node as docker network, exactly like in NeoGo. This allows to extend the network configuration in future.

Makefile
AnnaShaleva: And BTW, these new targets are all `.PHONY`.

Makefile
AnnaShaleva: @roman-khimov?

genesis_base.json
```diff
@@ -0,0 +1,29 @@
+{
```
AnnaShaleva: We need to put this file in a special directory and add some `privnet` suffix.

genesis_base.json
```diff
@@ -0,0 +1,29 @@
...
+    }
+  },
+  "difficulty": "1",
+  "gasLimit": "7000000",
+  "extradata": "0x0000000000000000000000000000000000000000000000000000000000000000node10000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
+  "alloc": {
+    "node1": { "balance": "100000000000000000000" },
+    "node2": { "balance": "100000000000000000000" }
+  }
+}
```
AnnaShaleva: No new line in the end of the file.

AnnaShaleva (changes requested):
After the setup structure is settle, we need to add the documentation on how to use these new `make` targets and how to start the privnet network and connect some local node to it. We also need to add the GH action that will check that building targets work as expected, exactly like NeoGo's https://github.com/nspcc-dev/neo-go/blob/master/.github/workflows/build.yml.

AnnaShaleva:
Don't forget to add `Close #8` message to PR description or commit description.

Makefile
tatiana-nspcc: @roman-khimov asked me not to use Docker if possible. So, here we are.

Makefile
tatiana-nspcc: There's no "single" environment. It starts with two nodes from the beginning. Also, I don't use Docker, so I think it's different from the NeoGo project. 
If you really want to, we should replace "single" with "init," at least.

Makefile
tatiana-nspcc: The same idea applies here. Just use privnet_stop and privnet_start without privnet_init, and you'll get the same chain ID.

Makefile
tatiana-nspcc: You don't need to generate it each time. As you can see, there are two steps to start privnet: `privnet_init` and `privnet_start`. The first time, you need both of them, but afterward, use only `privnet_stop` and `privnet_start` to use the same privnet without generating new accounts.
Even more than that, I don't want to store passwords on GitHub. I believe it's bad practice. So, I've written a generator.

AnnaShaleva: IMO, we need a reproducible environment and solid addresses at every local machine every time when you clone the repo or checkout wallet changes. What we literally need is to generate these accounts once and store them in some folder like it is done for NeoGo privnet wallets (https://github.com/nspcc-dev/neo-go/tree/master/.docker/wallets). It's development privnet, no one cares about accounts/passwords safety here.

What we may have additionally is some `regenerate_wallets_test.go` test that regenerates privnet wallets or instructions inside the README file or at some other documentation file that describes how to regenerate these wallets. We even can go with this separate make target, but we don't need to call it each time on new setup.

Makefile
AnnaShaleva: Well, my concern was about `prnet` mostly, it looks strange a bit, I got used to `privnet` or `env` in our projects.

Makefile
AnnaShaleva: OK.

Makefile
tatiana-nspcc: Account's addresses and keys are stored in separate files, so we can use them in tests. And in case of regenerating we don't need to update hard-coded config.

roman-khimov: Static keys are easier to handle for test environments. You can keep this logic in case we need to recreate them, but in general static well-known wallets are OK for this particular purpose (it's not production keys). We do this in NeoGo, NeoFS dev-env, AIO. Makes things more predictable.

Makefile
roman-khimov: We better be aligned wrt these values between different instances, hardcoding them is OK. Again, see dev-env, AIO, NeoGo privnets.

Just to reiterate, it's a test environment, we're not trying to create some production networks here.

privnet/node1/geth/chaindata/ancient/chain/bodies.meta
```diff
@@ -0,0 +1 @@
+��
```
AnnaShaleva: Please, add `chaindata` to .gitignore :)

.github/workflows/tests.yml
```diff
@@ -0,0 +1,111 @@
...
+
+jobs:
+  lint:
+    name: Lint
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/setup-go@v4
+        with:
+          go-version: '1.18'
```
AnnaShaleva: Minimum supported Go version for this project is 1.19. And let's use go 1.21 for this, as suggested in https://github.com/ethereum/go-ethereum/blob/90d5bd85bcf2919ac2735a47fde675213348a0a6/.travis.yml#L12.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,111 @@
...
+    - name: Perform CodeQL Analysis
+      uses: github/codeql-action/analyze@v2
+
+  tests:
+    name: Run tests
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-20.04, windows-2022, macos-12]
+        go_versions: [ '1.18', '1.19', '1.20' ]
```
AnnaShaleva: Remove 1.18, add 1.21.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,111 @@
...
+
+    - name: Perform CodeQL Analysis
+      uses: github/codeql-action/analyze@v2
+
+  tests:
+    name: Run tests
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-20.04, windows-2022, macos-12]
```
AnnaShaleva: Well, I'm not sure we need tests for Windows and macOS, although they are supported by the geth. Check out the way how they run tests, it's only for linux distributives: https://github.com/ethereum/go-ethereum/blob/90d5bd85bcf2919ac2735a47fde675213348a0a6/.travis.yml#L108

.github/workflows/tests.yml
```diff
@@ -0,0 +1,111 @@
+name: Tests
```
AnnaShaleva, line 1: We also need a building workflow, it should be run on Ubuntu and it should use `make geth`. Check out the way how NeoGo `Build CLI` GA job works.

privnet/bootnode.log
```diff
@@ -0,0 +1,742 @@
+enode://0cfebf0ca4056001706eb6c793ed29ce8626b51c940ae12db2eb419d5914dfe7926399ff87a4b85165d2bb5ff3704c906ee45004b4f3573ba415fd168470312d@127.0.0.1:0?discport=30305
```
AnnaShaleva: We don't need this log to be included in git, do we?

privnet/genesis_base.json
```diff
@@ -0,0 +1,29 @@
+{
```
AnnaShaleva: Let's change the file name to `genesis_template.json`?

privnet/genesis_privnet.json
```diff
@@ -0,0 +1,29 @@
+{
```
AnnaShaleva: What is the purpose of `privnet/genesis_privnet.json` while we have `privnet/node1/genesis_privnet.json` and `privnet/node2/genesis_privnet.json`?

privnet/node1/geth_create_account.log
```diff
@@ -0,0 +1,13 @@
+INFO [09-17|17:10:08.866] Maximum peer count                       ETH=50 LES=0 total=50
```
AnnaShaleva: I doubt we need this file as far. Please, add all unnecessary files to .gitignore.

AnnaShaleva (commented):
@tatiana-nspcc, looks better, you got the idea right. Finish your current tasks and re-request review, I'll review one more time and try to run it locally.

.gitignore
tatiana-nspcc: Changed.

Makefile
tatiana-nspcc: Thanks, fixed.

genesis_base.json
tatiana-nspcc: Moved it into `privnet` directory

genesis_base.json
tatiana-nspcc: Added.

Makefile
tatiana-nspcc: `prnet` was changed to `privnet`.

.github/workflows/tests.yml
AnnaShaleva: The last step in this job will be:
```
      - name: Build node
        run: make geth
```

privnet/node1/geth/chaindata/ancient/chain/bodies.meta
tatiana-nspcc: I checked, I can't start privnet without this directory.

privnet/genesis_privnet.json
tatiana-nspcc: The triple copy of the `genesis.json` file was mentioned in https://geth.ethereum.org/docs/fundamentals/private-network. It's required to init a new privnet, and it won't work without it. I've added it to `.gitignore`.

Makefile
tatiana-nspcc: Okay, I've added privnet/config.json with addresses, passwords, and chain_id.

Makefile
tatiana-nspcc: Okay, I've added privnet/config.json with addresses, passwords, and chain_id.

.github/workflows/tests.yml
tatiana-nspcc: Fixed.

.github/workflows/tests.yml
tatiana-nspcc: Fixed.

.github/workflows/tests.yml
tatiana-nspcc: I'm leaving only `ubuntu-20.04` here.

.github/workflows/tests.yml
tatiana-nspcc: Added `build.yaml`

privnet/bootnode.log
tatiana-nspcc: Added to `.gitignore`.

privnet/genesis_base.json
tatiana-nspcc: Ok, changed.

privnet/node1/geth_create_account.log
tatiana-nspcc: Added to `.gitignore`.

privnet/node1/geth/chaindata/ancient/chain/bodies.meta
AnnaShaleva: How about adding the content of `chaindata`? We can have some files that are really necessary for the node run, but we don't need the full DB to be stored in git.

tatiana-nspcc: If you mean keeping the empty directory `privnet/node1/geth/chaindata/`, that can also cause problems. Perhaps for now, the better solution is to leave it as it is and sort through the essential and unnecessary files later?

tatiana-nspcc:
In the last force push I rebased the branch on the fresh `bane-main` state and corrected the spelling of `bane-main` in the workflows.

privnet/node1/geth/chaindata/ancient/chain/bodies.meta
AnnaShaleva: What I mean is to add to leave some files that are really necessary for the node run from scratch and add to .gitignore and remove the others.

>to leave it as it is and sort through the essential and unnecessary files later

The problem with this approach is that these files look like the node's DB files. Thus, these files will be changed on every privnet run. Thus, they will be marked in git as "changed" during the development process, and I'd like to avoid this and to have the git index clean after the subsequent privnet run.

privnet/README.md
```diff
@@ -0,0 +1,63 @@
...
+localhost (127.0.0.1). Also, ensure that RPC ports 8552 and 8553 are 
+not being used.
+
+3. Start the private network by running `make privnet_start`.
+Now you can look into the files `privnet/node1/geth_node.log` and
+`privnet/node2/geth_node.log` to see the logs. Or you can use the next command:
+```
+tail -f ./privnet/node1/geth_node.log
+```
+
```
AnnaShaleva: We're building the development privnet, thus it is expected that the following sequens of commands will work:
```
$ cd go-ethereum
$ make geth
$ make privnet_start
```
And currently the result of this sequence is `/bin/sh: 1: geth: not found` in the node logs. That's because we have `geth` binary in the default build folder (in `./build/bin/geth`). What we need is to adjust the Makefile commands so that they run exactly this binary, not just `geth`.

.github/workflows/build.yml
```diff
@@ -0,0 +1,45 @@
...
+        required: false
+        default: ''
+
+jobs:
+  build_cli:
+    name: Build geth
+    runs-on: ${{matrix.os.name}}
+    strategy:
+      matrix:
+        os: [{ name: ubuntu-20.04, bin-name: linux }]
```
AnnaShaleva: You don't need `bin-name`.

.gitignore
```diff
@@ -47,4 +47,10 @@ profile.cov
...
 **/yarn-error.log
-logs/
\ No newline at end of file
+logs/
+
+# privnet
+/privnet/node1/geth_create_account.log
+/privnet/node2/geth_create_account.log
+/privnet/bootnode.log
+/privnet/genesis_privnet.json
```
AnnaShaleva: Then we don't need `genesis_privnet.json` to be ignored, otherwise we won't be able to commit the result of `make privnet_init` on chain regeneration.

.gitignore
```diff
@@ -47,4 +47,10 @@ profile.cov
 /dashboard/assets/package-lock.json
 
 **/yarn-error.log
-logs/
\ No newline at end of file
+logs/
+
+# privnet
+/privnet/node1/geth_create_account.log
+/privnet/node2/geth_create_account.log
```
AnnaShaleva: We need to ignore all logs that are changed from run to run. E.g. ignore `privnet/node1/geth_node.log` and etc. Try to ignore `*.log` files except the ones that are really necessary.

Makefile
```diff
@@ -2,12 +2,104 @@
...
+GENESIS_WORK_JSON = genesis_privnet.json
+
+BOOTNODE_PORT = 30305
+BOOTNODE_LOGLEVEL = 5
+
+RESTRICTED_NETWORK = 127.0.0.0/24
+NAT_POLICY = none
+
+define generate_bootnode
+	@bootnode -genkey $(MAIN_DIR)/bootnode.key
```
AnnaShaleva: @tatiana-nspcc, could you, please, provide the link to `bootnode` command? It's not in the `Build from source code` section of https://geth.ethereum.org/docs/getting-started/installing-geth (we're interested in it as a developers), and it's not a part of the `geth` itself AFAIU.

tatiana-nspcc: Yes, for sure, it's link from the `README.md` file: https://geth.ethereum.org/docs/fundamentals/private-network section `End-to-end example`.

privnet/README.md
tatiana-nspcc: Fixed `README.md` file and `Makefile` accordingly to this comment.

.github/workflows/build.yml
tatiana-nspcc: Fixed.

.gitignore
tatiana-nspcc: Fixed.

.gitignore
tatiana-nspcc: Fixed.

Makefile
```diff
@@ -36,3 +132,75 @@ devtools:
 	env GOBIN= go install ./cmd/abigen
 	@type "solc" 2> /dev/null || echo 'Please install solc'
 	@type "protoc" 2> /dev/null || echo 'Please install protoc'
+
+# Privnet targets
+
+privnet_init_: privnet_clean
```
AnnaShaleva: Remove this target, we already have `privnet_init`, it does all the things properly. As discussed earlier, "Init nodes" step shouldn't be performed at the `privnet_init` stage.

Makefile
```diff
@@ -36,3 +132,75 @@ devtools:
...
+
+privnet_start:
+	@echo "Init nodes"
+	$(call init_node,$(NODE1))
+	$(call init_node,$(NODE2))
+	@echo "Starting nodes..."
+	$(call run_bootnode)
+	$(call run_miner_node,$(NODE1),$(NODE1_PORT),$(NODE1_RPC_PORT),$(NODE1))
+	$(call run_node,$(NODE2),$(NODE2_PORT),$(NODE2_RPC_PORT))
+	@echo "OK! Check logs in privnet/<node_dir>/geth_node.log"
```
AnnaShaleva: We don't need two targets for the same action. As discussed, what we need is "Init nodes" stage to be presented in `privnet_start` target (like it is done currently), but we also need to add a condition to skip nodes initialization in case if the DB is already initialized. That's exactly the purpose `make` is designed for.

And after that `privnet_start_` should be removed.

Makefile
```diff
@@ -2,12 +2,108 @@
...
+	@echo "\t\t\t\"key\": \""$$(cat $(MAIN_DIR)/$(1)/password.txt)\" >> $(MAIN_DIR)/config.json
+	@echo "\t\t}," >> $(MAIN_DIR)/config.json
+	@echo "\t\t{" >> $(MAIN_DIR)/config.json
+    @echo "\t\t\t\"address\": \""$$(cat $(MAIN_DIR)/$(2)/node_address.txt)\", >> $(MAIN_DIR)/config.json
+    @echo "\t\t\t\"key\": \""$$(cat $(MAIN_DIR)/$(2)/password.txt)\" >> $(MAIN_DIR)/config.json
+    @echo "\t\t}" >> $(MAIN_DIR)/config.json
+	@echo "\t]," >> $(MAIN_DIR)/config.json
+	@echo "\t\"chain_id\": \""$$(cat $(MAIN_DIR)/networkid.txt)\" >> $(MAIN_DIR)/config.json
+	@echo "}" >> $(MAIN_DIR)/config.json
+endef
```
AnnaShaleva: We don't need this definition. You have `genesis_template.json`, just copy it and replace all necessary data inside it.

Makefile
```diff
@@ -2,12 +2,108 @@
...
+		--netrestrict $(RESTRICTED_NETWORK) \
+		$(4) >  $(MAIN_DIR)/$(1)/geth_node.log 2>&1 &
+endef
+
+define copy_genesis
+	@cp $(MAIN_DIR)/$(GENESIS_WORK_JSON) $(MAIN_DIR)/$(1)/$(GENESIS_WORK_JSON)
+endef
+
+define init_node
+    @geth init --datadir $(MAIN_DIR)/$(1)  $(MAIN_DIR)/$(GENESIS_WORK_JSON) > $(MAIN_DIR)/$(1)/geth_init.log 2>&1
```
AnnaShaleva:
```suggestion
    @geth init --datadir $(MAIN_DIR)/$(1) $(MAIN_DIR)/$(GENESIS_WORK_JSON) > $(MAIN_DIR)/$(1)/geth_init.log 2>&1
```

Makefile
```diff
@@ -2,12 +2,108 @@
...
+
+define run_bootnode
+    @bootnode -nodekey $(MAIN_DIR)/bootnode.key \
+    	-addr :$(BOOTNODE_PORT) \
+    	-verbosity $(BOOTNODE_LOGLEVEL) > $(MAIN_DIR)/bootnode.log 2>&1 &
+endef
+
+define run_miner_node
+	$(call run_node,$(1),$(2),$(3),--mine --miner.etherbase="0x$$(cat $(MAIN_DIR)/$(4)/node_address.txt)")
+endef
```
AnnaShaleva, line 66: Why it's `miner_node`? We need `signer_node`, because we run a Clique.

And we need the following syntax here:
```
geth <other-flags> --unlock 0x7df9a875a174b3bc565e6424a0050ebc1b2d1d82 --mine
```

privnet/README.md
```diff
@@ -0,0 +1,67 @@
+# Private Ethereum network (privnet)
+
+Here are the corrected instructions for running a private Ethereum network 
+(privnet):
+
+1. Build `geth`:
+```
+$ cd go-ethereum
+$ make geth
+```
```
AnnaShaleva: OK, then it's not enough, because we use a separate bootnode proces (not the fully-qualified node) for it. Thus, we need to build the bootnode binary. Thus, we need to use `make all` instead of just `make geth`.

privnet/README.md
```diff
@@ -0,0 +1,67 @@
+# Private Ethereum network (privnet)
+
+Here are the corrected instructions for running a private Ethereum network 
+(privnet):
```
AnnaShaleva: Add `with Clique as a consensus algorithm.`.

privnet/README.md
```diff
@@ -0,0 +1,67 @@
+# Private Ethereum network (privnet)
+
+Here are the corrected instructions for running a private Ethereum network
```
AnnaShaleva: Remove `corrected`?

privnet/README.md
AnnaShaleva: Add the description of the network, i.e. enumerate its components: bootnode service, Clique consensus node and a simple RPC node.

privnet/README.md
```diff
@@ -0,0 +1,67 @@
...
+`privnet/node2/geth_node.log` to see the logs. Or you can use the next command:
+```
+tail -f ./privnet/node1/geth_node.log
+```
+
+4. To stop the private network, use the command `make privnet_stop`.
+
+Network/chain identifier, node addresses, and passwords are stored in the file
+`privnet/config.json`.
+
```
AnnaShaleva: Add step 5 about cleaning the private network.

.gitignore
```diff
@@ -47,4 +47,10 @@ profile.cov
...
 **/yarn-error.log
-logs/
\ No newline at end of file
+logs/
+
+# privnet
+/privnet/**/*.log
+/privnet/bootnode.log
+/privnet/node1/geth/
+/privnet/node2/geth/
```
AnnaShaleva: No new line in the end of the file.

.gitignore
```diff
@@ -47,4 +47,10 @@ profile.cov
 /dashboard/assets/package-lock.json
 
 **/yarn-error.log
-logs/
\ No newline at end of file
+logs/
+
+# privnet
+/privnet/**/*.log
+/privnet/bootnode.log
```
AnnaShaleva: Create a separate folder for bootnode-related files and logs, please.

privnet/genesis_privnet.json
AnnaShaleva: >The triple copy of the genesis.json file was mentioned

The documentation says only `In each data directory save a copy of the following genesis.json to the top level project directory. `. Both node1 and node2 have their own copy of `genesis.json` files, and the bootnode service doesn't need this. So I'm still not sure we need this file here.

.github/workflows/build.yml
```diff
@@ -0,0 +1,45 @@
...
+          fetch-depth: 0
+
+      - name: Set up Go
+        uses: actions/setup-go@v4
+        with:
+          go-version: '1.21'
+          cache: true
+
+      - name: Build geth
+        run: make geth
```
AnnaShaleva: Let's do `make all` as we use other services as far.

.github/workflows/build.yml
```diff
@@ -0,0 +1,45 @@
...
+  workflow_dispatch:
+    inputs:
+      ref:
+        description: 'Ref to build geth [default: latest bane-main; examples: v0.92.0, 0a4ff9d3e4a9ab432fd5812eb18c98e03b5a7432]'
+        required: false
+        default: ''
+
+jobs:
+  build_cli:
+    name: Build geth
```
AnnaShaleva: s/`Build geth`/`Build geth and services`.

Makefile
```diff
@@ -36,3 +132,75 @@ devtools:
...
+	@find $(MAIN_DIR)/* -type f -not \( -name 'genesis_template.json' -or -name 'README.md' \) -exec rm -f {} +
+
+privnet_start_:
+	@echo "Starting nodes..."
+	$(call run_bootnode)
+	$(call run_miner_node,$(NODE1),$(NODE1_PORT),$(NODE1_RPC_PORT),$(NODE1))
+	$(call run_node,$(NODE2),$(NODE2_PORT),$(NODE2_RPC_PORT))
+	@echo "OK! Check logs in privnet/<node_dir>/geth_node.log"
+
+privnet_start:
```
AnnaShaleva: `privnet_start` still won't work after `make all` because of the wrong executable paths. Correct them, please. (paths to `geth` and `bootnode` executable).

privnet/README.md
AnnaShaleva: The second part is not fixed yet.

Makefile
tatiana-nspcc: Oops, yes, this target shouldn't be in this commit.

Makefile
tatiana-nspcc: Yes, this one is also my fault.

Makefile
tatiana-nspcc: I can't put passwords in the `genesis_template.json`. Do you just want to keep them in separate files like `password.txt`? I thought that you wanted to collect and store this data in one file.

AnnaShaleva: I've confused this file with the base genesis json config, so this comment is irelevant, let's keep it as is.

privnet/genesis_privnet.json
tatiana-nspcc: You are right. I misread it, tried again and remove it.

privnet/README.md
tatiana-nspcc: Sorry, some wrong or non-final commit was pushed. I fixed it.

Makefile
tatiana-nspcc: Fixed.

privnet/README.md
tatiana-nspcc: Replaced to `make all`.

privnet/README.md
tatiana-nspcc: I rewrote this part.

tatiana-nspcc: Done.

privnet/README.md
tatiana-nspcc: Removed. I meant that these instructions are different from the original end-to-end example. But it wasn't clear, so I rewrote the entire section.

.gitignore
tatiana-nspcc: Fixed.

.gitignore
tatiana-nspcc: Done

.github/workflows/build.yml
tatiana-nspcc: Replaced to `make all`.

Makefile
tatiana-nspcc: Sorry, fixed.

privnet/README.md
tatiana-nspcc: Added. But in my opinion, it shouldn't be used separately; it's needed for make privnet_init to re-init privnet.

AnnaShaleva:
@tatiana-nspcc, if everything except the https://github.com/bane-labs/go-ethereum/pull/10#discussion_r1334223842 is fixed, then let's move on to have some initial implementation of https://github.com/bane-labs/go-ethereum/issues/11. I'll review this PR one more time and experiment with setups a bit later.

Makefile
tatiana-nspcc: It works as you have written. I checked `miner.etherbase` parameter, it doesn't work without it, the error is:
```
ERROR[09-25|10:30:50.624] Cannot start mining without etherbase    err="etherbase must be explicitly specified"
Fatal: Failed to start mining: etherbase missing: etherbase must be explicitly specified

```

AnnaShaleva:
@roman-khimov, could you review, please?

Makefile
```diff
@@ -159,18 +159,18 @@ privnet_init: privnet_clean
 
 privnet_nodes_stop:
 	@echo "Killing nodes processes"
-	@killall -9 geth > /dev/null 2> /dev/null || :
+	@killall -w -v -9 geth || :
 
 privnet_bootnode_stop:
 	@echo "Killing bootnode processes"
-	@killall -9 bootnode > /dev/null 2> /dev/null || :
+	@killall -w -v -9 bootnode || :
```
tatiana-nspcc, line 147: Typos in the commit message: prcesses

Makefile
```diff
@@ -50,9 +50,7 @@ endef
 define create_account
     @mkdir -p $(MAIN_DIR)/$(1)
     @echo $(call generate_password) > $(MAIN_DIR)/$(1)/password.txt
-    @$(GOBIN)/geth --datadir $(MAIN_DIR)/$(1) \
-    	account new \
-    	--password $(MAIN_DIR)/$(1)/password.txt >  $(MAIN_DIR)/$(1)/geth_create_account.log 2>&1
+    @$(GOBIN)/geth --datadir $(MAIN_DIR)/$(1) account new --password $(MAIN_DIR)/$(1)/password.txt
```
tatiana-nspcc: Typo in the commit message: goot

Makefile
```diff
@@ -31,7 +31,7 @@ NAT_POLICY = none
 define generate_bootnode
 	@mkdir -p $(MAIN_DIR)/$(BOOTNODE)
 	@$(GOBIN)/bootnode -genkey $(MAIN_DIR)/$(BOOTNODE)/bootnode.key
-	@echo $$($(GOBIN)/bootnode --writeaddress -nodekey $(MAIN_DIR)/$(BOOTNODE)/bootnode.key) >> $(MAIN_DIR)/$(BOOTNODE)/bootnode_address.txt
+	@echo $$($(GOBIN)/bootnode --writeaddress -nodekey $(MAIN_DIR)/$(BOOTNODE)/bootnode.key) > $(MAIN_DIR)/$(BOOTNODE)/bootnode_address.txt
```
tatiana-nspcc, line 34: Could you please clarify what issues you encountered with it not working properly? I'm asking in a general sense because it appears to be functioning correctly on my local computer. I have noticed some mistakes in the file writing process.

tatiana-nspcc (commented):
I appreciate your valuable contribution to this PR. It looks great, especially the addition of 400 lines of minimal instructions in README.md :)

Makefile
AnnaShaleva: > appears to be functioning correctly on my local computer

That's true, it worked with the old `make privnet_clean` scheme when you removed the whole `privnet` directory except a couple of files. And now `privnet_clean` has different behaviour, it cleans only the DB files and keeps all network configuration settings so that the user is ab;e to clean the DB and start the new chain from the genesis with the old settings. It is needed for development and this scheme matches the one we have in NeoGo/AIO/NeoFS devenv.

Thus, we have to replace the existing content of mentioned config files on `make privnet_init`, not just append to it.

Makefile
AnnaShaleva: Fixed.

Makefile
AnnaShaleva: Fixed.

AnnaShaleva (approved):

AnnaShaleva:
@tatiana-nspcc, agree to merge?

tatiana-nspcc:
> @tatiana-nspcc, agree to merge?

@AnnaShaleva, Yes!

PR #15, opened 2023-09-15
add policy contract

Tobeyw:
This PR is a code implementation of #3,and depends on goverranceVote contract.


txhsl:
Do we need the `blackList` and `getBlackList()`? I don't find them in Neo's Policy contract.

PR #37, opened 2023-11-28
update genesis.json file

nicolegys:
fix the wrong bytecode for 0x1212100000000000000000000000000000000001


chenquanyu:
I used solidity 0.8.18 compiler and set optimizer runs to 200. And now the bytecode is latest.

nicolegys:
out of date

PR #38, opened 2023-12-04
crypto: add basic multisig primitives

roman-khimov:
Just a minimal basic building block to share.

### DIFF
--- a/crypto/multisig.go
+++ b/crypto/multisig.go
@@ -1,0 +1,87 @@
+// Copyright 2023 NeoSPCC
+// This file is part of the go-ethereum library.
+//
+// The go-ethereum library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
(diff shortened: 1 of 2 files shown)


AnnaShaleva (approved):

crypto/multisig.go
```diff
@@ -0,0 +1,87 @@
...
+		sPubs = append(sPubs, pubkey)
+	}
+	sort.Slice(sPubs, func(i, j int) bool {
+		return bytes.Compare(sPubs[i], sPubs[j]) < 0
+	})
+	var vi int
+	for si := range sPubs {
+		for ; vi < len(vPubs); vi++ {
+			if bytes.Compare(vPubs[vi], sPubs[si]) == 0 {
+				break
```
chenquanyu, line 79: This break will not add vi, so the check will pass with duplicate sPubs[si]. For example, we build the signatures with signers [1,1,3,4,5].

PR #39, opened 2023-12-04
dbft: add message pool

roman-khimov:
Based on NeoGo, no extensibles, just pure dBFT data.

### DIFF
--- a/eth/protocols/dbft/payload.go
+++ b/eth/protocols/dbft/payload.go
@@ -1,0 +1,55 @@
+// Copyright 2023 NeoSPCC
+//
+// MIT License.
+
+package dbft
+
+import (
+	"errors"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/ethereum/go-ethereum/rlp"
+	"golang.org/x/crypto/sha3"
+)
+
+// Message is any broadcasted DBFT message.
+type Message struct {
+	// ValidBlockStart is the starting height for a payload to be valid.
+	ValidBlockStart uint32
+	// ValidBlockEnd is the height after which a payload becomes invalid.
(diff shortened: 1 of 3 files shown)


eth/protocols/dbft/payload.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+var invalidSig = errors.New("invalid signature")
+
+func (m Message) Hash() common.Hash {
+	var h common.Hash
+
+	m.Witness = nil // m is a copy and witness is not hashed.
+	hw := sha3.NewLegacyKeccak256()
+	rlp.Encode(hw, m)
```
AnnaShaleva, line 38: Don't we want to implement `rlp.Encoder` interface for proper payload encoding?

eth/protocols/dbft/payload.go
```diff
@@ -0,0 +1,53 @@
...
+	Sender common.Address
+	// Data is custom payload data.
+	Data []byte
+	// Witness is payload signature.
+	Witness []byte
+}
+
+var invalidSig = errors.New("invalid signature")
+
+func (m Message) Hash() common.Hash {
```
AnnaShaleva, line 33: Do we pay attention to the coding style? Exported method doesn't have comment, but it's OK for initial implementation.

AnnaShaleva (commented):
LGTM otherwise.

eth/protocols/dbft/payload.go
roman-khimov: I'm not sure yet we need it, it's supposed to be used for types that require custom encoding rules and we have a simple structure here.

eth/protocols/dbft/payload.go
roman-khimov: Fixed!

AnnaShaleva (approved):

PR #42, opened 2023-12-06
.github: drop testing job and reduce building job

AnnaShaleva:
We don't need testing jobs at the current stage, as not all geth tests are healthy and we're not going to fix them. We also don't need linter job, as we're not going to fix linter problems within the whole project to make it possible to merge changes from the geth repo without conflicts.


roman-khimov (approved):
Doesn't make me happy, but we don't have a lot of choice for now.

steven1227:
> Doesn't make me happy, but we don't have a lot of choice for now.

Let us do it now.

PR #43, opened 2023-12-06
Initial P2P subprotocol for dBFT

roman-khimov:


eth/backend.go
```diff
@@ -255,6 +257,8 @@ func New(stack *node.Node, config *ethconfig.Config) (*Ethereum, error) {
 	}
 	eth.APIBackend.gpo = gasprice.NewOracle(eth.APIBackend, gpoParams)
 
+	eth.dbftSrv = dbftproto.New(ethapi.NewBlockChainAPI(eth.APIBackend))
```
AnnaShaleva, line 260: Just a question, I'm not sure, but don't we need dBFT protocol for light nodes?

roman-khimov: Light node won't be able to verify message validity, it doesn't have current state with the list of CNs.

AnnaShaleva (approved):

Issue #32, opened 2023-11-23
Shut down Geth gracefully


nicolegys:
https://github.com/bane-labs/go-ethereum/blob/fd059099a26002fa053bac14a6ce42c7d671a338/Makefile#L142
The database can be corrupted if we kill the geth instance.

![image](https://github.com/bane-labs/go-ethereum/assets/43694095/9d9de5bc-b4c9-4281-b50f-d960b4845093)

We can use `SIGINT` or some other options.

AnnaShaleva:
Nice catch, I'll check it.

nicolegys:
FYI.

When using a shutdown command such as `kill -INT`, the consensus node will write all relevant information into the database.
![1700819487198](https://github.com/bane-labs/go-ethereum/assets/43694095/6c93c89c-16fd-47e5-97d8-d0ebc4881c25)

Then restart, it can load the most recent block and keep going, instead of waiting for `Inserted known block...`
![1700819971573](https://github.com/bane-labs/go-ethereum/assets/43694095/4bb05ee7-8d3b-48ce-8bc1-86419cda8410)

But I'm confused that, I use the same command to shutdown node2. However, the node2 process never ends after blockchain stopped, it just keep `looking for peers`.
![1700820706642](https://github.com/bane-labs/go-ethereum/assets/43694095/000a86b2-bb2e-4406-b41a-1590bfb48ff2)

PR #73, opened 2023-12-24
dbft: Use RLP encoding for dBFT payloads

### LINKED ISSUE #62: Use RLP encoding for dBFT payloads
After #49 all dBFT payloads use N3 style of payloads encoding or mixed RLP and N3 style. We need to migrate to RLP style of encoding. Ref. https://github.com/bane-labs/go-ethereum/pull/49#discussion_r1423117686.

This issue is not critical for the network functioning, however, #57 is related and still needs to be implemented even with RLP-style encoding.

chenquanyu:
Close #62. Use RLP encoding for dBFT payloads.

### DIFF
--- a/consensus/dbft/change_view.go
+++ b/consensus/dbft/change_view.go
@@ -2,31 +2,19 @@ package dbft
 
 import (
 	"github.com/nspcc-dev/dbft/payload"
-	"github.com/nspcc-dev/neo-go/pkg/io"
 )
 
 // changeView represents dBFT ChangeView message.
 type changeView struct {
+	// newViewNumber is not marshaled to rlp
 	newViewNumber byte
-	// timestamp is nanoseconds-precision payload timestamp, exactly like the one
+	// TimestampExt is nanoseconds-precision payload TimestampExt, exactly like the one
 	// that dBFT library operates internally with.
-	timestamp uint64
-	reason    payload.ChangeViewReason
+	TimestampExt uint64
+	ReasonExt    payload.ChangeViewReason
 }
 
 var _ payload.ChangeView = (*changeView)(nil)
-
-// EncodeBinary implements the io.Serializable interface.
-func (c *changeView) EncodeBinary(w *io.BinWriter) {
-	w.WriteU64LE(c.timestamp)
-	w.WriteB(byte(c.reason))
-}
-
-// DecodeBinary implements the io.Serializable interface.
-func (c *changeView) DecodeBinary(r *io.BinReader) {
-	c.timestamp = r.ReadU64LE()
-	c.reason = payload.ChangeViewReason(r.ReadB())
-}
 
 // NewViewNumber implements the payload.ChangeView interface.
 func (c changeView) NewViewNumber() byte { return c.newViewNumber }
@@ -35,13 +23,13 @@ func (c changeView) NewViewNumber() byte { return c.newViewNumber }
 func (c *changeView) SetNewViewNumber(view byte) { c.newViewNumber = view }
 
 // Timestamp implements the payload.ChangeView interface.
-func (c changeView) Timestamp() uint64 { return c.timestamp }
+func (c changeView) Timestamp() uint64 { return c.TimestampExt }
 
 // SetTimestamp implements the payload.ChangeView interface.
-func (c *changeView) SetTimestamp(ts uint64) { c.timestamp = ts }
+func (c *changeView) SetTimestamp(ts uint64) { c.TimestampExt = ts }
 
 // Reason implements the payload.ChangeView interface.
-func (c changeView) Reason() payload.ChangeViewReason { return c.reason }
+func (c changeView) Reason() payload.ChangeViewReason { return c.ReasonExt }
 
 // SetReason implements the payload.ChangeView interface.
-func (c *changeView) SetReason(reason payload.ChangeViewReason) { c.reason = reason }
+func (c *changeView) SetReason(reason payload.ChangeViewReason) { c.ReasonExt = reason }
--- a/consensus/dbft/payload.go
+++ b/consensus/dbft/payload.go
@@ -3,11 +3,11 @@ package dbft
 import (
 	"bytes"
 	"fmt"
+	"io"
 
 	dbftproto "github.com/ethereum/go-ethereum/eth/protocols/dbft"
 	"github.com/ethereum/go-ethereum/rlp"
 	"github.com/nspcc-dev/dbft/payload"
-	"github.com/nspcc-dev/neo-go/pkg/io"
 	"github.com/nspcc-dev/neo-go/pkg/util"
 )
 
@@ -19,8 +19,17 @@ type (
 		BlockIndex     uint64
 		ValidatorIndex byte
 		ViewNumber     byte
+		msgPayload     interface{}
+	}
 
-		payload io.Serializable
+	// messageAux is an auxiliary structure for message RLP encoding.
+	messageAux struct {
+		Type           messageType
+		BlockIndex     uint64
+		ValidatorIndex byte
+		ViewNumber     byte
+		// rlp encoded bytes of msgPayload
+		MsgPayload rlp.RawValue
 	}
 
 	// Payload is a type for consensus-related messages.
@@ -61,38 +70,42 @@ func (p *Payload) SetType(t payload.MessageType) {
 
 // Payload implements the payload.ConsensusPayload interface.
 func (p Payload) Payload() any {
-	return p.payload
+	return p.msgPayload
 }
 
 // SetPayload implements the payload.ConsensusPayload interface.
 func (p *Payload) SetPayload(pl any) {
-	p.payload = pl.(io.Serializable)
+	p.msgPayload = pl
 }
 
 // GetChangeView implements the payload.ConsensusPayload interface.
-func (p Payload) GetChangeView() payload.ChangeView { return p.payload.(payload.ChangeView) }
+func (p Payload) GetChangeView() payload.ChangeView {
+	return p.msgPayload.(*changeView)
+}
 
 // GetPrepareRequest implements the payload.ConsensusPayload interface.
 func (p Payload) GetPrepareRequest() payload.PrepareRequest {
-	return p.payload.(payload.PrepareRequest)
+	return p.msgPayload.(*prepareRequest)
 }
 
 // GetPrepareResponse implements the payload.ConsensusPayload interface.
 func (p Payload) GetPrepareResponse() payload.PrepareResponse {
-	return p.payload.(payload.PrepareResponse)
+	return p.msgPayload.(*prepareResponse)
 }
 
 // GetCommit implements the payload.ConsensusPayload interface.
-func (p Payload) GetCommit() payload.Commit { return p.payload.(payload.Commit) }
+func (p Payload) GetCommit() payload.Commit {
+	return p.msgPayload.(*commit)
+}
 
 // GetRecoveryRequest implements the payload.ConsensusPayload interface.
 func (p Payload) GetRecoveryRequest() payload.RecoveryRequest {
-	return p.payload.(payload.RecoveryRequest)
+	return p.msgPayload.(*recoveryRequest)
 }
 
 // GetRecoveryMessage implements the payload.ConsensusPayload interface.
 func (p Payload) GetRecoveryMessage() payload.RecoveryMessage {
-	return p.payload.(payload.RecoveryMessage)
+	return p.msgPayload.(*recoveryMessage)
 }
 
 // ValidatorIndex implements the payload.ConsensusPayload interface.
@@ -156,45 +169,6 @@ func (p *Payload) Hash() util.Uint256 {
 	return p.Message.Hash().Uint256()
 }
 
-// EncodeBinary implements the io.Serializable interface.
-func (m *message) EncodeBinary(w *io.BinWriter) {
-	w.WriteB(byte(m.Type))
-	w.WriteU64LE(m.BlockIndex)
-	w.WriteB(m.ValidatorIndex)
-	w.WriteB(m.ViewNumber)
-	m.payload.EncodeBinary(w)
-}
-
-// DecodeBinary implements the io.Serializable interface.
-func (m *message) DecodeBinary(r *io.BinReader) {
-	m.Type = messageType(r.ReadB())
-	m.BlockIndex = r.ReadU64LE()
-	m.ValidatorIndex = r.ReadB()
-	m.ViewNumber = r.ReadB()
-
-	switch m.Type {
-	case changeViewType:
-		cv := new(changeView)
-		// newViewNumber is not marshaled
-		cv.newViewNumber = m.ViewNumber + 1
-		m.payload = cv
-	case prepareRequestType:
-		m.payload = new(prepareRequest)
-	case prepareResponseType:
-		m.payload = new(prepareResponse)
-	case commitType:
-		m.payload = new(commit)
-	case recoveryRequestType:
-		m.payload = new(recoveryRequest)
-	case recoveryMessageType:
-		m.payload = new(recoveryMessage)
-	default:
-		r.Err = fmt.Errorf("invalid type: 0x%02x", byte(m.Type))
-		return
-	}
-	m.payload.DecodeBinary(r)
-}
-
 // String implements fmt.Stringer interface.
 func (t messageType) String() string {
 	switch t {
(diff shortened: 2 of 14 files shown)


Makefile
```diff
@@ -178,11 +178,11 @@ privnet_init_four: privnet_clean
 
 privnet_nodes_stop:
 	@echo "Killing nodes processes"
-	@killall -w -v -9 geth || :
+	@killall -v -9 geth || :
 
 privnet_bootnode_stop:
 	@echo "Killing bootnode processes"
-	@killall -w -v -9 bootnode || :
+	@killall -v -9 bootnode || :
```
AnnaShaleva: Not related to #62. Should be removed or implemented in a separate PR wrt comments of #32.

consensus/dbft/change_view.go
```diff
@@ -1,47 +1,48 @@
...
+
+// EncodeRLP serializes a changeView as RLP.
+func (obj *changeView) EncodeRLP(_w io.Writer) error {
+	w := rlp.NewEncoderBuffer(_w)
+	_tmp0 := w.List()
+	w.WriteUint64(obj.TimestampNano)
+	w.WriteUint64(uint64(obj.ReasonExt))
+	w.ListEnd(_tmp0)
+	return w.Flush()
+}
```
AnnaShaleva: Do we really need to implement this method on `changeView`?  Can't we rely on default RLP encoding? Because `changeView` itself doesn't have any "special" fields, all exported fields should be serialized without any additional rules.

consensus/dbft/change_view.go
```diff
@@ -1,47 +1,48 @@
...
 type changeView struct {
+	// newViewNumber is not marshaled to rlp
 	newViewNumber byte
-	// timestamp is nanoseconds-precision payload timestamp, exactly like the one
+	// TimestampNano is nanoseconds-precision payload TimestampNano, exactly like the one
 	// that dBFT library operates internally with.
-	timestamp uint64
-	reason    payload.ChangeViewReason
+	TimestampNano uint64
+	ReasonExt     payload.ChangeViewReason
```
AnnaShaleva: Let's keep them unified, s/`TimestampNano`/`TimestampExt`.

consensus/dbft/payload.go
```diff
@@ -20,7 +20,8 @@ type (
 		ValidatorIndex byte
 		ViewNumber     byte
 
-		payload io.Serializable
+		MsgPayload []byte
+		// payload io.Serializable
```
AnnaShaleva: Remove this line.

consensus/dbft/change_view.go
```diff
@@ -1,47 +1,48 @@
...
 // Reason implements the payload.ChangeView interface.
-func (c changeView) Reason() payload.ChangeViewReason { return c.reason }
+func (c changeView) Reason() payload.ChangeViewReason { return c.ReasonExt }
 
 // SetReason implements the payload.ChangeView interface.
-func (c *changeView) SetReason(reason payload.ChangeViewReason) { c.reason = reason }
+func (c *changeView) SetReason(reason payload.ChangeViewReason) { c.ReasonExt = reason }
+
+// EncodeRLP serializes a changeView as RLP.
+func (obj *changeView) EncodeRLP(_w io.Writer) error {
```
roman-khimov: Can you avoid a custom encoder? Or is it an optimization?

consensus/dbft/payload.go
```diff
@@ -61,38 +62,59 @@ func (p *Payload) SetType(t payload.MessageType) {
...
 }
 
 // GetChangeView implements the payload.ConsensusPayload interface.
-func (p Payload) GetChangeView() payload.ChangeView { return p.payload.(payload.ChangeView) }
+func (p Payload) GetChangeView() payload.ChangeView {
+	obj := &changeView{}
+	rlp.DecodeBytes(p.MsgPayload, obj)
+	obj.newViewNumber = p.message.ViewNumber + 1
+	return obj
+}
```
AnnaShaleva: I'm a little worried about this approach, because: dBFT may call this method a lot of times even per a single view. And I doubt we need to decode it each time. Thus, I'd suggest to have some kind of cache with deserialized payload, exactly like it was done before. Is it possible?

consensus/dbft/payload.go
```diff
@@ -61,38 +62,59 @@ func (p *Payload) SetType(t payload.MessageType) {
 
 // Payload implements the payload.ConsensusPayload interface.
 func (p Payload) Payload() any {
-	return p.payload
+	return p.MsgPayload
```
AnnaShaleva: And then here we need to return this deserialized payload (not bytes).

consensus/dbft/change_view.go
chenquanyu: > Can you avoid a custom encoder? Or is it an optimization?

OK, I will remove these redundant custom encoders.

consensus/dbft/payload.go
chenquanyu: For RLP encode and decode, we need a struct type of feild. I have tried interface with rlp.Encoder and rlp.Decoder, but failed.

AnnaShaleva: We can't keep this code, we need a cache for it.

> we need a struct type of feild

We may keep `MsgPayload []byte` field and then extend `message` with additional unexported fields: `changeView`, `prepareRequest`, etc. Once `p.GetChangeView` or `p.GetPrepareRequest` or etc. method is called, we may check if `p.changeView` is nil, and if it is, then deseialize it (once!) and set `p.changeView` to deserialized change view payload. And if `p.changeView` is not nil, then just return it.

AnnaShaleva (commented):
Please, squash all commits, so that a single commit corresponds to a single logical contribution. Also, it would be nice to add `Close #62.` message to this commit and to the PR description.

consensus/dbft/payload.go
AnnaShaleva: BTW, how about reusing `rlp.RawValue` type for `MsgPayload`?

chenquanyu: Used ```msgCache atomic.Value``` to cache `MsgPayload` and change `MsgPayload`  to `rlp.RawValue`.

consensus/dbft/payload.go
```diff
@@ -61,7 +64,29 @@ func (p *Payload) SetType(t payload.MessageType) {
...
+		return p.GetPrepareRequest()
+	case prepareResponseType:
+		return p.GetPrepareResponse()
+	case commitType:
+		return p.GetCommit()
+	case recoveryRequestType:
+		return p.GetPrepareRequest()
+	case recoveryMessageType:
+		return p.GetRecoveryMessage()
+
```
AnnaShaleva: Useless empty line.

consensus/dbft/payload.go
```diff
@@ -20,7 +21,9 @@ type (
 		ValidatorIndex byte
 		ViewNumber     byte
 
-		MsgPayload []byte
+		MsgPayload rlp.RawValue
+		// MsgPayload cache
+		msgCache atomic.Value
```
AnnaShaleva: What is the reason for using atomic value? dBFT is a single-thread process completely managed by `eventLoop`, no concurrent access to `msgCache` could happen.

AnnaShaleva: s/msgCache/msgPayloadCache?

consensus/dbft/prepare_request.go
```diff
@@ -1,13 +1,10 @@
...
 
 import (
-	"fmt"
-
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
-	"github.com/ethereum/go-ethereum/rlp"
 	"github.com/nspcc-dev/dbft/payload"
-	"github.com/nspcc-dev/neo-go/pkg/io"
+
```
AnnaShaleva: Useless empty line.

consensus/dbft/prepare_response.go
```diff
@@ -2,29 +2,19 @@ package dbft
 
 import (
 	"github.com/nspcc-dev/dbft/payload"
-	"github.com/nspcc-dev/neo-go/pkg/io"
+
```
AnnaShaleva: Ditto.

consensus/dbft/recovery_message.go
```diff
@@ -268,25 +179,31 @@ func (m *recoveryMessage) GetCommits(p payload.ConsensusPayload, validators []cr
...
+	if err != nil {
+		panic(fmt.Errorf("fromPayload rlp encode error: %w", err))
+	}
+
 	return &Payload{
 		Message: dbftproto.Message{
 			ValidBlockEnd: recovery.BlockIndex,
 			Sender:        recovery.Sender,
 		},
+
```
AnnaShaleva: Useless change.

consensus/dbft/recovery_message.go
```diff
@@ -268,25 +179,31 @@ func (m *recoveryMessage) GetCommits(p payload.ConsensusPayload, validators []cr
...
-	m.preparationHash = h
+	m.PreparationHashExt = h
 }
 
-func fromPayload(t messageType, recovery *Payload, p io.Serializable) *Payload {
+func fromPayload(t messageType, recovery *Payload, p any) *Payload {
+	bytes, err := rlp.EncodeToBytes(p)
+	if err != nil {
+		panic(fmt.Errorf("fromPayload rlp encode error: %w", err))
+	}
```
AnnaShaleva: This pattern is used very often. We need to replace it with the following logic: set `msgPayloadCache` once where needed, and after this on serialisation we need to check if `MsgPayload` is empty, and if it is, then serialize `msgPayloadCache` and put the result into `MsgPayload`.

consensus/dbft/recovery_message.go
```diff
@@ -200,7 +111,7 @@ func (m *recoveryMessage) GetPrepareRequest(p payload.ConsensusPayload, validato
 		return nil
 	}
 
-	req := fromPayload(prepareRequestType, p.(*Payload), m.prepareRequest.payload)
+	req := fromPayload(prepareRequestType, p.(*Payload), m.PrepareRequest.MsgPayload)
```
AnnaShaleva: Shouldn't it be `m.PrepareRequest.msgPayloadCache` then?

consensus/dbft/payload.go
```diff
@@ -61,38 +64,112 @@ func (p *Payload) SetType(t payload.MessageType) {
...
+	if msg := p.msgCache.Load(); msg != nil {
+		return msg.(*changeView)
+	}
+
+	obj := &changeView{}
+	rlp.DecodeBytes(p.MsgPayload, obj)
+	obj.newViewNumber = p.message.ViewNumber + 1
+	p.msgCache.Store(obj)
+	return obj
+}
```
AnnaShaleva: OK, and now I see that decoding errors are not handled in any way. Could we create a special RLP decoder/encoder for `message` structure so that `MsgPayload` decoding happens immediately after `message` decoding instead of during the first call `GetChangeView` or other getters? And then in this custom RLP `message` decoder we can properly set `msgPayloadCache`.

And in the custom `message` RLP encoder we can serialize `msgPayloadCache` into `MsgPayload` before further standard RLP encoding (if `MsgPayload` is empty).

AnnaShaleva (commented):
Please, organize commits wrt https://github.com/bane-labs/go-ethereum/pull/85#discussion_r1445707505 and https://github.com/nspcc-dev/.github/blob/master/git.md#commits.

consensus/dbft/recovery_message.go
chenquanyu: Use custom RLP decoder/encoder to avoid saving encoded bytes in message.

consensus/dbft/payload.go
chenquanyu: Custom RLP decoder/encoder with `messageExt` type now.

consensus/dbft/payload.go
```diff
@@ -215,25 +190,71 @@ func (t messageType) String() string {
...
+		obj = &prepareResponse{}
+	case commitType:
+		obj = &commit{}
+	case recoveryRequestType:
+		obj = &recoveryRequest{}
+	case recoveryMessageType:
+		obj = &recoveryMessage{}
+	default:
+		err := fmt.Errorf("invalid type: 0x%02x", byte(m.Type))
+		log.Warn(err.Error())
```
AnnaShaleva: Do we need a warning here? This error will be returned to RLP decoder and it will throw this error anyway, won't it?

consensus/dbft/payload.go
```diff
@@ -215,25 +190,71 @@ func (t messageType) String() string {
...
+		return err
+	}
+	if err := rlp.DecodeBytes(em.MsgPayload, obj); err != nil {
+		return err
+	}
+	if m.Type == changeViewType {
+		obj.(*changeView).newViewNumber = m.ViewNumber + 1
```
AnnaShaleva: Let's preserve the original comment: `// newViewNumber is not marshaled`.

consensus/dbft/payload.go
```diff
@@ -215,25 +190,71 @@ func (t messageType) String() string {
...
+	}
+	m.msgPayload = obj
+	return nil
```
AnnaShaleva: Will the following code work? Less change comparing to the previous non-rlp version.
```suggestion
	switch m.Type {
	case changeViewType:
		m.msgPayload = &changeView{
			// newViewNumber is not marshaled
			newViewNumber: m.ViewNumber + 1,
		}
	case prepareRequestType:
		m.msgPayload = new(prepareRequest)
	case prepareResponseType:
		m.msgPayload = new(prepareResponse)
	case commitType:
		m.msgPayload = &commit{}
	case recoveryRequestType:
		m.msgPayload = new(recoveryRequest)
	case recoveryMessageType:
		m.msgPayload = new(recoveryMessage)
	default:
		err := fmt.Errorf("invalid type: 0x%02x", byte(m.Type))
		return err
	}
	return rlp.DecodeBytes(em.MsgPayload, m.msgPayload)
```

consensus/dbft/recovery_message.go
```diff
@@ -43,153 +40,59 @@ type (
...
 
 	switch p.Type() {
 	case payload.PrepareRequestType:
-		m.prepareRequest = &message{
+		pr := p.GetPrepareRequest().(*prepareRequest)
+		m.PrepareRequest = &message{
 			Type:       prepareRequestType,
 			ViewNumber: p.ViewNumber(),
-			payload:    p.GetPrepareRequest().(*prepareRequest),
+			msgPayload: pr,
```
AnnaShaleva: Less changes in code is better:
```suggestion
			msgPayload: p.GetPrepareRequest().(*prepareRequest),
```

consensus/dbft/recovery_message.go
```diff
@@ -43,153 +40,59 @@ type (
 
 var _ payload.RecoveryMessage = (*recoveryMessage)(nil)
 
-// DecodeBinary implements the io.Serializable interface.
-func (m *recoveryMessage) DecodeBinary(r *io.BinReader) {
```
AnnaShaleva, line 46: It's not that easy with `recoveryMessage`, because Recoveries may not contain some fields. For example, current decoding code leads to inability to decode some Recovery messages on node start with the following error:
```
INFO [01-16|15:47:47.257] can't decode payload data                hash=061b1e..653181 error="rlp: input string too short for util.Uint256, decoding into (dbft.message)(dbft.recoveryMessage).PreparationHashExt"

INFO [01-16|15:47:47.261] can't decode payload data                hash=635eb9..fe0054 error="rlp: input string too short for util.Uint256, decoding into (dbft.message)(dbft.recoveryMessage).PreparationHashExt"
```

It happens because this recovery likely doesn't contain `PreparationHashExt` field because no PrepareRequest was sent to the network. And it's perfectly OK to have missing (or empty) `PreparationHashExt`, it should be allowed.

Thus, please, pay attention to the previous non-RLP decoder/encoder of `recoveryMessage`. We either need some RLP tags to mark fields of `recoveryMessage` as optional or we need a custom RLP encoder/decoder for it.

AnnaShaleva (commented):
Good! Now the encoding/decoding behaviour is very close to the desired one.

@chenquanyu, the only issue is left (proper Recovery messages encoding/decoding), let's fix it. We're going to include this PR to NeoX T2, since this PR is almost ready.

AnnaShaleva:
Also please, rebase onto https://github.com/bane-labs/go-ethereum/pull/99.

consensus/dbft/recovery_message.go
chenquanyu: Add `rlp:"optional"` for `PreparationHashExt`  and `PrepareRequest` .

consensus/dbft/recovery_message.go
```diff
@@ -1,23 +1,20 @@
...
-		preparationHash     *util.Uint256
-		preparationPayloads []*preparationCompact
-		commitPayloads      []*commitCompact
-		changeViewPayloads  []*changeViewCompact
-		prepareRequest      *message
+		PreparationPayloads []*preparationCompact
+		CommitPayloads      []*commitCompact
+		ChangeViewPayloads  []*changeViewCompact
+		PreparationHashExt  *util.Uint256 `rlp:"optional"`
+		PrepareRequest      *message      `rlp:"optional"`
```
AnnaShaleva: After a bit of testing I think that custom RLP encoder/decoder is still needed for `recoveryMessage` because current implementation missing verification checks. I'd suggest the following: create auxiliary structure `recoveryMessageAux` for recovery RLP marshalling, remove `rlp` tags from the original `recoveryMessage` and then implement a custom RLP encoder/decoder for `recoveryMessage` (decoder should perform all necessary verification checks):
```go
type (
	// recoveryMessage represents dBFT Recovery message.
	recoveryMessage struct {
		PreparationPayloads []*preparationCompact
		CommitPayloads      []*commitCompact
		ChangeViewPayloads  []*changeViewCompact
		PreparationHashExt  *util.Uint256 `rlp:"optional"`
		PrepareRequest      *message      `rlp:"optional"`
	}

	// recoveryMessageAux is an auxiliary structure for recoveryMessage RLP encoding.
	recoveryMessageAux struct {
		PreparationPayloads []*preparationCompact
		CommitPayloads      []*commitCompact
		ChangeViewPayloads  []*changeViewCompact
		PreparationHashExt  *util.Uint256 `rlp:"optional"`
		PrepareRequest      *message      `rlp:"optional"`
	}
)


// EncodeRLP serializes recoveryMessage as RLP.
func (m *recoveryMessage) EncodeRLP(w io.Writer) error {
	return rlp.Encode(w, &recoveryMessageAux{
		PreparationPayloads: m.PreparationPayloads,
		CommitPayloads:      m.CommitPayloads,
		ChangeViewPayloads:  m.ChangeViewPayloads,
		PreparationHashExt:  m.PreparationHashExt,
		PrepareRequest:      m.PrepareRequest,
	})
}

// DecodeRLP decodes recoveryMessage from RLP.
func (m *recoveryMessage) DecodeRLP(s *rlp.Stream) error {
	var aux recoveryMessageAux
	if err := s.Decode(&aux); err != nil {
		return err
	}

	// Perform some validity checks.
	if aux.PrepareRequest != nil && aux.PrepareRequest.Type != prepareRequestType {
		return errors.New("recovery message PrepareRequest has wrong type")
	}

	m.PreparationPayloads = aux.PreparationPayloads
	m.CommitPayloads = aux.CommitPayloads
	m.ChangeViewPayloads = aux.ChangeViewPayloads
	m.PreparationHashExt = aux.PreparationHashExt
	m.PrepareRequest = aux.PrepareRequest
	return nil
}
```

consensus/dbft/payload.go
```diff
@@ -19,8 +19,17 @@ type (
 		BlockIndex     uint64
 		ValidatorIndex byte
 		ViewNumber     byte
+		msgPayload     interface{}
+	}
 
-		payload io.Serializable
+	// "external" message rlp encoding type.
+	messageExt struct {
```
AnnaShaleva: Let's rename it to have common code style:
```suggestion
	// messageAux is an auxiliary structure for message RLP encoding.
	messageAux struct {
```

AnnaShaleva (commented):
Otherwise LGTM.

AnnaShaleva (approved):
Good!

AnnaShaleva:
@roman-khimov, @nicolegys, agree to merge? I did some basic tests, but maybe you would also like to test the final version.

nicolegys:
Let me make some tests.

roman-khimov (approved):

nicolegys (approved):

PR #79, opened 2023-12-26
dbft: do not rely on Seal during dBFT context reset

AnnaShaleva:
Close #71. Without this PR transactions can't be properly accepted to the network due to often sealing tasks submission.

This PR also fixes a minor issue that prevents node from starting with an existing DB and some known blocks that are not yet stored in that DB, see the logs below. This problem is described in the logs of https://github.com/bane-labs/go-ethereum/issues/32#issue-2007676339 when the node is trying to recover after ungraceful shutdown.
<details>
  <summary>Node log with problem situation reproduced, before the fix</summary>
  
```
INFO [12-26|19:08:38.064] --------------------------------------------------------------------------------------------------------------------------------------------------------- 
INFO [12-26|19:08:38.064]  
INFO [12-26|19:08:38.065] Loaded most recent local block           number=135 hash=abd23b..568191 td=265 age=12m29s
WARN [12-26|19:08:38.065] Head state missing, repairing            number=135 hash=abd23b..568191 snaproot=7a7785..9749a9
INFO [12-26|19:08:38.078] Loaded most recent local header          number=135 hash=abd23b..568191 td=265 age=12m29s
INFO [12-26|19:08:38.078] Loaded most recent local block           number=13  hash=ba16dd..e1412d td=26  age=50m
INFO [12-26|19:08:38.078] Loaded most recent local snap block      number=135 hash=abd23b..568191 td=265 age=12m29s
WARN [12-26|19:08:38.100] Enabling snapshot recovery               chainhead=13 diskbase=13
WARN [12-26|19:08:38.100] Loaded snapshot journal                  diffs=missing
INFO [12-26|19:08:38.101] Initialized transaction indexer          limit=2,350,000
INFO [12-26|19:08:38.101] Loaded local transaction journal         transactions=0 dropped=0
INFO [12-26|19:08:38.101] Regenerated local transaction journal    transactions=0 accounts=0
WARN [12-26|19:08:38.109] Switch sync mode from snap sync to full sync 
INFO [12-26|19:08:38.109] Gasprice oracle is ignoring threshold set threshold=2
WARN [12-26|19:08:38.110] Unclean shutdown detected                booted=2023-12-26T18:54:51+0300 age=13m47s
WARN [12-26|19:08:38.111] Engine API enabled                       protocol=eth
WARN [12-26|19:08:38.111] Engine API started but chain not configured for merge yet 
[...]

PR #85, opened 2024-01-03
dbft: Shut down Geth gracefully

### LINKED ISSUE #32: Shut down Geth gracefully

chenquanyu:
Close #32 Shut down Geth gracefully. This issue is caused by the DBFT ```Close()``` method reads channel from not started ```eventLoop()```.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -222,7 +222,7 @@ privnet_init_seven: privnet_clean
 
 privnet_nodes_stop:
 	@echo "Killing nodes processes"
-	@killall -w -v -9 geth || :
+	@killall -w -v -INT geth || :
 
 privnet_bootnode_stop:
 	@echo "Killing bootnode processes"
--- a/consensus/dbft/dbft.go
+++ b/consensus/dbft/dbft.go
@@ -1456,8 +1456,10 @@ func encodeSigHeader(w io.Writer, header *types.Header) {
 
 // Close implements consensus.Engine.
 func (c *DBFT) Close() error {
-	close(c.quit)
-	<-c.finished
+	if c.dbftStarted.Load() {
+		close(c.quit)
+		<-c.finished
+	}
 	return nil
 }
 


nicolegys:
Works well, except that bootnode process cannot be killed by `kill -2`

chenquanyu:
> Works well, except that bootnode process cannot be killed by `kill -2`
> 
> 

The ```bootnode``` is only a dev tool of geth, and the kill signal is not handled in this tool. We will use bootstrap node instead of ```bootnode``` in test and prod environment.

nicolegys:
OK for me.

Makefile
```diff
@@ -222,11 +222,11 @@ privnet_init_seven: privnet_clean
 
 privnet_nodes_stop:
 	@echo "Killing nodes processes"
-	@killall -w -v -9 geth || :
+	@killall -v -2 geth || :
```
roman-khimov: I'd rather do `-INT` here and keep `-w` to have nodes stopped after the command. This fix is not related to `consensus/dbft` changes anyway, can you make it a separate commit?

chenquanyu: > I'd rather do `-INT` here and keep `-w` to have nodes stopped after the command. This fix is not related to `consensus/dbft` changes anyway, can you make it a separate commit?

We can use ```killall -INT``` here and I removed the ```-w``` option because it is not compatible in Mac OS. I modify make file in this PR because ```killall -INT``` will work after this fix of issue #32 .

roman-khimov: `-w` can be emulated with a loop (`killall` returns 1 when it doesn't actually send any signal), the problem is predictable behavior after `make privnet_nodes_stop`, without `-w` you have to check for the status somehow to ensure nodes have stopped.

PR is OK, I'm more about commit structure. I know NGD uses squashing for PRs as a policy, but I'd really prefer to see something like #79 with each commit solving exactly one problem and overall PR being merged unsquashed. This leaves much more relevant details about changes for the future. Not critical here, the PR is rather small, but notice for example our `-w` discussion, nothing in commits says that it's a problem for Mac. In ideal world I'd do `git log` and see this change as a separate commit with an explanation of _why_ it is needed. So, just a suggestion.

chenquanyu: OK, I will fix ```-w``` in a separate commit.

AnnaShaleva (commented):
Please, organize commits wrt https://github.com/bane-labs/go-ethereum/pull/85#discussion_r1445707505. So in this PR we'd expect to see two commits: the first one is Makefile change with appropriate commit message and commit body containing the reason of changes and `Close #85.` line. And the second one is `consensus/dbft/dbft.go` fix with appropriate commit message.

PR content itself LGTM.

AnnaShaleva:
@chenquanyu, see also https://github.com/nspcc-dev/.github/blob/master/git.md#commits, please.

AnnaShaleva (commented):
Sorry for nitpicking, but please, split into two commits like described in https://github.com/bane-labs/go-ethereum/pull/85#pullrequestreview-1813583676:
The first commit's header/message is expected to be like that:
```
Makefile: shut down Geth privnet gracefully

Close #32.
```
The second one is something like this:
```
dbft: properly shutdown DBFT engine

Close `quit` channel only if DBFT engine was started.
```

AnnaShaleva (approved):

PR #93, opened 2024-01-12
dBFT: get rid of snapshot dependency

### LINKED ISSUE #51: Improve dBFT snapshot mechanism
Current version of dBFT snapshot (that we have since #49) has a lot of obsolete things inherited from Clique consensus. We need to carefully review and evaluate the necessity of every snapshot field and likely to remove some unnecessary logic, e.g.:
* drop any votes related logic (no snapshot-based votes caching, no votes tally, etc);
* review and likely drop or refactor coinbase snapshot managing, because coinbase itself connected with primary;
* refactor snapshot signers managing wrt that fact that we have block validators specified in the block itself and nextBlockValidators calculation function available for every possible block height;
* likely there's something else we can improve in the current dBFT snapshot implementation

### LINKED ISSUE #58: Proper header difficulty initialization and verification
After #49 verification of header difficulty is disabled (ref. 577ef773b9e9151e935e935b2efe00e0687d4fbc). It is done because dBFT has its own mechanism of primary node selections and in case of change views, primary node may change from in-turn node to out-of-turn node. Currently, header difficulty is being initialized from the sealing miner proposal based on the snapshot set of signers.

What we need to do within the scope of this issue is:
1. Initialize header difficulty based on the dBFT's context Primary and a set of validators got from dBFT's context so that every PrepareRequest has correct difficulty set (`diffInTurn` for Primary that is a primary of the zero view (and each N-h subsequent view) for the current height and `diffNoTurn` for the rest of Primaries in case of CVs). This can be properly done only after #54.
2. Enable header difficulty verification disabled in 577ef773b9e9151e935e935b2efe00e0687d4fbc of #49.
3. Adjust header difficulty verification rules so that they are based not on the snapshot's signers indexes, but on the index of validators retrieved from the block's witness (see `(s *Snapshot) inturn` method).

This issue is roughly depends on #54.

### LINKED ISSUE #86: The block miner is always zero address.
Now the block miner is always zero address. 

image

AnnaShaleva:
Close #86. Close #51. Close #58.

- [x] Decide about Header's Coinbase meaning.
- [x] Implement Coinbase verification.
- [x] Remove the rest of snapshot mentions.
- [x] Ensure that code works when unprocessed blocks are in the chain.
- [x] Refactor dBFT API which is closely based on snapshot.
- [x] Format code and improve comments.


consensus/dbft/dbft.go
```diff
@@ -411,6 +407,15 @@ func New(config *params.DBFTConfig, db ethdb.Database) (*DBFT, error) {
 			// PrimaryIndex -> Nonce
 			binary.BigEndian.PutUint64(h.Nonce[:], uint64(ctx.PrimaryIndex))
 
+			// PrimaryAddress -> Coinbase
+			// TODO: @roman-khimov, @chenquanyu we need to decide about this: do we need a
+			// custom address corresponding to every consensus node to be set as Coinbase
+			// or it's OK to have the consensus node address itself be set as Coinbase?
+			// Currently, the second option is implemented.
+			h.Coinbase = ctx.Validators[ctx.PrimaryIndex].(*PublicKey).Account
```
AnnaShaleva: @roman-khimov, @chenquanyu, need your opinion on this topic. We need to decide about Primary and Coinbase relation.

consensus/dbft/dbft.go
```diff
@@ -629,13 +634,29 @@ func (c *DBFT) postBlock(b *types.Block) {
...
+	//    will provide this. Given this fact, Coinbase is always the address of Primary
+	//    that was the block's author or the accepted block at the current height. And
+	//    then Coinbase then can be used for reward calculations.
+	// 2. If Coinbase is expected to have another meaning (i.e. it's expected to be
+	//    some side address), then we must share the rules of Coinbase detection between
+	//    consensus nodes. In this case we need to decide what's the desired meaning of Coinbase.
+	//
+	// Currently option 1 is implemented, but without enforced miner.etherbase check. Please,
+	// write your thoughts on this.
+	return header.Coinbase, nil
```
AnnaShaleva: And this one is also closely related to the same Coinbase-Primary problem.

consensus/dbft/dbft.go
chenquanyu: > @roman-khimov, @chenquanyu, need your opinion on this topic. We need to decide about Primary and Coinbase relation.

I think they should be the same, they are the miner of block.

AnnaShaleva:
@roman-khimov, ready for review. Coinbase topic discussion is moved to Discord, the second option (customizable Coinbase via `miner.etherbase`) is implemented in this PR.

consensus/dbft/dbft.go
```diff
@@ -771,94 +739,12 @@ func (c *DBFT) verifyCascadingFields(chain consensus.ChainHeaderReader, header *
...
-		headers[i], headers[len(headers)-1-i] = headers[len(headers)-1-i], headers[i]
-	}
-	snap, err := snap.apply(headers)
-	if err != nil {
-		return nil, err
-	}
-	c.recents.Add(snap.Hash, snap)
+	// TODO: ensure Coinbase is valid. If Coinbase is a Governance (or some other) contract, then ensure that
+	// Coinbase matches the contract address. If it's a validator address, then ensure Coinbase matches the
+	// Speaker address for the corresponding PrimaryIndex.
```
AnnaShaleva: Once we decide on Coinbase meaning, I'll add Coinbase verification commit to this PR.

nicolegys:
Better not return error when getSigners of genesis block

AnnaShaleva:
> Better not return error when getSigners of genesis block

@nicolegys, fixed! Could you also make tests for other APIs and Coinbase/Difficulty fields of accepted blocks? Also, maybe you have some thoughts about RPC API extension (ref. https://github.com/bane-labs/go-ethereum/issues/96)?

nicolegys:
> @nicolegys, fixed! Could you also make tests for other APIs and Coinbase/Difficulty fields of accepted blocks? Also, maybe you have some thoughts about RPC API extension (ref. #96)?

Of course! Let me make some tests and then think about extension.

nicolegys:
Wonder if it is appropriate to return the `index 0` validator as a primary of genesis block.

**getPrimary**


**getValidators**



Since miner of `block 0` is zero address.

AnnaShaleva:
> Wonder if it is appropriate to return the index 0 validator as a primary of genesis block.

It's appropriate because the PrimaryIndex of Genesis block is 0, and thus, the 0-th validator from the sorted list of validators should be returned. And for block 1 it's expected to be 1-th validator address to be a primary (if no CVs happens during block 1 acceptance). Also, Coinbase in general shouldn't match the Primary since we've decide that Coinbase is a customizable value that can be set via `miner.etherbase` for every consensus node. It's supposed to be set to Governance contract address in future.

> Since miner of block 0 is zero address

But this is something that may be improved, we need to decide about it. We may set Genesis Coinbase to some address (Governance contract as we've agreed), but then we have to provide it via node's configuration. And then likely to validate `miner.etherbase` against Genesis Coinbase (two values should match). But then we won't be able to "change" `miner.etherbase` in the existing running network without hardfork or some node/dBFT setting extension. From the other hand, we may leave Genesis Coinbase empty which also won't hurt much.

@roman-khimov, what do you think?

roman-khimov:
We have to have some primary index for all blocks and zero is always some valid index -> we have zero in genesis, it' OK. See N3 mainnet as well (where primary 0 had not produced this block as well):
```
{
   "result" : {
      "confirmations" : 4744189,
      "version" : 0,
      "merkleroot" : "0x0000000000000000000000000000000000000000000000000000000000000000",
      "time" : 1468595301000,
      "nextblockhash" : "0x5f3fbb43d1e516fe07771e2e873ebc9e2810662401acf603a775aace486220bd",
      "hash" : "0x1f4d1defa46faa5e7b9b8d3f79a06bec777d7c26c4aa5f6f5899a291daa87c15",
      "witnesses" : [
         {
            "invocation" : "",
            "verification" : "EQ=="
         }
      ],
      "nonce" : "000000007C2BAC1D",
      "primary" : 0,
      "index" : 0,
      "size" : 114,
      "previousblockhash" : "0x0000000000000000000000000000000000000000000000000000000000000000",
      "tx" : [],
      "nextconsensus" : "NVg7LjGcUSrgxgjX3zEgqaksfMaiS8Z6e1"
   },
   "id" : 5,
   "jsonrpc" : "2.0"
}
```

There is no real miner for block 0 and there are no fees to distribute -> zero miner address is fine.

consensus/dbft/dbft.go
```diff
@@ -81,6 +81,7 @@ var (
 
 	uncleHash = types.CalcUncleHash(nil) // Always Keccak256(RLP([])) as uncles are meaningless outside of PoW.
 
+	diffStub   = big.NewInt(3) // Block difficulty stub that is used for miner's task preparation.
```
roman-khimov, line 70: Shouldn't it be zero like in PoS engines? Although Clique uses 2 from what I see...

consensus/dbft/dbft.go
roman-khimov: Looks like we have an agreement on this.

consensus/dbft/dbft.go
AnnaShaleva: If it's zero, then sealing tasks will be rejected by Beacon chain.

roman-khimov: OK.

consensus/dbft/dbft.go
```diff
@@ -753,9 +756,10 @@ func (c *DBFT) verifyCascadingFields(chain consensus.ChainHeaderReader, header *
 		// Verify the header's EIP-1559 attributes.
 		return err
 	}
-	// TODO: ensure Coinbase is valid. If Coinbase is a Governance (or some other) contract, then ensure that
-	// Coinbase matches the contract address. If it's a validator address, then ensure Coinbase matches the
-	// Speaker address for the corresponding PrimaryIndex.
+	// This check forces Coinbase to be the same among all consensus nodes.
+	if header.Coinbase != c.config.Coinbase {
+		return fmt.Errorf("invalid Coinbase: expected %s, got %s", c.config.Coinbase, header.Coinbase)
+	}
```
AnnaShaleva: @roman-khimov, @chenquanyu let's double-check, are we OK with this change? With this code we can't use different Coinbase addresses for different CNs anymore. This address should be shared between all CNs and should be the same within the whole network lifetime.

privnet/four/genesis_template.json
```diff
@@ -21,7 +21,8 @@
         "0xnode2",
         "0xnode3",
         "0xnode4"
-      ]
+      ],
+      "coinbase": "0x1212100000000000000000000000000000000001"
```
AnnaShaleva, line 25: @nicolegys, please, check one more time, now the following problem should be fixed:
```
Now the code requires miner.etherbase to be a local account to sign, so it can't be an external account (like a contract address). Need some changes if we choose to set coinbase by configure miner.etherbase. 
```

consensus/dbft/dbft.go

consensus/dbft/dbft.go
roman-khimov: > we can't use different Coinbase addresses for different CNs anymore

It's OK if we're managing fee distribution via a contract.

> This address should be shared between all CN

Sure.

> should be the same within the whole network lifetime.

But this can be a problem. Still, if we want to change the contract, likely we can do this with a hardfork.

AnnaShaleva: > It's OK if we're managing fee distribution via a contract.

This is not yet accepted, that's why I've created this conversation.

> Still, if we want to change the contract, likely we can do this with a hardfork.

We can. Or, if needed for NeoX T2, I can add something like NativeUpdateHistory or ValidatorsHistory but for Coinbase:
```
coinbaseHistory:
    0: "0x1212100000000000000000000000000000000001"
    100500: "0x2323200000000000000000000000000000000002"
```

Do we need this right now?

roman-khimov: I don't think it's needed right now, not hard to extend in future.

consensus/dbft/dbft.go
```diff
@@ -214,11 +214,9 @@ type DBFT struct {
 	recents    *lru.Cache[common.Hash, *Snapshot] // Snapshots for recent block to speed up reorgs
 	signatures *sigLRU                            // Signatures of recent blocks to speed up mining
 
-	proposals map[common.Address]bool // Current list of proposals we are pushing
```
roman-khimov, line 221: Seems to be mixed into 96fb5e9a8d3e40771a3c38daa7f8edf329849698.

consensus/dbft/api.go
```diff
@@ -129,34 +108,29 @@ type status struct {
...
+		numBlocks = defaultStatisticsPeriod
 		header    = api.chain.CurrentHeader()
 		diff      = uint64(0)
 		optimals  = 0
 	)
-	snap, err := api.clique.snapshot(api.chain, header.Number.Uint64(), header.Hash(), nil)
-	if err != nil {
-		return nil, err
+	if n != nil {
+		numBlocks = *n
```
roman-khimov: So I can request statistics for 1M blocks? It'll take some time for the node to get it.

AnnaShaleva: Added max constraint.

consensus/dbft/dbft.go
AnnaShaleva: Fixed.

roman-khimov (approved):

roman-khimov:
Let's wait for @nicolegys to confirm.

consensus/dbft/dbft.go
chenquanyu: > should be the same within the whole network lifetime.

We design the governance contract as upgradable proxy contract, so it's not a critical problem.

nicolegys:
`dbft_status` consumes cpu, we may need a param that can be configured via node configuration to limit the number of blocks.

No-api call


Call `dbft_status` with no param (default: 64), 1 thread, thread delay in 1 ms


Call `dbft_status` with param=56000, 1 thread, thread delay in 1 ms

privnet/four/genesis_template.json
nicolegys: Yes, now the beneficiary can be configured by `coinbase`.

nicolegys (commented):
I think the `coinbase` field should be immutable.



 just like if I shut down the geth instance, change the value of `standbyValidators` and then run `geth init`, it will fail.

chenquanyu (approved):

AnnaShaleva:
I have a set of fixes for questions posted by @nicolegys, so not ready for the merge yet.

nicolegys:
> I think the `coinbase` field should be immutable.
> 
> 
> 
> just like if I shut down the geth instance, change the value of `standbyValidators` and then run `geth init`, it will fail.
> 
> 

@AnnaShaleva By the way, I've found that the CNs can still reach consensus if all of them change the `coinbase` to the same value and reinit. But If `coinbase` is immutable, it doesn't matter.

AnnaShaleva:
> @AnnaShaleva By the way, I've found that the CNs can still reach consensus if all of them change the `coinbase` to the same value and reinit.

This is expected result since all dBFT nodes have an agreement on new Coinbase (they don't care what's the old one). But conceptually we need to store the whole Coinbase history for proper old blocks verification, and I'll create a separate issue for that. And for now I've prohibited to change Coinbase at all at the dBFT engine level, take a look at the latest two commits. So now after node reinitialisation with changed Coinbase or StandByValidators we've got an error after node start:
```
anna@kiwi:~/Documents/GitProjects/bane-labs/go-ethereum$ ./build/bin/geth init --datadir privnet/four/node1/ privnet/four/node1/genesis_privnet.json >> privnet/four/node1/geth_init.log 
INFO [01-18|12:27:09.176] Maximum peer count                       ETH=50 LES=0 total=50
INFO [01-18|12:27:09.178] Smartcard socket not found, disabling    err="stat /run/pcscd/pcscd.comm: no such file or directory"
INFO [01-18|12:27:09.180] Using pebble as the backing database 
INFO [01-18|12:27:09.180] Allocated cache and file handles         database=/home/anna/Documents/GitProjects/bane-labs/go-ethereum/privnet/four/node1/geth/chaindata cache=512.00MiB handles=2048
INFO [01-18|12:27:09.180] Opened ancient database                  database=/home/anna/Documents/GitProjects/bane-labs/go-ethereum/privnet/four/node1/geth/chaindata/ancient/chain readonly=true
INFO [01-18|12:27:09.180] State scheme set to already existing     scheme=hash
INFO [01-18|12:27:09.181] Set global gas cap                       cap=50,000,000
INFO [01-18|12:27:09.181] Initializing the KZG library             backend=gokzg
INFO [01-18|12:27:09.212] Using pebble as the backing database 
INFO [01-18|12:27:09.212] Allocated cache and file handles         database=/home/anna/Documents/GitProjects/bane-labs/go-ethereum/privnet/four/node1/geth/chaindata cache=16.00MiB  handles=16
INFO [01-18|12:27:09.222] Opened ancient database                  database=/home/anna/Documents/GitProjects/bane-labs/go-ethereum/privnet/four/node1/geth/chaindata/ancient/chain readonly=false
INFO [01-18|12:27:09.222] State scheme set to already existing     scheme=hash
INFO [01-18|12:27:09.223] Successfully wrote genesis state         database=chaindata hash=98bb61..611263
INFO [01-18|12:27:09.223] Using pebble as the backing database 
INFO [01-18|12:27:09.223] Allocated cache and file handles         database=/home/anna/Documents/GitProjects/bane-labs/go-ethereum/privnet/four/node1/geth/lightchaindata cache=16.00MiB  handles=16
INFO [01-18|12:27:09.233] Opened ancient database                  database=/home/anna/Documents/GitProjects/bane-labs/go-ethereum/privnet/four/node1/geth/lightchaindata/ancient/chain readonly=false
INFO [01-18|12:27:09.233] State scheme set to already existing     scheme=hash
INFO [01-18|12:27:09.235] Successfully wrote genesis state         database=lightchaindata hash=98bb61..611263
anna@kiwi:~/Documents/GitProjects/bane-labs/go-ethereum$ make privnet_start_four 
Starting nodes...
OK! Check logs in ./privnet/four/<node_dir>/geth_node.log
anna@kiwi:~/Documents/GitProjects/bane-labs/go-ethereum$ cat privnet/four/node1/geth_node.log 
INFO [01-18|12:27:21.352] Enabling metrics collection 
INFO [01-18|12:27:21.352] Maximum peer count                       ETH=50 LES=0 total=50
INFO [01-18|12:27:21.354] Smartcard socket not found, disabling    err="stat /run/pcscd/pcscd.comm: no such file or directory"
INFO [01-18|12:27:21.356] Using pebble as the backing database 
INFO [01-18|12:27:21.356] Allocated cache and file handles         database=/home/anna/Documents/GitProjects/bane-labs/go-ethereum/privnet/four/node1/geth/chaindata cache=512.00MiB handles=2048
INFO [01-18|12:27:21.356] Opened ancient database                  database=/home/anna/Documents/GitProjects/bane-labs/go-ethereum/privnet/four/node1/geth/chaindata/ancient/chain readonly=true
INFO [01-18|12:27:21.356] State scheme set to already existing     scheme=hash
INFO [01-18|12:27:21.357] Set global gas cap                       cap=50,000,000
INFO [01-18|12:27:21.357] Initializing the KZG library             backend=gokzg
INFO [01-18|12:27:21.449] Allocated trie memory caches             clean=154.00MiB dirty=256.00MiB
INFO [01-18|12:27:21.449] Using pebble as the backing database 
INFO [01-18|12:27:21.449] Allocated cache and file handles         database=/home/anna/Documents/GitProjects/bane-labs/go-ethereum/privnet/four/node1/geth/chaindata cache=512.00MiB handles=2048
INFO [01-18|12:27:21.486] Opened ancient database                  database=/home/anna/Documents/GitProjects/bane-labs/go-ethereum/privnet/four/node1/geth/chaindata/ancient/chain readonly=false
Fatal: Failed to register the Ethereum service: failed to create dBFT engine: dBFT config has been changed: Coinbase mismatch: expected 0x1212100000000000000000000000000000000001, got 0x1212100000000000000000000000000000000002
```

AnnaShaleva:
@nicolegys, I've created a separate issue for RPC dBFT statistics call, see https://github.com/bane-labs/go-ethereum/issues/102, since it's not critical it will be done in a separate PR.

The issue with immutable coinbase is resolved in the current PR since it's a good practice to notify the node's operator about invalid config. But as an enhancement we'll have https://github.com/bane-labs/go-ethereum/issues/101.

So take a look one more time, please, we're going to include this PR into NeoX T2. @roman-khimov @chenquanyu

consensus/dbft/api.go
```diff
@@ -129,34 +113,29 @@ type status struct {
...
-		numBlocks = uint64(64)
+		numBlocks = defaultStatisticsPeriod
 		header    = api.chain.CurrentHeader()
 		diff      = uint64(0)
 		optimals  = 0
 	)
-	snap, err := api.clique.snapshot(api.chain, header.Number.Uint64(), header.Hash(), nil)
-	if err != nil {
-		return nil, err
+	if n != nil && *n <= maxStatisticsPeriod {
```
nicolegys: I think if `n>maxStatisticsPeriod`, n should be set as `maxStatisticsPeriod` instead of `defaultStatisticsPeriod`.

AnnaShaleva: Sure, fixed.

consensus/dbft/config.go
```diff
@@ -0,0 +1,106 @@
...
+	"github.com/ethereum/go-ethereum/params"
+)
+
+// dbVersion is a version of storage currentlu used by DBFT module.
+const dbVersion = "0.0.1"
+
+// A set of Eth database prefixes related to DBFT module.
+var (
+	prefixDBVersion = byte(0x01)
+	prefixDBConfig  = byte(0x02)
```
chenquanyu: We stored the dbft config when genesis init, so can we just check the config in `(c *ChainConfig) checkCompatible()`?

nicolegys (approved):
OK from test side, wait the unresolved coversation to be confirmed.

consensus/dbft/config.go
AnnaShaleva: Oh, my bad, I haven't noticed the existing verification code. Will fix.

AnnaShaleva: Nice catch!

AnnaShaleva: OK, finally fixed. Geth has quite complicated genesis initialisation process, and the case is that we can't just extend `(c *ChainConfig) checkCompatible()`, because this method should return only special hardfork-related errors that require the node to be reset to some particular height after this error detected. And for some reason we always have genesis block returned from `rawdb.ReadHeadHeader(db)`, there might be some problem with block persist, it shouldn't be like that, will create an issue for this.

And to solve the problem you've mentioned I created a special `(c *ChainConfig) CheckConsensusEngineCompatible`, so now it's forbidden to overwrite an existing genesis on node reinitialisation if consensus engine config is incompatible. Check the latest commit.

AnnaShaleva: Ref. https://github.com/bane-labs/go-ethereum/issues/105 for the head header problem.

AnnaShaleva:
@nicolegys, take a look one more time please, I've refactored dBFT config compliance check and moved it to `init` level.

chenquanyu (approved):

nicolegys (approved):

PR #109, opened 2024-01-22
dbft: fix dbft api for Geth JavaScript console. 

### LINKED ISSUE #36: DBFT module error
cannot call dbft functions
image

chenquanyu:
Close #36 DBFT module error. This issue is caused by the command in `"JavaScript console` needs to be added in `web3ext`.

### DIFF
--- a/internal/web3ext/web3ext.go
+++ b/internal/web3ext/web3ext.go
@@ -20,6 +20,7 @@ package web3ext
 var Modules = map[string]string{
 	"admin":    AdminJs,
 	"clique":   CliqueJs,
+	"dbft":     DbftJs,
 	"ethash":   EthashJs,
 	"debug":    DebugJs,
 	"eth":      EthJs,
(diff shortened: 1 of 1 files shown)


nicolegys (approved):

internal/web3ext/web3ext.go
```diff
@@ -90,6 +91,51 @@ web3._extend({
...
+		}),
+		new web3._extend.Method({
+			name: 'status',
+			call: 'dbft_status',
+			params: 1
+		}),
+		new web3._extend.Method({
+			name: 'getPrimary',
+			call: 'dbft_getPrimary',
+			params: 1,
```
AnnaShaleva, line 129: We have `inputFormatter: [null]` for the analogue Clique method and it seems to me like a good practice. Don't we need to set the same one for `getPrimary` and `getCoinbase`?

AnnaShaleva (commented):
Could you also please, add `Close #36.` message to the PR description? It's needed for proper GitHub "Development" section links displaying.

internal/web3ext/web3ext.go
chenquanyu: This `inputFormatter: [null]` seems to disable the check for params count, so I think we only need this for `dbft_status`.

nicolegys:
now `dbft.status` can be called without param

internal/web3ext/web3ext.go
nicolegys: I think `getPrimary` and `getCoinbase` also need this because they can be called without param in the code implementation.

chenquanyu: > I think `getPrimary` and `getCoinbase` also need this because they can be called without param in the code implementation.

I have tried to add it for `getPrimary` and `getCoinbase`, but call them without param will cause exception. 

image

nicolegys: Oh sorry, my bad. I mean to say `getSigners` and `getValidators` can be called without param, and they actually work well. It's OK OK.

AnnaShaleva (approved):

PR #112, opened 2024-01-23
privnet: move genesis config to upper level

### LINKED ISSUE #111: Move privnet nodes genesis configuration to upper level
Originally posted by @chenquanyu. Starting from #29 we have a set of identical `genesis_privnet,json` files for every privnet setup. We need to move `genesis_privnet.json` outside of the node-specific folders to the upper level. As a result we'll have only three `genesis_privnet.json` files and three corresponding templates.

Corresponding Makefile adjustments should be implemented.

AnnaShaleva:
Close #111. Adjust Makefile correspondingly.


AnnaShaleva:
@chenquanyu, @nicolegys, you can also review. Due to bane-labs org restrictions I can't request review from multiple people.

roman-khimov (approved):

PR #114, opened 2024-01-24
privnet: add Bridge contract to genesis config

AnnaShaleva:
The Bridge config is exactly the same that we use for NeoX T2. Also, make 7-nodes privnet config closer to T2 config.


AnnaShaleva:
@roman-khimov, I'm not sure it's a good idea to keep "unknown" version of contract in the privnet genesis, but at the same time it's still a way to check that Testnet environment works, so I think we need to merge this PR. What do you think?

privnet/seven/genesis_privnet.json
```diff
@@ -25,18 +25,36 @@
...
-      "coinbase": "0x1212100000000000000000000000000000000001"
+      "coinbase": "0x1212100000000000000000000000000000000003"
     }
   },
   "difficulty": "1",
-  "gasLimit": "7000000",
+  "gasLimit": "30000000",
   "alloc": {
-    "74f4effb0b538baec703346b03b6d9292f53a4cd": { "balance": "100000000000000000000" },
+    "74f4effb0b538baec703346b03b6d9292f53a4cd": { "balance": "100000000000000000000000" },
```
roman-khimov, line 34: Why this one?

roman-khimov (commented):
We can have it as is for now (synchronized with T2) and then update when needed.

privnet/seven/genesis_privnet.json
AnnaShaleva: Just a single validator from the batch to follow the notion of T2 config.

AnnaShaleva:
> We can have it as is for now (synchronized with T2)

Good, then current version is in sync.

roman-khimov (approved):

PR #118, opened 2024-01-29
dbft: introduce Extra versioning

### LINKED ISSUE #50: Change header's Extra format
Currently (after #49) header's Extra field have the following format (assuming N validators and M out of N is dBFT honest nodes count):
* 32 bytes reserved for extra vanity, they are a part of sealing data and currently are always empty;
* N validators addresses (each of 20 bytes) that are effectively the full validators set for this block sorted by byte order;
* M validators signatures (each of 65 bytes) sorted by corresponding validators bytes.

We don't need first 32 bytes of Extra to be reserved. We need to change Extra format and replace 32 reserved bytes with a single reserved byte denoting version of Extra. In future we'll be able to change Extra format relying to the first Version byte. The scheme itself is very similar to the Block versioning we currently have in N3. Within the scope of this issue we also need to evaluate compatibility with existing Eth hardforks supported by Geth node (no conflicts with hardforks should happen).

AnnaShaleva:
Close #50. Improvement mentioned in https://github.com/bane-labs/go-ethereum/issues/50#issuecomment-1862816942 may be implemented in `ExtraV1`.

### DIFF
--- a/consensus/dbft/dbft.go
+++ b/consensus/dbft/dbft.go
@@ -64,8 +64,6 @@ import (
 
 // DBFT proof-of-authority protocol constants.
 var (
-	extraVanity = 32 // Fixed number of extra-data prefix bytes reserved for signer vanity
-
 	uncleHash = types.CalcUncleHash(nil) // Always Keccak256(RLP([])) as uncles are meaningless outside of PoW.
 
 	diffStub   = big.NewInt(3) // Block difficulty stub that is used for miner's task preparation.
@@ -134,7 +132,7 @@ func getSignersAndSigs(cfg *params.DBFTConfig, extra []byte) ([]common.Address, 
 		addrsBytesLen = common.AddressLength * n
 		sigsBytesLen  = extraSeal * m
 	)
-	if len(extra) < extraVanity+addrsBytesLen+sigsBytesLen {
+	if len(extra) < dbftutil.ExtraVersionLen+addrsBytesLen+sigsBytesLen {
 		return nil, nil, errMissingSignature
 	}
 
@@ -146,7 +144,7 @@ func getSignersAndSigs(cfg *params.DBFTConfig, extra []byte) ([]common.Address, 
 		sigs  = make([][]byte, m)
 	)
 	for i := range addrs {
-		addrOffset := extraVanity + i*common.AddressLength
+		addrOffset := dbftutil.ExtraVersionLen + i*common.AddressLength
 		copy(addrs[i][:], extra[addrOffset:addrOffset+common.AddressLength])
 	}
 	for i := range sigs {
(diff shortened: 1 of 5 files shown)


consensus/dbft/dbft.go
```diff
@@ -865,10 +867,12 @@ func (c *DBFT) Prepare(chain consensus.ChainHeaderReader, header *types.Header)
...
 	header.Difficulty = diffStub
 
-	// Ensure the extra data has all its components
+	// Ensure the extra data has all its components. Set default Extra version if not
+	// provided by miner.
 	if len(header.Extra) < dbftutil.ExtraVersionLen {
-		header.Extra = append(header.Extra, bytes.Repeat([]byte{0x00}, dbftutil.ExtraVersionLen-len(header.Extra))...)
+		header.Extra = []byte{dbftutil.ExtraV0}
 	}
+
```
roman-khimov: :eyes:

core/genesis.go
```diff
@@ -495,6 +495,7 @@ func (g *Genesis) ToBlock() *types.Block {
 		)
 		if len(g.ExtraData) == 0 {
 			extra := make([]byte, dbftutil.ExtraVersionLen+n*common.AddressLength+m*crypto.SignatureLength)
+			extra[0] = dbftutil.ExtraV0
```
roman-khimov, line 498: It's zero already and genesis isn't likely to be changed wrt this.

miner/miner.go
```diff
@@ -193,6 +194,9 @@ func (miner *Miner) SetExtra(extra []byte) error {
 	if uint64(len(extra)) > params.MaximumExtraDataSize {
 		return fmt.Errorf("extra exceeds max length. %d > %v", len(extra), params.MaximumExtraDataSize)
 	}
+	if miner.worker.chainConfig.DBFT != nil && len(extra) > dbftutil.ExtraV0 {
```
roman-khimov: I'm not sure how it could happen.

core/genesis.go
AnnaShaleva: Could we keep an explicit setter, just in case if we're to change this in future? It's easy to forget about it while refactoring.

consensus/dbft/dbft.go

miner/miner.go
AnnaShaleva: Extra can be customizable via two ways: miner's setting on node start and exposed miner API during runtime. So DBFT engine sets Extra only if it's not provided by miner (and currently the default Extra value for miner is ExtraV0). I also thought about prohibiting Extra customization for dBFT, but we may need this feature later. Do you prefer to prohibit?

roman-khimov: I'd say so. `Extra` is not really "extra" for us, we need some very specific data there, so any "miner value" is wrong immediately.

AnnaShaleva: Agree, fixed.

roman-khimov (approved):

PR #87, opened 2024-01-09
dbft: Implement block verification callback

chenquanyu:
Close #65 Implement block verification callback.

Issue #65, opened 2023-12-21
Implement block verification callback


AnnaShaleva:
After #49 we're missing block verification callback, so almost all proposed block fields remain unverified. Within the scope of this issue we need to implement `WithVerifyBlock` dBFT callback to check all header's field for correctness and all in-block transactions. Most of verification rules mentioned in https://github.com/bane-labs/go-ethereum/issues/53 should be done not on the PrepareRequest level verification, but on the block verification level.

Issue #82, opened 2023-12-29
Optimize PrepareRequest construction


AnnaShaleva:
NeoGo has mechanism that in case of CV allows to fetch some of the known transactions from the previous dBFT view and construct new view PrepareRequest based on these known hashes. It's needed to make it easier for CN to collect necessary set of transactions to approve block: https://github.com/nspcc-dev/neo-go/blob/3176f72878bab11fc824c53e8b24b584971163f7/pkg/consensus/consensus.go#L642-L672.

However, implementing it on Geth is a tricky task because of the current way of dBFT integration with miner. dBFT doesn't create proposal by itself, it waits for the miner's proposal to be accepted, and thus, it can't precisely manage the set of transactions in proposal. So that's the caveat.

Ref. #59.

AnnaShaleva:
After anti-MEV dBFT extension integration (#287) a set of other fields become useless at PrepareRequest level, in particular, those fields that may be changed in the final `Block` due to decrypted transactions addition:
- ReceiptHash
- MixDigest
- Merkle root
- Transactions root
- Receipts hash
- Bloom
- may be something else, need to check carefully.

These fields can't be verified by CNs at PrepareRequest level with AMEV logic and hence, they should be removed from PrepareRequest.

Issue #14, opened 2023-09-14
dBFT stage 4: integrate voting


roman-khimov:
Header chain should be verifiable on its own, but we don't want to include the set of not-often-changed keys into every block, so:
 * allow for a key list to be included into `extra` as well, but only use it when `nextConsensus` differs from the current set
 * it should be included into the genesis block
 * once it's in the header, take them, verify they equal the `nextConsensus` hash and use to verify the next block
 * CNs should get the list from the voting contract once it's ready

AnnaShaleva:
After #28 I have two reminders that should be carefully handled:

* `WithGetValidators` dBFT callback must sort validators, because primary calculation relies on this. Currently it's just the list of accounts taken from snapshot, so we need to ensure that they are properly sorted.
* `WithGetValidators` callback must return two separate lists of validators depending on the `txs` argument presence on the dBFT epoch bounds. We should be careful while implementing validators calculations for this callback, the same logic with `NextBlokValidators` and `NewEpochNextBlockValidators` from https://github.com/nspcc-dev/neo-go/pull/3110 may be implemented.

roman-khimov:
Also adjust [eth/protocols/dbft/ledger.go](https://github.com/bane-labs/go-ethereum/compare/8f32d59cee70e37e95b9b40e7089fb44e6126f36..6d6fd68af4a6b7d112c07eb4c29ea7a7b3a7b259#diff-db0285e0e60d5a696aee679c608b6bccc6e6e95da1912163bf3e303388715be7) to check for allowed dbft message senders.

AnnaShaleva:
Ref. https://github.com/bane-labs/go-ethereum/pull/7/files#r1462195034, `getPhaseByHeight` and `getCurrentPhase` APIs of Governance contract need to be used for proper Validators/NextConsensus retrieval on the dBFT side.

AnnaShaleva:
Need to pay special attention to the `lastBlockNextNextConsensus` field of DBFT engine initialisation. There might be some problems with state-dependend NextConsensus initialisation after forced node shutdown when node doesn't have complete state in its storage to compute NextConsensus field for the next block. Ref. https://github.com/bane-labs/go-ethereum/pull/121#discussion_r1475895142 and ref. #100.

PR #121, opened 2024-01-31
Properly verify PrepareRequest fields

### LINKED ISSUE #53: Proper PrepareRequest verification
Currently (after #49) we have some basic PrepareRequest verification rules mostly focusing on block hash finalisation and checking that parent-related data (final parent hash and extra) filled properly (ref. `WithVerifyPrepareRequest` dBFT setting). What we're completely missing is proposed header fields verification. Within the scope of this issue we need to:
* check that transactions root retrieved from proposed transaction hashes matches the one specified in proposed header;
* check that receipts root declared in the proposed header matches the real receipts root got from transactions execution;
* check that proper difficulty is set
* perform any other necessary verification actions (they are very similar to the `VerifyHeader` consensus engine API).

As a result of this issue I'd expect that potentially invalid blocks won't be accepted by dBFT consensus and proposals (PrepareRequests) for such blocks will be immediately rejected by honest consensus nodes. Note that this issue is only about `WithVerifyPrepareRequest` callback extension.

AnnaShaleva:
Close https://github.com/bane-labs/go-ethereum/issues/53. Related to #65 and #87.

This PR introduces PrepareRequest verification on `WithVerifyPrepareRequest` callback level. It reuses as much code of `verifyHeader` as possible.

### DIFF
--- a/consensus/dbft/dbft.go
+++ b/consensus/dbft/dbft.go
@@ -462,8 +462,14 @@ func New(config *params.DBFTConfig, _ ethdb.Database) (*DBFT, error) {
 			if req.SealingProposal == nil {
 				return errors.New("failed to verify PrepareRequest: sealing proposal is nil")
 			}
-			if req.SealingProposal.Coinbase != c.config.Coinbase {
-				return fmt.Errorf("invalid Coinbase: expected %s, got %s", c.config.Coinbase, req.SealingProposal.Coinbase)
+			parent := c.chain.GetBlockByNumber(req.SealingProposal.Number.Uint64() - 1)
+			if parent == nil {
+				return fmt.Errorf("no parent found for height %d", req.SealingProposal.Number.Uint64()-1)
+			}
+			parentHeader := parent.Header()
+			err := c.verifyHeader(c.chain, req.SealingProposal, []*types.Header{parentHeader}, false)
+			if err != nil {
+				return fmt.Errorf("invalid header: %w", err)
 			}
 			if req.SealingProposal.ParentHash != c.lastBlockHash {
 				// Genesis block  is hard-coded, thus its hash (as a parent hash) must always match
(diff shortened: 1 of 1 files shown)


AnnaShaleva:
@nicolegys, @chenquanyu, you are welcomed to review&test.

consensus/dbft/dbft.go
```diff
@@ -688,14 +689,14 @@ func (c *DBFT) VerifyHeaders(chain consensus.ChainHeaderReader, headers []*types
...
+func (c *DBFT) verifyHeader(chain consensus.ChainHeaderReader, header *types.Header, parents []*types.Header, isSealed bool) error {
 	if header.Number == nil {
 		return errUnknownBlock
 	}
 	number := header.Number.Uint64()
 
-	// Don't waste time checking blocks from the future
-	if header.Time > uint64(time.Now().Unix()) {
+	// Don't waste time checking blocks from the future.
+	if isSealed && header.Time > uint64(time.Now().Unix()) {
```
roman-khimov, line 722: We can't accept _any_ time in proposal.

consensus/dbft/dbft.go
```diff
@@ -710,29 +711,32 @@ func (c *DBFT) verifyHeader(chain consensus.ChainHeaderReader, header *types.Hea
...
+		signersBytes := len(header.Extra) - dbftutil.ExtraVersionLen - sigBytesLen
+		if signersBytes == 0 {
+			return fmt.Errorf("missing validators addresses")
+		}
+		if signersBytes%common.AddressLength != 0 {
+			return errInvalidExtraSigners
+		}
+
+		// Ensure that the mix digest is not zero.
+		if header.MixDigest == (common.Hash{}) {
```
roman-khimov, line 753: Can we have it in proposal and check local one against proposed?

consensus/dbft/dbft.go
AnnaShaleva: We do check the proposal time, the condition is not to be lower than the previous block. This check is located at the `WithVerifyBlock` level, see https://github.com/bane-labs/go-ethereum/pull/87#discussion_r1472866452. I agree that this check can be done at the PrepareRequest verification level, but currently it's done exactly like in NeoGo, and it works properly (so *any* timestamp can't be accepted by CNs, only larger than the previous block's timestamp).

consensus/dbft/dbft.go
AnnaShaleva: We can, but why do we need this? MixDigest depends on validators list, nodes take validators list from the dBFT context, and if something is wrong with the validators list, the consensus just won't work. And nodes fill this field by their own before commit is send to include this field in the hashable part, exactly like in NeoGo. If malicious node wants to change this field, then Commit verification will fail on a good node. If malicios node will do everything right and intentionally change MixDigest after Commit release, then the resulting block relayed to the network by this node won't be accepted by other nodes because it has wrong MixDigest. But blocks relayed by other honest consensus nodes will be accepted.

So I don't think we need to include this field in PrepareRequest. And its stub will be removed from PrepareRequest in #82. Am I missing something here?

roman-khimov: Fail fast. And better with some meaningful reason. Right that would be "wrong commit signature", go figure.

consensus/dbft/dbft.go
roman-khimov: OK, that will do the trick. The most important thing is for it to be checked _before_ sending PrepareResponse.

consensus/dbft/dbft.go
AnnaShaleva: Should we perform the same update in NeoGo?

AnnaShaleva: > Should we perform the same update in NeoGo?

OK, silly question, can't do this in NeoGo since NextConsensus is not a part of PrepareRequest and protocol extension is needed there. OK, will do it in Geth at least.

AnnaShaleva: Fixed in a separate commit. But we need to recheck it once Governance contract is integrated and state-dependent validators calculations are available.

AnnaShaleva:
@roman-khimov, ready for review.

consensus/dbft/dbft.go
AnnaShaleva: See https://github.com/bane-labs/go-ethereum/issues/14#issuecomment-1923566099.

roman-khimov: It's going to be tested wrt validator list change anyway, I think.

roman-khimov (approved):

PR #122, opened 2024-01-31
Enable Shanghai hardfork for dBFT-based chains

### LINKED ISSUE #113: Fix/enable Shanghai fork
We don't have https://github.com/ethereum/execution-specs/blob/master/network-upgrades/mainnet-upgrades/shanghai.md enabled in our configs which makes us incompatible with the most recent Solidity tools producing PUSH0 (likely other problems can happen as well). Enabling it breaks dbft though (as well as clique), so this requires some fixes from the consensus part.

Refs. #110.

AnnaShaleva:
Close #113.

~WIP, need to carefully check the rest of Shanghai changes.~ All Shanghai upgrades described in https://github.com/ethereum/execution-specs/blob/master/network-upgrades/mainnet-upgrades/shanghai.md has been checked, safe to enable now.

NOTE: for now, only empty withdrawals are allowed and thus, no withdrawals are included into PrepareRequest (only withdrawals root that is being validated to match empty withdrawals root). Once we need to support non-empty withdrawals, it may be done via additional PrepareRequest field and with minimal modifications in the `verifyHeader`/`WithVerifyPrepareRequest` method.

### DIFF
--- a/consensus/dbft/dbft.go
+++ b/consensus/dbft/dbft.go
@@ -69,6 +69,8 @@ var (
 	diffStub   = big.NewInt(3) // Block difficulty stub that is used for miner's task preparation.
 	diffInTurn = big.NewInt(2) // Block difficulty for in-turn signatures
 	diffNoTurn = big.NewInt(1) // Block difficulty for out-of-turn signatures
+
+	emptyWithdrawals = make([]*types.Withdrawal, 0)
 )
 
 const (
@@ -327,6 +329,9 @@ func New(config *params.DBFTConfig, _ ethdb.Database) (*DBFT, error) {
 			res := types.NewBlockWithHeader(ethBlock.header)
 			// Uncles are always nil in dBFT-like consensus.
 			res = res.WithBody(ethBlock.transactions, nil)
+			if ethBlock.withdrawals != nil {
+				res = res.WithWithdrawals(ethBlock.withdrawals)
+			}
 
 			// Firstly, notify chain about new block.
 			if err := c.blockQueue.PutBlock(res); err != nil {
@@ -371,9 +376,14 @@ func New(config *params.DBFTConfig, _ ethdb.Database) (*DBFT, error) {
 			// (dBFT has only the full set of their hashes). Once all transactions are
 			// fetched and the commits are collected, SetTransactions callback will be
 			// called by dBFT library to properly initialize block's transactions.
-			return &Block{
+			res := &Block{
 				header: h,
 			}
+			// Withdrawals are temporary empty if Shanghai is passed.
+			if c.chain.Config().IsShanghai(h.Number, h.Time) {
+				res.withdrawals = emptyWithdrawals
+			}
+			return res
 		}),
 		dbft.WithWatchOnly(func() bool {
 			return false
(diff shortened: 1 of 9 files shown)


AnnaShaleva:
@roman-khimov, the only thing I worry about is that we are starting to mix the post-merge behaviour (Shanghai is a post-merge hardfork) with pre-merge miner behaviour (difficulty/forks resolving related things belong to pre-merge). It's OK with Shanghai, but with some other fork it may break something inside the node.

AnnaShaleva:
@roman-khimov, @chenquanyu, @nicolegys ready for review and testing.

consensus/dbft/dbft.go
```diff
@@ -890,27 +913,41 @@ func (c *DBFT) Prepare(chain consensus.ChainHeaderReader, header *types.Header)
 	return nil
 }
 
-// Finalize implements consensus.Engine. There is no post-transaction
-// consensus rules in dbft, do nothing here.
+// Finalize implements consensus.Engine. For now, it only manages block withdrawals.
 func (c *DBFT) Finalize(chain consensus.ChainHeaderReader, header *types.Header, state *state.StateDB, txs []*types.Transaction, uncles []*types.Header, withdrawals []*types.Withdrawal) {
+	// Withdrawals processing.
```
roman-khimov, line 918: Do we need to process them? Or you're adding just in case we'd have non-zero withdrawals?

consensus/dbft/dbft.go
```diff
@@ -890,27 +913,41 @@ func (c *DBFT) Prepare(chain consensus.ChainHeaderReader, header *types.Header)
...
-	c.Finalize(chain, header, state, txs, uncles, nil)
+	c.Finalize(chain, header, state, txs, uncles, withdrawals)
 
 	// Assign the final state root to header.
 	header.Root = state.IntermediateRoot(chain.Config().IsEIP158(header.Number))
 
 	// Assemble and return the final block for sealing.
-	b := types.NewBlock(header, txs, nil, receipts, trie.NewStackTrie(nil))
-
+	b := types.NewBlockWithWithdrawals(header, txs, nil, receipts, withdrawals, trie.NewStackTrie(nil))
```
roman-khimov, line 950: No Shanghai condition for this?

consensus/dbft/dbft.go
AnnaShaleva: > Do we need to process them?

Right now - we don't. And to say more, post-merge beacon chain handles them by itself:
https://github.com/bane-labs/go-ethereum/blob/ecc874308ce9274e9fdd9c8f041d3754178e025b/consensus/beacon/consensus.go#L350-L363

But our code is pre-merge oriented, and thus, we should handle them manually.

>Or you're adding just in case we'd have non-zero withdrawals?

You're right, this code is for future if one day we decide to support non-empty withdrawals.

consensus/dbft/dbft.go
AnnaShaleva: We have this condition above, and `types.NewBlockWithWithdrawals` is able to properly handle nil/empty/non-empty withdrawals.

roman-khimov (approved):

PR #124, opened 2024-02-01
dbft: use precalculated receipts (if available) to store block

### LINKED ISSUE #60: Optimize the way how dBFT-accepted blocks are being inserted into chain
Currently after #49 every block built by dBFT is being routed to `blockQueue` instance that is trying to send the block back to miner (if a proper sealing task is found) or to insert it to chain directly. In case of direct chain insertion `blockQueue` instance doesn't have transaction receipts being calculated for this block. Latter occurs mostly when the node is not a primary of the current block. Receipts are being recalculated by `Blockchain` during the block insertion. However, after #53 we'll have a proper set of receipts calculated for every PrepareRequest on the backup node side (for verification purposes).

Within the scope of this issue we can store receipts calculated during PrepareRequest verification and pass the new dBFT block to the `blockQueue` instance along with the corresponding receipts. This allows to omit double receipts recalculation and save time on computations. For example of how block with receipts, state and log can be stores, refer to miner's code: https://github.com/bane-labs/go-ethereum/blob/94233fe05ce42ac2e4c4556ea1b3b871f2741964/miner/worker.go#L686

This issue is roughly depends on #53 and is not a high priority.

AnnaShaleva:
WIP, strongly depends on #87.

**Warning**
I'm not quite sure about code correctness, there may be some caveats during in-flight transactions processing, so this PR seems quite dangerous to me. It was an attempt to solve #100, but the problem described in #100 isn't fixed with this code, so I've checked that receipts-less block insertion (via `InsertChain` API instead of `WriteBlockAndSetHead`) works OK for us and this PR may be safely postponed (it's kind of optimization) or even closed until we have enough time to properly review the correctness of receipts calculations presented in this PR.

**PR content**
The idea of this PR is to keep block's receipts got after PrepareRequest level block verification and use them while submitting the resulting block to the chain in order to save some time on receipts recalculation. Close https://github.com/bane-labs/go-ethereum/issues/60.

The previous behaviour is the following:
* if no CV happens after sealing proposal initialisation by miner (i.e. on primary node), then receipts are not being recalculated (they are calculated and submitted to chain by miner)
* on backup nodes receipts recalculation happens firstly inside the `WithVerifyBlock` callback to verify proposal; and secondly during the resulting block insertion into chain via BlockQueue using `InsertChain`to store the resulting receipts.
* on watch-only nodes recalculation happens once while inserting final block via BlockQueue using `InsertChain` since no block verification is being performed by watch-only nodes.

Current behaviour is the following:
* if no CV happens after sealing proposal initialisation by miner (i.e. on primary node), then receipts are not being recalculated (they are calculated and submitted to chain by miner)
* on backup nodes receipts recalculation happens once inside the `WithVerifyBlock` callback to verify proposal; the resulting receipts are being inserted into chain in the end via `WriteBlockAndSetHead` API
* on watch-only nodes receipts recalculation still happens once, no behaviour changes.

**Motivation**
Originally this commit aimed to solve the problem described in https://github.com/bane-labs/go-ethereum/issues/100. However, seems that direct receipts-less block insertion also works fine and it's not the root of the problem described in https://github.com/bane-labs/go-ethereum/issues/100.

### DIFF
--- a/consensus/dbft/block.go
+++ b/consensus/dbft/block.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/state"
 	"github.com/ethereum/go-ethereum/core/types"
 	ecrypto "github.com/ethereum/go-ethereum/crypto"
 	"github.com/nspcc-dev/dbft/block"
@@ -23,6 +24,10 @@ type Block struct {
 	header              *types.Header
 	transactions        []*types.Transaction
 	localSignatureBytes []byte
+
+	// Local data filled in during constructed block verification. Allowed to be empty.
+	receipts types.Receipts
+	state    *state.StateDB
 }
 
 // Version implements block.Block interface.
--- a/consensus/dbft/blockqueue.go
+++ b/consensus/dbft/blockqueue.go
@@ -6,6 +6,7 @@ import (
 	"sync"
 
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/state"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/log"
 )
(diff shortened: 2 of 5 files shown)


AnnaShaleva:
Ported to #141, thus, closing this PR.

PR #128, opened 2024-02-06
dbft: use MimetypeTextPlain for dbft Sign()

### LINKED ISSUE #61: Custom Mimetype for dBFT signatures
After #49 dBFT signer implementation signs all payloads/blocks with `accounts.MimetypeClique` type. We need to replace this memetype either with one of other available memetypes or to create our own memetype. Ref. https://github.com/bane-labs/go-ethereum/pull/49#discussion_r1423118845, ref. fd28f6a38678a436642734fdd1e8fbf4e2c6cb91.

This issue is not critical for network liveness, but we better have appropriate signatures.

chenquanyu:
Close #61 
For now we don't need a special type for dBFT block signatures since no special operations should be performed with the signature. Let's reuse the existing  `MimetypeTextPlain` type.

### DIFF
--- a/consensus/dbft/signer.go
+++ b/consensus/dbft/signer.go
@@ -22,5 +22,5 @@ type Signer struct {
 // In case of block, msg is expected to be dbftRLP(header) that must be
 // guaranteed by Block.Sign method.
 func (s *Signer) Sign(msg []byte) ([]byte, error) {
-	return s.SignFn(accounts.Account{Address: s.Signer}, accounts.MimetypeClique, msg)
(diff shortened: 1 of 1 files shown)


AnnaShaleva (approved):
@roman-khimov, octet stream type doesn't exist in the Geth and text plain is the most suitable from the existing ones.

roman-khimov (approved):
Yeah, I've seen that there are not a lot of options there. On one hand, "text/plain" is incorrect since it's not a text that we're signing, on the other it's certainly not worse than "application/x-clique-header". So OK for now.

Issue #116, opened 2024-01-29
Sync with Geth source code


AnnaShaleva:
Our bane-main is out-of-date with https://github.com/ethereum/go-ethereum, we need to review changes happened since the last merge and sync our codebase.

We also need to think about sync strategy: how often, how to sync with Bane releases, etc.

A side note: #113 may already be fixed in the original Geth Clique code.

PR #131, opened 2024-02-08
Sync master with original Geth master (v1.13.11 stable)

AnnaShaleva:
Fetch Geth up to v1.13.11 stable. Ref. #116.


roman-khimov:
Let's leave `master` as is, the original one doesn't need to be copied here. Frankly, I'd rather use `master` for our own work instead of `bane-main`.

PR #134, opened 2024-02-13
Add http and ws without authentication to privnet

gsmachado:
Before, just `--authrpc.port` was set. 

I'm adding `--http` and `--ws` for test convenience purposes -- otherwise it'd become complex if we had to generate a JWT token to use the JSON-RPC interface.

😅


AnnaShaleva (approved):
Good, hardly ever we'll need auth in privnet, but let's keep `NODEX_AUTH_PORT` for now.

PR #132, opened 2024-02-08
dbft: rework transactions routing

AnnaShaleva:
This commit contains two changes in the dBFT-related transactions handling:
    1. Store the sorted set of requested transactions and filter incoming ones wrt this set.
    2. Do not subscribe for mempooled transactions. Instead, receive transactions directly via P2P bypassing mempool since mempool assumes additional tx filtering.

Close #98.

Tested on multinodes privnet under high load, works as expected. Debugging logs are removed:
```
2024-02-08T17:41:56.760+0300	DEBUG	dbft@v0.0.0-20240130131324-6d59675ad370/dbft.go:214	received message	{"type": "PrepareResponse", "from": 6, "height": 93, "view": 0, "my_height": 93, "my_view": 0}
2024-02-08T17:41:56.760+0300	INFO	dbft@v0.0.0-20240130131324-6d59675ad370/dbft.go:437	received PrepareResponse	{"validator": 6}
[...]

PR #135, opened 2024-02-19
Add `--http.vhosts "*"` to privnet launch command

gsmachado:
This enables we execute the following request:

`curl -v -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"eth_blockNumber","params":[],"id":1}' http://neox-node:8563`

using the `neox-node` hostname, instead of the IP address of the HTTP endpoint.

This is useful when running geth in docker-compose, where an internal network is established with hostnames. 😅

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -118,6 +118,7 @@ define run_node
 		--http \
 		--http.port $(5) \
 		--http.addr 0.0.0.0 \
+		--http.vhosts "*" \
 		--ws \
 		--ws.addr 0.0.0.0 \
 		--ws.port $(6) \


Makefile
```diff
@@ -118,6 +118,7 @@ define run_node
 		--http \
 		--http.port $(5) \
 		--http.addr 0.0.0.0 \
+		--http.vhosts "*" \
```
AnnaShaleva, line 121: I'm kindly asking to follow the https://github.com/nspcc-dev/.github/blob/master/git.md#commits during commit creation (add `privnet:` prefix to the commit title since this PR is related to the privnet only,  not to the whole node; and make commit title shorter, any additional info can be placed into the commit body). For the sake of clear and coincide Git history. You may refer to https://github.com/bane-labs/go-ethereum/pull/132/commits as an example of commit messages title/content.

AnnaShaleva (commented):
The change itself LGTM.

Makefile
gsmachado: oh, yes. Sure! Didn't know you guys follow strict rules. But I agree!
Fixed.

AnnaShaleva (approved):

PR #130, opened 2024-02-07
Sync bane-main with original Geth master (v1.13.11 stable)

AnnaShaleva:
Merge Geth master updates starting from 766272ff8c1c3b1574d40db2eefff53371b84ce7 ending by 99dc3fe118a4d881d9b5347b5345669f52de8143. Close #116.

@roman-khimov, the last several commits contain corresponding dBFT changes that need to be done after this merge. I've carefully reviewd the list of changes that happened since the last sync, so here's the list of PRs that might be important for us:

- https://github.com/ethereum/go-ethereum/pull/28147 removal of rollback mechanism in downloader. We may still need this mechanism since we're pre-merge and allow forks;
- https://github.com/ethereum/go-ethereum/commit/3dc45a3e1dd50ce1812438a79226b470a839f89c related to release pipeline, example of pre-release version update;
- https://github.com/ethereum/go-ethereum/commit/dc34fe8291bfcaefbce97f559e9610beffb2e470 related to release pipeline, example of post-release version update;
- https://github.com/ethereum/go-ethereum/pull/28098 BLOBBASEFEE opcode implemented as a part of Cancun (we have to support it eventually);
- https://github.com/ethereum/go-ethereum/pull/28195 introduce new Cancun-related block fields to eth tools;
- https://github.com/ethereum/go-ethereum/pull/28243 introduce blob transactions support for internal eth services;
- https://github.com/ethereum/go-ethereum/pull/28084 allows to invoke contract at specific block hash, may be useful for Governance contract integration;
- https://github.com/ethereum/go-ethereum/pull/28538 Dockerfile.alltools update example, just to remember that it should be in sync with Dockerfile;
- https://github.com/ethereum/go-ethereum/pull/28605 Improved Cancun- and Shanghai- related consensus verification. Although we must enable Shanghai and Cancun in dBFT, this PR is a hint of what should be changed wrt Clique/Ethash implementation for this;
- https://github.com/ethereum/go-ethereum/pull/28549 GitHub actions are added for tests;
- https://github.com/ethereum/go-ethereum/pull/27766 Beacon Committee chain implementation (light chain that is able to verify signed blocks once synced);
    
TODO:

- [x] Fix building workflow
- [x] Create a separate PR for master branch update (?)
- [x] Decide whether we need to sync to the current master or to the latest stable revision (v1.13.11).

PR #136, opened 2024-02-20
core: increase the size limit of header RLP for chain accessor

AnnaShaleva:
Note: this PR is not for merge, since I've just discovered that this problem is already solved on Geth master (although the fix itself is not for our problem originally) and fresh bane-main.

**Problem**

Chain skeleton sync (batched headers sync) is impossible with the current block size. Once the new stale peer is trying to sync its headers, the following errors occurs during skeleton filling (some useless logs are omitted):
```
DEBUG[02-20|14:00:46.499] Fetching batch of headers                id=8205ce1a7d69b885 conn=staticdial      count=1  fromnum=4       skip=0  reverse=false
DEBUG[02-20|14:00:46.908] Fetching batch of headers                id=8205ce1a7d69b885 conn=staticdial      count=1  fromnum=2       skip=0  reverse=false
DEBUG[02-20|14:00:47.153] Fetching batch of headers                id=8205ce1a7d69b885 conn=staticdial      count=1  fromnum=1       skip=0  reverse=false
DEBUG[02-20|14:00:47.426] Found common ancestor                    peer=8205ce1a number=0       hash=000000..000000
DEBUG[02-20|14:00:47.426] Directing header downloads               peer=8205ce1a origin=1
TRACE[02-20|14:00:47.427] Fetching skeleton headers                peer=8205ce1a count=192 from=1
DEBUG[02-20|14:00:47.427] Fetching batch of headers                id=8205ce1a7d69b885 conn=staticdial      count=128 fromnum=192     skip=191 reverse=false
DEBUG[02-20|14:00:47.427] Downloading receipts                     origin=1
DEBUG[02-20|14:00:47.427] Downloading block bodies                 origin=1
DEBUG[02-20|14:00:48.255] Filling up skeleton                      from=1
TRACE[02-20|14:00:48.256] Requesting new batch of headers          peer=8205ce1a from=1
DEBUG[02-20|14:00:48.256] Fetching batch of headers                id=8205ce1a7d69b885 conn=staticdial      count=192 fromnum=1       skip=0   reverse=false
...
TRACE[02-20|14:00:48.657] Skeleton filling not accepted            peer=8205ce1a7d69b885 from=1
DEBUG[02-20|14:00:48.657] Failed to deliver retrieved headers      peer=8205ce1a         err="delivery not accepted"
TRACE[02-20|14:00:48.657] Requesting new batch of headers          peer=8205ce1a         from=193
DEBUG[02-20|14:00:48.657] Fetching batch of headers                id=8205ce1a7d69b885 conn=staticdial      count=192 fromnum=193     skip=0   reverse=false
...
```

So there's a "delivery not accepted" error on attempt to build the skeleton from received batch of received headers. Extended logs give us that every batch of blocks contain only one header (the last from requested batch):
```
TRACE[02-20|14:00:48.656] Delivering headers                       accepted=false len(headers)=1 headers[0].Hash=5bb77b..06997d headers[0].Number=192 MaxHeaderFetch=192
```
So based on this information it's impossible to build the skeleton

**Explaination**

The problem is traced down to the node's peer. During headers retrieval peer fetches the last block of the batch from its cache and tries to retrieve the rest of them from ancients. Given the old maximum RLP header size constraint of 700 bytes per header, the peer can't retrieve the whole set of requested headers (MaxHeaderFetch, 192 headers each at least of 1006 bytes). Since the number of retrieved headers doesn't match the requested one, chain accessor returns only the first header that was retrieved from cache:
```
	// read remaining from ancients
	max := count * 700
	data, err := db.AncientRange(ChainFreezerHeaderTable, i+1-count, count, max)
	if err == nil && uint64(len(data)) == count {
		// the data is on the order [h, h+1, .., n] -- reordering needed
		for i := range data {
			rlpHeaders = append(rlpHeaders, data[len(data)-1-i])
		}
	}
	return rlpHeaders
```

**Solution**

Since `ReadHeaderRange` function specifies that the caller should limit the number of requested headers to prevent DoS attack, it's safe to remove the maximum bytes constraint.

**Additional notes**

After the fix I found that almost the same code change is present in the original Geth commit ported to NeoX node implementation in #130: https://github.com/bane-labs/go-ethereum/commit/447945e43834fdb24a71885eebc72996ca451f8b. And thus, the problem should not be reproducible on the post-sync NeoX node.

### DIFF
--- a/core/rawdb/accessors_chain.go
+++ b/core/rawdb/accessors_chain.go
@@ -334,8 +334,7 @@ func ReadHeaderRange(db ethdb.Reader, number uint64, count uint64) []rlp.RawValu
 		return rlpHeaders
 	}
 	// read remaining from ancients
-	max := count * 700
-	data, err := db.AncientRange(ChainFreezerHeaderTable, i+1-count, count, max)
+	data, err := db.AncientRange(ChainFreezerHeaderTable, i+1-count, count, 0)
 	if err == nil && uint64(len(data)) == count {
 		// the data is on the order [h, h+1, .., n] -- reordering needed
 		for i := range data {


AnnaShaleva:
@roman-khimov agree to roll this branch on T2? It's needed to be able to sync new node from genesis.

roman-khimov:
There is not a lot of options that we have. OK for T2. Maybe we need a temporary branch for it?

AnnaShaleva:
> Maybe we need a temporary branch for it?

Also thought about it, here we have the same problem like in NeoGo. Let's create a separate branch.

AnnaShaleva:
Changed the base, let's use `t2` branch for T2 testnet. PR can be merged then. `t2` is based on 81ea1cec0e3701ab9e52387be29ada71f4d5dbcf.

roman-khimov:
It'll go away with T3, no big deal.

roman-khimov (approved):

PR #141, opened 2024-02-26
dBFT: recalculate proposed Merkle state if needed (T2)

### LINKED ISSUE #60: Optimize the way how dBFT-accepted blocks are being inserted into chain

### LINKED ISSUE #140: Invalid merkle root of accepted block 128928 in T2
#### System information

Commit hash: originally discovered in early version deployed on T2 (81ea1cec0e3701ab9e52387be29ada71f4d5dbcf) with dBFT proposal verification disabled, but supposed to be present at the current bane-main (7b56bf71af205bb183265e49513e6226d69a9f13) and t2 branches (a4cea5bb6957b3deaa39da10e2e6736487c61dec).

Bug occured on T2 13 Feb 2024.

#### Expected behaviour

Merkle root of accepted block is always valid.

#### Actual behaviour

Somehow Merkle root of accepted block is invalid. Seed node can't import block with invalid Merkle root and is stuck at block 128927 of T2. Consensus nodes can't import block with invalid Merkle root via P2P, but can store this invalid block received via consensus engine. CNs continue new blocks acceptance afterwords.

Caught on T2 with 5 out of 7 CNs alive and single RPC (seed) node. I won't attach the full logs since GH doesn't allow large files, here are only logs around the "broken" block that are enough to recreate the picture:

[cn1_bad-block.log](https://github.com/bane-labs/go-ethereum/files/14375742/cn1_bad-block.log)
[cn2_bad-block.log](https://github.com/bane-labs/go-ethereum/files/14375748/cn2_bad-block.log)
[cn3_bad-block.log](https://github.com/bane-labs/go-ethereum/files/14375749/cn3_bad-block.log)
[cn4_bad-block.log](https://github.com/bane-labs/go-ethereum/files/14375752/cn4_bad-block.log)
[cn5_bad-block.log](https://github.com/bane-labs/go-ethereum/files/14375753/cn5_bad-block.log)
[seed1_bad-block.log](https://github.com/bane-labs/go-ethereum/files/14375839/seed1_bad-block.log)

Nodes genesis configuration file:
[genesis.json](https://github.com/bane-labs/go-ethereum/files/14375913/genesis.json)

AnnaShaleva:
Close #140. Close #60. For T2 only, I'll port changes to bane-main after the current PR is merged.

In this PR:
* Ensure all header's fields that affect miner's state calculations are properly set in `(*DBFT).Prepare`. A part of #140.
* Recalculate Merkle state depending on updated dBFT context information in case of CV. Close #140.
* Optimize accepted block insertion into chain using state calculated during PrepareRequest-level block verification (port the #124). Close #60.

@roman-khimov, this PR doesn't contain miner's "slimming" part that we've discussed, mainly because this part contains some transactions filtering, I'd like to keep this code in the node:
https://github.com/bane-labs/go-ethereum/blob/7b56bf71af205bb183265e49513e6226d69a9f13/miner/worker.go#L855-L873
I've tried to completely move this part to dBFT, but this requires moving a substantial amount of related transaction management logic from worker to the dBFT and the code doesn't look nice. However, after a bit of thinking I've checked the set of fields that affect miner's transactions processing and turns out that in pre-merge it's only Difficulty. So I suggest the following modification of our initial concept:
1. Allow miner to execute transactions initially with `in-turn` difficulty (and the rest of related fields partially set by `(*DBFT).Prepare`). Get some initial state, this state will always be relevant for the 0 dBFT view. If block is accepted in this view, then OK, the state is valid.
2. If CV happens, then recalculate block state with the updated data inside the dBFT. Propose the updated state in PrepareRequest. Any further CVs lead to the same dBFT state recalculation.
3. Always recalculate state on PrepareRequest verification and store it for further block insertion optimisation.

This scheme in future allow us to integrate anti-MEV transactions processing since state recalculation at the last anti-MEV stage may be done using the same API that dBFT uses in this PR. Also, initial transactions state processing made by miner is useful during the first anti-MEV stage when nodes have to agree on unencrypted set of transactions.

What do you think?

@chenquanyu, @nicolegys you're welcomed to review and test.

TODO:
- [ ] Establish privnet test where transaction with DIFFICULTY opcode is being pushed to the network before\after CV. I need to ensure that this code works in the situation described in #140.

### DIFF
--- a/consensus/dbft/block.go
+++ b/consensus/dbft/block.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/state"
 	"github.com/ethereum/go-ethereum/core/types"
 	ecrypto "github.com/ethereum/go-ethereum/crypto"
 	"github.com/nspcc-dev/dbft/block"
@@ -23,6 +24,10 @@ type Block struct {
 	header              *types.Header
 	transactions        []*types.Transaction
 	localSignatureBytes []byte
+
+	// Local data calculated during dBFT block verification. Allowed to be empty.
+	state    *state.StateDB
+	receipts types.Receipts
 }
 
 // Version implements block.Block interface.
--- a/consensus/dbft/blockqueue.go
+++ b/consensus/dbft/blockqueue.go
@@ -6,6 +6,7 @@ import (
 	"sync"
 
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/state"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/log"
 )
@@ -52,7 +53,7 @@ func (bq *blockQueue) SetChain(chain ChainHeaderWriter) {
 // PutBlock routs block either to miner or (if there's no suitable sealing task)
 // directly to blockchain. No block verification is performed, it is assumed that
 // provided block is sealed and valid.
-func (bq *blockQueue) PutBlock(b *types.Block) error {
+func (bq *blockQueue) PutBlock(b *types.Block, state *state.StateDB, receipts []*types.Receipt) error {
 	h := WorkerSealHash(b.Header())
 
 	bq.tasksLock.Lock()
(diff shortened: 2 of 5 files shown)


roman-khimov (approved):

nicolegys:
Seems OK. 
In t2 branch, CNs keep CV after receiving tx with DIFFICULTY opcode, in this pr, they look good.
Let me do more test.

chenquanyu (approved):

PR #145, opened 2024-03-04
.github: push Alltools Docker image to the same repo with Geth

AnnaShaleva:
Don't need a separate Dockerhub repo for Alltools images, let's use the same one that we use to push standalone Geth image.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -127,7 +127,7 @@ jobs:
         if: ${{ (github.event_name == 'release' && github.event.release.target_commitish == 'bane-main') || (github.event_name == 'workflow_dispatch' && github.event.inputs.use_latest_tag == 'true') }}
         run: |
           echo "latest=,developerneox/neox_nodes:latest" >> $GITHUB_OUTPUT
-          echo "alltoolslatest=,developerneox/neox_nodes_alltools:latest" >> $GITHUB_OUTPUT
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -127,7 +127,7 @@ jobs:
         if: ${{ (github.event_name == 'release' && github.event.release.target_commitish == 'bane-main') || (github.event_name == 'workflow_dispatch' && github.event.inputs.use_latest_tag == 'true') }}
         run: |
           echo "latest=,developerneox/neox_nodes:latest" >> $GITHUB_OUTPUT
-          echo "alltoolslatest=,developerneox/neox_nodes_alltools:latest" >> $GITHUB_OUTPUT
+          echo "alltoolslatest=,developerneox/neox_nodes:alltools-latest" >> $GITHUB_OUTPUT
```
roman-khimov, line 130: I think the intention here was to have a small node-only image and a bigger one with the whole set. @steven1227, @gsmachado, do we need this separation or we're fine with a single fat image?

AnnaShaleva: > I think the intention here was to have a small node-only image and a bigger one with the whole set.

Exactly. It’s the way how original Geth ships images.

AnnaShaleva:
I'll take the responsibility to go with the presented version (two images: one is standalone Geth binary, another one is all binaries packed together) unless some other option is requested.

PR #146, opened 2024-03-07
Governance V2

txhsl:
- Vote for a candidate instead of a proposal;
- Only `Governance.sol` and `GovReward.sol` have a V2, others remain compatible.


contracts/solidity/GovernanceV2.sol
```diff
@@ -0,0 +1,404 @@
...
+contract GovernanceV2 is IGovernanceV2 {
+    // the min balance for voting
+    uint public constant MIN_VOTE_AMOUNT = 1 ether;
+    // register fee
+    uint public constant REGISTER_FEE = 1000 ether;
+    // minimum duration of an epoch
+    uint public constant EPOCH_DURATION = 1209600;
+    // GovReward contract
+    address public constant govReward =
+        0x1212000000000000000000000000000000000003;
```
nicolegys, line 49: https://github.com/bane-labs/go-ethereum/blob/d875c4eacf9d045b30811c17d56d67d9311f7d31/privnet/seven/genesis_template.json#L28

inconsistent

contracts/solidity/GovernanceV2.sol
```diff
@@ -0,0 +1,404 @@
...
+    ) internal returns (address[7] memory) {
+        address[7] memory cache = consensusCache[epoch];
+        if (cache[0] == address(0)) {
+            cache = _getConsensus(epoch);
+            consensusCache[epoch] = cache;
+        }
+        return cache;
+    }
+
+    function _getConsensus(
```
nicolegys: `_getConsensu` requires at least seven candidates at each epoch (length of `candidateList` never less than 7), otherwise, it will going wrong. Is this expected?

txhsl: I think we cannot expect it to return a `candidateList` whose length is more or less than 7, but I can make a change for the code to avoid error.

contracts/solidity/GovernanceV2.sol
txhsl: This issue happened on T2 as well, let's come back a bit latter.

contracts/solidity/GovernanceV2.sol
txhsl: Now the code requires at least 7 voted candidates to change epoch, so it won't go wrong.

contracts/solidity/GovernanceV2.sol
```diff
@@ -0,0 +1,425 @@
...
+        for (
+            uint i = claimStartEpochOf[msg.sender];
+            i < epoch - 1;
+            i++
+        ) {
+            _tryGetAndCacheConsensus(i);
+        }
+
+        // reorg candidate list
+        address[] memory candidates = candidateList;
```
chenquanyu: Use EnumerableSet or mapping type to avoid reorg candidate list. This will costs about 600w gas for 1000 records.

contracts/solidity/GovernanceV2.sol
```diff
@@ -0,0 +1,425 @@
...
+    function revokeVote() external;
+
+    // withdraw past vote
+    function voterWithdraw() external;
+
+    // claim rewards for being a consensus member
+    function candidateClaim() external;
+
+    // get the current selected consensus group
+    function getCurrentConsensus() external returns (address[7] memory);
```
AnnaShaleva: The number of CNs should be configurable. For example, via constant those value is set in the corresponding Genesis block allocations.

contracts/solidity/GovernanceV2.sol
```diff
@@ -0,0 +1,425 @@
...
+            address(0x74f4EFFb0B538BAec703346b03B6d9292f53A4CD),
+            address(0x910AD1641B7125Eff746acCdCa1F11148b22f472),
+            address(0xfEf5F250aF14DF73f983cAAb7b1F5002189c42E0),
+            address(0xc51964013acbC6b271FEeCB0feBD9E7A01202930),
+            address(0xC5bbD9652546BC96bE3DEc97a38eE335f7873Dfa),
+            address(0x26F1794B81dF2B832545b8B6bbcA196b82E4fEB1),
+            address(0x0B51369D02e47EE3f143391B837Aa08c31AAA19b)
+        ];
+        consensusCache[0] = initialConsensus;
+    }
```
AnnaShaleva: Why not to setup it via genesis allocations? I don't think it's acceptable to change these constants and recompile the contract for mainnet/testnet deployments (and for private development environments also). We'd better have a single compiled instance of Governance and all necessary settings specified in the `genesis.json` network-specific config file.

contracts/solidity/GovRewardV2.sol
```diff
@@ -0,0 +1,52 @@
...
+interface IGovReward {
+    function getMiners() external view returns (address[7] memory);
+
+    function withdraw() external;
+}
+
+contract GovReward is IGovReward {
+    // governance contact
+    address public constant governance =
+        0x1212000000000000000000000000000000000001;
```
AnnaShaleva, line 36: Need to deal with the final set of addresses.

This particular address should be a Governance or a Governance Proxy?

contracts/solidity/GovernanceV2.sol
```diff
@@ -0,0 +1,425 @@
...
+
+    function _getAndUpdateEpochCount() internal returns (uint) {
+        if ((block.timestamp > lastEpochTime + EPOCH_DURATION && totalVotes[epochCount] >= MIN_TOTAL_VOTE && votedCandidates[epochCount] >= 7) || epochCount == 0) {
+            IGovReward(govReward).withdraw();
+            epochCount += 1;
+            lastEpochTime = block.timestamp;
+        }
+        return epochCount;
+    }
```
AnnaShaleva: @roman-khimov, need your opinion: do we really need time-dependent epoch duration (not height-dependent)? It makes epoch counter less predictable given the fact that dBFT has CVs and doesn't accept blocks strictly according to the SecondsPerBlock setting.

AnnaShaleva (commented):
Just a set of quick thoughts, will properly review the contract later.

contracts/solidity/GovernanceV2.sol
txhsl: Only because it is much easier to read for review, all things will be moved to genesis file later.

contracts/solidity/GovernanceV2.sol
txhsl: That's fine with me, I will update it.

contracts/solidity/GovernanceV2.sol
roman-khimov: That's wrong to me. Epochs must be block-based, we elect nodes to produce blocks so we can expect an elected set of validators to produce some exact number of blocks before it's changed. Time is a tricky thing that can also be abused in various ways (delaying blocks/putting slightly invalid time/pushing the proposal prematurely), block counter can't be.

AnnaShaleva: Agree.

txhsl: So what block time should we use to evaluate the epoch duration, 10 seconds?

roman-khimov: It's network-specific technically, but I'm not sure contract can access this setting (unless it's some kind of genesis parameter again).

txhsl: We can put this setting into genesis as well, just like the consensus size. 😃

AnnaShaleva: We may have a set of special Governance-related contract parameters that will be set during genesis block construction automatically depending on the `dbft` section setting. It's not hard to implement, do we need it?

roman-khimov: What if I'd like to have 1s blocks and 10-block epoch to test something?

AnnaShaleva: OK, then EPOCH_DURATION (in blocks) should be filled manually in the genesis allocations.

contracts/solidity/GovRewardV2.sol
chenquanyu: > This particular address should be a Governance or a Governance Proxy?

This should be Governance Proxy, the implementation contract addresses will be changed if we upgrade the contract in the future.

contracts/solidity/GovernanceV2.sol
chenquanyu: We should use "0x1212000000000000000000000000000000000003", all the contracts exposed to users should start with "0x12120" . The implementation contract addresses start with "0x12121" will be changed if we upgrade the contract in the future.

contracts/solidity/GovernanceV2.sol
txhsl: Done.

contracts/solidity/GovernanceV2.sol
txhsl: https://github.com/bane-labs/go-ethereum/pull/146#discussion_r1519655264, update soon.

contracts/solidity/GovernanceV2.sol
AnnaShaleva: Ref. https://github.com/bane-labs/go-ethereum/issues/149, I'd propose to replace contract-related genesis storage items with contract template-based approach in future.

contracts/solidity/GovRewardV2.sol
```diff
@@ -0,0 +1,52 @@
...
+
+    function safeTransferETH(address to, uint256 value) internal {
+        (bool success, ) = to.call{value: value}(new bytes(0));
+        require(success, "safeTransferETH: ETH transfer failed");
+    }
+}
+
+interface IGovernance {
+    // get current consensus group
+    function getCurrentConsensus() external view returns (address[7] memory);
```
AnnaShaleva: This interface (and below) should also be adjusted, it shouldn't depend on committee size.

contracts/solidity/GovernanceV2.sol
```diff
@@ -0,0 +1,386 @@
...
+        emit CandidateClaim(msg.sender, totalReward);
+    }
+
+    function getCurrentConsensus() public view returns (address[] memory) {
+        return consensusOf[getRealCurrentEpoch() - 1];
+    }
+
+    function _computeConsensus(
+        uint epoch
+    ) internal view returns (address[] memory) {
```
AnnaShaleva: From the dBFT POW we need two APIs: first one that for block `N` returns validators of the epoch block `N` belongs to (it's `getCurrentConsensus` in the current implementation). Second one that for block `N` returns validators for the next block `N+1`, where if `N` is the last block in the current epoch, then validators should be re-calculated for the next epoch based on the Nth state (like `_computeConsensus` currently does) and if `N` is not the last block in the current epoch then validators of `N`th block's epoch should be returned.

Could you implement the second API? `getCurrentConsensus` is OK and useful.

AnnaShaleva: To clearly understand what I'm asking about, please, read the documentation of https://github.com/nspcc-dev/neo-go/blob/b12ef701f0560a9a19e297906d819a6b8401048d/pkg/core/blockchain.go#L2768 and https://github.com/nspcc-dev/neo-go/blob/b12ef701f0560a9a19e297906d819a6b8401048d/pkg/core/blockchain.go#L2778. We need the same APIs to be exposed from Governance contract (`getCurrentConsensus` corresponds to `GetNextBlockValidators`).

txhsl: So I think it is splitting the current `getCurrentConsensus` into 2.

txhsl: I find an issue there, which is about the event sequence.

The epoch change happens with governance reward withdraw, which is summing up transaction fees of the previous epoch and 'Mint' (actually transfer) it from `GovReward` to `Governance`. It is like the [OnPersist](https://github.com/neo-project/neo/blob/master/src/Neo/SmartContract/Native/GasToken.cs#L39) in N3.

When the 'change' transaction is executed, then epoch reward is settled, but if the coinbase transfer happens later than execution, then the reward of the last `N` block may be counted to `N+1` epoch.

contracts/solidity/GovernanceV2.sol
```diff
@@ -0,0 +1,410 @@
...
+        if (block.number > lastEpochHeight + EPOCH_DURATION) {
+            return epochCount + 1;
+        } else {
+            return epochCount;
+        }
+    }
+
+    function getRealCurrentEpoch() public view returns (uint) {
+        return epochCount;
+    }
```
AnnaShaleva: Why do we return epoch count from this method? It's a bit misleading, let's rename it to `getRealCurrentEpochCount`. Because current epoch index is `epochCount -1`, right?

contracts/solidity/GovernanceV2.sol
```diff
@@ -0,0 +1,410 @@
...
+    // the min vote amount to change epoch
+    uint public constant MIN_TOTAL_VOTE = 3000000 ether;
+    // minimum duration of an epoch (in blocks)
+    uint public constant EPOCH_DURATION = 120960;
+    // GovReward contract
+    address public constant govReward =
+        0x1212000000000000000000000000000000000003;
+
+    // counter of epoch index
+    uint public epochCount = 1;
```
AnnaShaleva: Adjust comment please, it's just an epoch counter, not indexes.

contracts/solidity/GovernanceV2.sol
```diff
@@ -0,0 +1,410 @@
...
+    // the last block height when voting starts
+    uint public lastEpochHeight = block.number;
+    // candidate list
+    EnumerableSet.AddressSet internal candidateList;
+    // epoch=>uint
+    mapping(uint => uint) public epochRewards;
```
AnnaShaleva: Here and below it's `epoch index` => `some value`, right?

contracts/solidity/GovernanceV2.sol
```diff
@@ -0,0 +1,410 @@
...
+        emit Exit(msg.sender);
+    }
+
+    function claimRegisterFee() external {
+        // require 2 epochs to exit candidate list, so that the last round of vote can work as expected
+        uint epoch = _getAndUpdateEpochCount();
+        require(epoch > exitEpochOf[msg.sender] + 1, "claim not allowed");
+
+        // reorg candidate list
+        candidateList.remove(msg.sender);
```
AnnaShaleva: Why sender is being removed from candidate list on `claimRegisterFee`? Isn't it more natural to remove it in `exitCandidate`? Especially given the fact that we have a public `getCandidates` that (now) returns potentially unregistered candidate.

contracts/solidity/GovernanceV2.sol
```diff
@@ -0,0 +1,410 @@
...
+
+    function registerCandidate(uint shareRate) external payable {
+        require(msg.value == REGISTER_FEE, "insufficient amount");
+        require(shareRate < 1000, "invalid rate");
+        require(!candidateList.contains(msg.sender), "candidate exists");
+        candidateList.add(msg.sender);
+
+        uint epoch = _getAndUpdateEpochCount();
+        // record register time, share rate and balance
+        registerEpochOf[msg.sender] = epoch;
```
AnnaShaleva: I think that `exitEpochOf` for this candidate should be cleared on registration. Imagine the following situation:
1. Candidate performs registration;
2. Then performs unregistration;
3. Then waits for more than two epochs;
4. Then performs registration one more time;
5. And here's unexpected result: candidate is able to claim register fee without calling `exitCandidate`, without clearing the `registerEpochOf` and without emission of the `Exit` event. Looks like a bug.

contracts/solidity/GovernanceV2.sol
```diff
@@ -0,0 +1,410 @@
...
+        return receivedVotes[candidate][epoch];
+    }
+
+    function getCandidates() public view returns (address[] memory) {
+        return candidateList.values();
+    }
+
+    function registerCandidate(uint shareRate) external payable {
+        require(msg.value == REGISTER_FEE, "insufficient amount");
+        require(shareRate < 1000, "invalid rate");
```
AnnaShaleva, line 117: No minimum required bound? Candidates are going to be selfish, I think if we treat this ability "to share" as NeoX feature, then we need set some minimum required rate. Although it's a subject of discussion.

@roman-khimov, @txhsl, @chenquanyu

contracts/solidity/GovernanceV2.sol
```diff
@@ -0,0 +1,410 @@
...
+        candidateBalanceOf[msg.sender] = msg.value;
+        // set the start point for claim, only if register in epoch 1 can get epoch 0 reward
+        claimStartEpochOf[msg.sender] = epoch > 1 ? epoch : 0;
+        emit Register(msg.sender);
+    }
+
+    function exitCandidate() external {
+        require(registerEpochOf[msg.sender] > 0, "candidate not exists");
+        // delete register time, cannot be voted
+        delete registerEpochOf[msg.sender];
```
AnnaShaleva: `shareRateOf` for the candidate is never cleared. From one side, I understand why: user can call `voterWithdraw` whenever he wants and should be able to get the reward. But it's dangerous, consider the following case:
1. Candidate registers with X value of share rate;
2. Candidate is being voted to be a part of consensus;
3. Candidate works for some time;
4. Candidate unregisters;
5. Candidate registers one more time with Y value of share rate;
6. Voter wants to withdraw his reward from candidate.

The problem is that with the current code share rate for this use will be Y, although some blocks were accepted with X share rate. So we need some kind of history for share rate.

contracts/solidity/GovernanceV2.sol
```diff
@@ -0,0 +1,410 @@
...
+        uint amount = candidateBalanceOf[msg.sender];
+        delete candidateBalanceOf[msg.sender];
+        _safeTransferETH(msg.sender, amount);
+    }
+
+    function vote(address candidateTo) external payable {
+        require(msg.value >= MIN_VOTE_AMOUNT, "insufficient amount");
+        require(registerEpochOf[candidateTo] > 0, "candidate not allowed");
+        // the first person vote in new epoch will pay for update
+        uint currentEpoch = _getAndUpdateEpochCount();
```
AnnaShaleva: This pattern with `_getAndUpdateEpochCount` is going to be a problem for us, because if no one interacts with Governance contract, then we'll stuck in a single epoch (while blocks will be produced). Also, I don't quite like this pattern because users will have to add some additional tx fee (comparing to the test invocation) in order to pay for possible epoch update if their transaction accidentally is accepted at the moment when epoch update is needed. And thirdly, this pattern makes epoch transitions more unpredictable (although it's unpredictable even without this pattern since epoch increment depends on users votes).

Maybe we need to think about an analogue of OnPersist or PostPersist for the Governance contract (or even for genesis contracts). I'm not talking about core modification, but maybe at least some kind of special free system transaction that should manage epoch updates in the start of every block.

@roman-khimov, @txhsl, @chenquanyu and everyone interested, what do you think?

contracts/solidity/GovernanceV2.sol
```diff
@@ -0,0 +1,410 @@
...
+        return epochCount;
+    }
+
+    function _getAndUpdateEpochCount() internal returns (uint) {
+        uint epoch = epochCount;
+        if (
+            block.number > lastEpochHeight + EPOCH_DURATION &&
+            totalVotes[epoch] >= MIN_TOTAL_VOTE &&
+            votedCandidates[epoch] >= CONSENSUS_SIZE
+        ) {
```
AnnaShaleva: I'm really worried about the fact that epoch update is unpredictable with this scheme. For example in N3 we know that every `committeeSize` number of blocks epoch transition occurs. Even if there's not enough votes in the current epoch, stand by committee will be used. But here epoch transition won't happen until enough votes are collected for the next epoch, which potentially makes epoch endless in worst case. It leads to the fact that epoch validators (speaking of non-stand-by, i.e. about some community driven nodes) have to keep their nodes up and running even in case if some of them don't want to be a validator anymore.

@roman-khimov, is it OK?

contracts/solidity/GovernanceV2.sol
```diff
@@ -0,0 +1,410 @@
...
+        } else {
+            return epochCount;
+        }
+    }
+
+    function getRealCurrentEpoch() public view returns (uint) {
+        return epochCount;
+    }
+
+    function _getAndUpdateEpochCount() internal returns (uint) {
```
AnnaShaleva: We need a documentation to this method. It returns the count of "completed" epochs or, in other words, it returns index of currently running epoch (enumeration starts from index 0), right?

contracts/solidity/GovernanceV2.sol
```diff
@@ -0,0 +1,410 @@
...
+            block.number > lastEpochHeight + EPOCH_DURATION &&
+            totalVotes[epoch] >= MIN_TOTAL_VOTE &&
+            votedCandidates[epoch] >= CONSENSUS_SIZE
+        ) {
+            // record gov reward and vote result
+            IGovReward(govReward).withdraw();
+            consensusOf[epoch] = _computeConsensus(epoch);
+            epoch += 1;
+            epochCount = epoch;
+            lastEpochHeight = block.number;
```
AnnaShaleva: s/`lastEpochHeight`/`currentEpochStartHeight`?

contracts/solidity/GovernanceV2.sol
```diff
@@ -0,0 +1,410 @@
...
+        }
+        receivedVotes[candidateTo][currentEpoch] = received + msg.value;
+        totalVotes[currentEpoch] += msg.value;
+
+        emit Vote(msg.sender, candidateTo, msg.value);
+    }
+
+    function revokeVote() external {
+        // revoke will not trigger epoch change
+        uint currentEpoch = getNominalCurrentEpoch();
```
AnnaShaleva: It may trigger epoch change, and it should be allowed to. Consider the following situation:
1. Enough votes are collected for the current epoch N, but `EPOCH_DURATION` number of blocks is not yet accepted counting from the epoch N start;
2. `EPOCH_DURATION` number of blocks is being accepted;
3. Someone sends `revokeVote` transaction; the transaction will remove some votes and make the overall amount of votes less than `MIN_TOTAL_VOTE`, but it's not important, because by the moment of transaction processing new epoch should be started.

So we must start new epoch in the beginning of this method if all conditions are matched. Especially given the fact that votes should not belong to some single epoch: we have adjusted this behaviour in the Governance proposal:
>  ~It is necessary to re-vote during each epoch~
Not necessary, can be vote once and keep the record in the following epochs

contracts/solidity/GovernanceV2.sol
```diff
@@ -0,0 +1,410 @@
...
+        epochRewards[getRealCurrentEpoch()] += msg.value;
+    }
+
+    function getNominalCurrentEpoch() public view returns (uint) {
+        if (block.number > lastEpochHeight + EPOCH_DURATION) {
+            return epochCount + 1;
+        } else {
+            return epochCount;
+        }
+    }
```
AnnaShaleva: We don't need this method given the current scheme of "unpredictable" epoch updates. The only data we have to rely on is "real" epoch. The only method where `getNominalCurrentEpoch` is used is in `revokeVote`, but `revokeVote` doesn't work correctly now, see the comment to `revokeVote`.

contracts/solidity/GovernanceV2.sol
```diff
@@ -0,0 +1,410 @@
...
+    function getRealCurrentEpoch() public view returns (uint) {
+        return epochCount;
+    }
+
+    function _getAndUpdateEpochCount() internal returns (uint) {
+        uint epoch = epochCount;
+        if (
+            block.number > lastEpochHeight + EPOCH_DURATION &&
+            totalVotes[epoch] >= MIN_TOTAL_VOTE &&
+            votedCandidates[epoch] >= CONSENSUS_SIZE
```
AnnaShaleva: These two last lines are not correct. The condition of transition to the next epoch is that we have total votes for *all* epochs more than `MIN_TOTAL_VOTE` and we have enough active voted candidates *not necessary voted during the current epoch*. We've changed this Governance scheme:
> ~~It is necessary to re-vote during each epoch~~
Not necessary, can be vote once and keep the record in the following epochs

contracts/solidity/GovernanceV2.sol
```diff
@@ -0,0 +1,410 @@
...
+        uint amount = votedAmount[msg.sender][currentEpoch];
+
+        uint received = receivedVotes[candidateFrom][currentEpoch];
+        if (received == amount) {
+            votedCandidates[currentEpoch] -= 1;
+        }
+        receivedVotes[candidateFrom][currentEpoch] = received - amount;
+        totalVotes[currentEpoch] -= amount;
+        delete votedTo[msg.sender][currentEpoch];
+        delete votedAmount[msg.sender][currentEpoch];
```
AnnaShaleva: These lines should be adjusted wrt the updated Governance scheme. If user revokes vote, then all his funds from *all* epochs should be transferred back.

contracts/solidity/GovernanceV2.sol
```diff
@@ -0,0 +1,410 @@
...
+        if (voted == 0) {
+            votedTo[msg.sender][currentEpoch] = candidateTo;
+            unclaimedEpochsOf[msg.sender].push(currentEpoch);
+        } else {
+            require(
+                votedTo[msg.sender][currentEpoch] == candidateTo,
+                "only one choice is allowed"
+            );
+        }
+        votedAmount[msg.sender][currentEpoch] = voted + msg.value;
```
AnnaShaleva: This code should also be adjusted wrt the updated Governance scheme. Users don't have to revote every epoch, their previous votes should be kept alive (and should make sense) even if the epoch was updated.

contracts/solidity/GovernanceV2.sol
```diff
@@ -0,0 +1,410 @@
...
+        receivedVotes[candidateFrom][currentEpoch] = received - amount;
+        totalVotes[currentEpoch] -= amount;
+        delete votedTo[msg.sender][currentEpoch];
+        delete votedAmount[msg.sender][currentEpoch];
+        _safeTransferETH(msg.sender, amount);
+
+        emit RevokeVote(msg.sender, candidateFrom, amount);
+    }
+
+    function voterWithdraw() external {
```
AnnaShaleva: This is a complicated one, will review this method after all previous conversations are resolved, because it also affected by the updated Governance scheme.

contracts/solidity/GovernanceV2.sol
```diff
@@ -0,0 +1,410 @@
...
+            // epoch changed in this block, so this is the last block
+            return consensusOf[getRealCurrentEpoch() - 2];
+        } else {
+            return consensusOf[getRealCurrentEpoch() - 1];
+        }
+    }
+
+    function computeNextBlockValidators() external view returns (address[] memory) {
+        return getCurrentConsensus();
+    }
```
AnnaShaleva: About these three APIs: after a bit of thinking I'm going to try to use the currently accepting (in-progress) block state for next consensus recalculation. If I manage to do this, then only `getCurrentConsensus` is required for dBFT, so I'll notify about the progress.

contracts/solidity/GovernanceV2.sol
```diff
@@ -0,0 +1,410 @@
...
+    function _computeConsensus(
+        uint epoch
+    ) internal view returns (address[] memory) {
+        // build up a votes array
+        address[] memory candidates = getCandidates();
+        uint length = candidates.length;
+        uint[] memory votes = new uint[](length);
+        for (uint i = 0; i < length; i++) {
+            votes[i] = receivedVotes[candidates[i]][epoch];
+        }
```
AnnaShaleva: Should consider all active votes, not only from the current epoch.

contracts/solidity/GovernanceV2.sol
txhsl: No, `epochCount` is the current voting epoch, `epochCount - 1` is the last ended voting epoch, which is meanwhile the running consensus. But yep, the name is a bit misleading.

contracts/solidity/GovernanceV2.sol
txhsl: Yes.

contracts/solidity/GovernanceV2.sol
txhsl: Considering all active votes is good, but because of the reward settlement, a revote is necessary.

contracts/solidity/GovernanceV2.sol
txhsl: The share rate cannot be changed after register, selfish candidates will receive few votes, because voters want higher APY. But if we need a bound, it's OK as well.

contracts/solidity/GovernanceV2.sol
txhsl: Yes, seems I forget to add it.

contracts/solidity/GovernanceV2.sol
txhsl: Sure.

contracts/solidity/GovernanceV2.sol
txhsl: When a candidate exits, he may be going to be selected as a CN or already be running as a CN, and that is the reason. We have set a locking period (2 epochs) for candidate to exit, he should behave as normal to get all his stake back.

The result is he is still in the `candidateList`, with a none-zero `exitEpochOf` and a zero `registerEpochOf`, and can no longer receive further votes, until he claims back and gets removed.

contracts/solidity/GovernanceV2.sol
txhsl: The only case that `revokeVote` triggers epoch change is that the revoke fails, it is not acceptable.

contracts/solidity/GovernanceV2.sol
txhsl: No, it updates and returns the current voting epoch, as described in [#L126](https://github.com/bane-labs/go-ethereum/pull/146/files#diff-d7c43a33da859540649b7f05b9d432afa2c1508d8c8c4d7d83dcce30a4929bdeR126).

contracts/solidity/GovernanceV2.sol
txhsl: The code is preventing voters to revoke after an epoch should be ended. So the problem here is the `if` condition.

contracts/solidity/GovernanceV2.sol
txhsl: [9d2c6a9](https://github.com/bane-labs/go-ethereum/pull/146/commits/9d2c6a9d0798ac8094094182b1e40021cb818a70).

contracts/solidity/GovernanceV2.sol
```diff
@@ -0,0 +1,422 @@
...
+            // epoch changed in this block, so this is the last block
+            return consensusOf[getCurrentRunningEpoch() - 1];
+        } else {
+            return consensusOf[getCurrentRunningEpoch()];
+        }
+    }
+
+    function computeNextBlockValidators() external view returns (address[] memory) {
+        return getCurrentConsensus();
+    }
```
AnnaShaleva: These two methods may be safely removed, I've adopted state-based validators computations, see https://github.com/bane-labs/go-ethereum/pull/147. We'll use only `getCurrentConsensus` in the dBFT.

And let's clarify (just in case, to be 100% sure): in the current contract version `getCurrentConsensus` returns the consensus members for the latest epoch for which voting process has been finished. So if now voting epoch 5 is in progress and users perform their votes in epoch 5, then `getCurrentConsensus` will return validators computed from the state of epoch 4, right?

txhsl: Sure, it will.

contracts/solidity/GovernanceV2.sol
chenquanyu: We may not need revote but need to traverse the epoch list, this will cost more gas with epoch list grows larger. Please ref. [https://github.com/bane-labs/go-ethereum/blob/82a8b49c14a19e0a1845f1c9756af53cfe49d675/contracts/solidity/Governance.sol#L201](url)

contracts/solidity/GovernanceV2.sol
```diff
@@ -0,0 +1,388 @@
...
+        candidateList.add(msg.sender);
+        delete exitEpochOf[msg.sender];
+
+        uint epoch = _getAndUpdateEpochCount();
+        // record register time, share rate and balance
+        registerEpochOf[msg.sender] = epoch;
+        shareRateOf[msg.sender] = shareRate;
+        candidateBalanceOf[msg.sender] = msg.value;
+        // set the start point for claim, only if register in epoch 1 can get epoch 0 reward
+        claimStartEpochOf[msg.sender] = epoch > 1 ? epoch : 0;
```
nicolegys: `registerCandidate` doesn't check whether the sender has unclaimed candidate rewards, it overrides the value of `claimStartEpochOf`. If someone forgets to claim the candidate reward and register again, he won't be able to claim the previous award.

contracts/solidity/GovernanceV2.sol
```diff
@@ -0,0 +1,388 @@
...
+    ) public view returns (uint) {
+        return receivedVotes[candidate][epoch];
+    }
+
+    function getCandidates() public view returns (address[] memory) {
+        return candidateList.values();
+    }
+
+    function registerCandidate(uint shareRate) external payable {
+        require(msg.value == REGISTER_FEE, "insufficient amount");
```
nicolegys: `insufficient amount` is inappropriate, because more than `REGISTER_FEE` is not allowed, too.

contracts/solidity/GovernanceV2.sol
nicolegys: If a vote tx makes all three of these conditions satisfied, it still cannot trigger epoch change. Have to wait for the next tx which calls `_getAndUpdateEpochCount()`..

https://github.com/bane-labs/go-ethereum/blob/c17b90dbaaf218153baf15c635090dbe6fa4d1b2/contracts/solidity/GovernanceV2.sol#L121-L125

txhsl: > If a vote tx makes all three of these conditions satisfied, it still cannot trigger epoch change. Have to wait for the next tx which calls `_getAndUpdateEpochCount()`..
> 
> https://github.com/bane-labs/go-ethereum/blob/c17b90dbaaf218153baf15c635090dbe6fa4d1b2/contracts/solidity/GovernanceV2.sol#L121-L125

Why?

txhsl:
@AnnaShaleva @roman-khimov @chenquanyu @nicolegys I rewrite the GovernanceV2 in a somehow N3 native contract way, any review is welcome.

contracts/solidity/GovernanceV2.sol
```diff
@@ -0,0 +1,342 @@
...
+
+contract GovernanceV2 is IGovernanceV2 {
+    using EnumerableSet for EnumerableSet.AddressSet;
+
+    // GovReward contract
+    address public constant govReward =
+        0x1212000000000000000000000000000000000003;
+    uint public constant SCALE_FACTOR = 10 ** 18;
+
+    uint public CONSENSUS_SIZE;
```
chenquanyu: We often use capital letters to name constants. I think it is better to use camel case naming for these variables and add methods that can modify the variables. Or we can put these configurable parameters into the policy contract.

contracts/solidity/GovernanceV2.sol
```diff
@@ -0,0 +1,342 @@
...
+    mapping(address => mapping(uint => uint)) public epochStartGasPerVote;
+
+    receive() external payable {
+        address[] memory validators = currentConsensus;
+        uint length = validators.length;
+        for (uint i = 0; i < length; i++) {
+            candidateGasPerVote[validators[i]] +=
+                (msg.value * shareRateOf[validators[i]] * SCALE_FACTOR) /
+                CONSENSUS_SIZE /
+                1000 /
```
chenquanyu, line 100: I suggest putting 1000 as the constant RATE_FACTOR.

contracts/solidity/GovernanceV2.sol
txhsl: In genesis file first, I'll change the naming.

contracts/solidity/GovernanceV2.sol
```diff
@@ -0,0 +1,342 @@
...
+
+    function claimReward() external {
+        address votedCandidate = votedTo[msg.sender];
+        require(votedCandidate != address(0), "claim not allowed");
+        _settleReward(msg.sender, votedCandidate);
+    }
+
+    function postPersist() external {
+        // NOTE: suppose postPersist always happens when equal
+        require(
```
chenquanyu: When the height is not reached this will show up as a failed transaction, I think an empty "return" is more suitable for this case.

AnnaShaleva: That's right, we just need an immediate return in this case.

contracts/solidity/GovernanceV2.sol
```diff
@@ -0,0 +1,342 @@
...
+        emit Revoke(msg.sender, candidateFrom, amount);
+    }
+
+    function claimReward() external {
+        address votedCandidate = votedTo[msg.sender];
+        require(votedCandidate != address(0), "claim not allowed");
+        _settleReward(msg.sender, votedCandidate);
+    }
+
+    function postPersist() external {
```
AnnaShaleva: This method is supposed to be called in the start of every block, and thus I'd suggest to rename it to `onPersist` to make it similar to N3. Then `block.number` here is a number of currently processing block.

contracts/solidity/GovernanceV2.sol
```diff
@@ -0,0 +1,342 @@
...
+
+    function _settleReward(address voter, address candidate) internal {
+        IGovReward(govReward).withdraw();
+
+        uint height = voteHeight[voter];
+        uint lastGasPerVote = voterGasPerVote[voter];
+        uint latestGasPerVote = candidateGasPerVote[candidate];
+
+        // NOTE: suppose postPersist always happens in the correct block at expected height
+        // NOTE: suppose postPersist always happens at the beginning of a block, then vote in that block should wait another epoch to farm reward
```
AnnaShaleva: That's true, the contract is safe to rely on that since it's a system call.

contracts/solidity/GovernanceV2.sol
AnnaShaleva: In the start of this method we need to check that this method is called by special system address `0xfffffffffffffffffffffffffffffffffffffffe` (like check witness in NeoN3). It's needed to disallow side calls.

contracts/solidity/GovernanceV2.sol
```diff
@@ -0,0 +1,342 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import "@openzeppelin/contracts/utils/structs/EnumerableSet.sol";
```
AnnaShaleva, line 4: On my local machine this contract needs openzeppelin dependency initialization to compile (run `npm install @openzeppelin/contracts` before compilation). We need to include a script that will fetch the contract dependency automatically.

Or maybe I'm not aware of something, how do you compile this contract?

contracts/solidity/GovernanceV2.sol
```diff
@@ -0,0 +1,342 @@
...
+
+    /*
+        The following should only be used by DBFT module, refer to https://github.com/nspcc-dev/neo-go/blob/master/pkg/core/blockchain.go
+    */
+
+    // get the consensus group before postPersist, which is the group that should produce this block
+    function getNextBlockValidators() external returns (address[] memory);
+
+    // select the latest consensus group after postPersist, which is the group that should produce the next block
+    function computeNextBlockValidators() external returns (address[] memory);
```
AnnaShaleva: Remove these two.

contracts/solidity/GovernanceV2.sol
AnnaShaleva: OK, going without bound for now. May be introduced later if needed.

contracts/solidity/GovernanceV2.sol
```diff
@@ -0,0 +1,342 @@
...
+            // record tag value
+            votedTo[msg.sender] = candidateTo;
+            voterGasPerVote[msg.sender] = candidateGasPerVote[candidateTo];
+        }
+
+        // update votes
+        votedAmount[msg.sender] += msg.value;
+        receivedVotes[candidateTo] += msg.value;
+        // NOTE: the left reward in current epoch will be unclaimable
+        voteHeight[msg.sender] = block.number;
```
AnnaShaleva: It's OK to be unclaimable if candidate is not yet a voter in the current epoch. But it should be claimable if a candidate is already a voter in the current epoch. It's important since we have quite long epochs.

contracts/solidity/GovernanceV2.sol
```diff
@@ -0,0 +1,342 @@
...
+        uint latestGasPerVote = candidateGasPerVote[candidate];
+
+        // NOTE: suppose postPersist always happens in the correct block at expected height
+        // NOTE: suppose postPersist always happens at the beginning of a block, then vote in that block should wait another epoch to farm reward
+        uint voteEpochEndGasPerVote = epochStartGasPerVote[candidate][
+            (height - 1) / epochDuration + 1
+        ];
+        if (voteEpochEndGasPerVote > lastGasPerVote) {
+            lastGasPerVote = voteEpochEndGasPerVote;
+        }
```
AnnaShaleva, line 252: Why do you bound reward settlement to the epochs? It should be bound to the blocks number and should not depend on epochs, exactly like in N3. We don't need the "history" of gas per vote for candidates. We only need to store latest gas per vote for a particular candidate and update this value once per epoch.

contracts/solidity/GovernanceV2.sol
```diff
@@ -0,0 +1,342 @@
...
+    // voter=>height
+    mapping(address => uint) public voteHeight;
+    // candidate=>height=>number
+    mapping(address => mapping(uint => uint)) public epochStartGasPerVote;
+
+    receive() external payable {
+        address[] memory validators = currentConsensus;
+        uint length = validators.length;
+        for (uint i = 0; i < length; i++) {
+            candidateGasPerVote[validators[i]] +=
```
AnnaShaleva, line 97: It's not correct to update this value on every `receive` call. Candidate's GasPerVote should be updated in the end of every epoch in PostPersist, and then the next epoch reward per vote will be a fixed value.

Also, why do we ever need this general `receive` function in Governance contract? Governance contract should prohibit side transfers. It should support only `vote` and `registerCandidate` payable methods. "Miner" block reward is distributed to the `GovernanceReward` contract without call to `receive`, and in `Governance` contract in the end of every epoch in PostPersist we need an access to `GovernanceReward` balance to manage candidate's GasPerVote updates.

contracts/solidity/GovernanceV2.sol
AnnaShaleva: I think it's OK.

contracts/solidity/GovernanceV2.sol
```diff
@@ -0,0 +1,342 @@
...
+            "withdraw not allowed"
+        );
+
+        // send back balance
+        uint amount = candidateBalanceOf[msg.sender];
+        delete candidateBalanceOf[msg.sender];
+        delete exitHeightOf[msg.sender];
+        _safeTransferETH(msg.sender, amount);
+        emit CandidateWithdraw(msg.sender, amount);
+    }
```
AnnaShaleva: We need to clear LastGasPerVote for those candidates that exited committee and doesn't have votes anymore. Currently candidate's LastGasPerVote is never cleaned which results in growing contract storage. It should be done exactly like in N3, ref. https://github.com/neo-project/neo/blob/b05501af882a0d1f2a1a7841c6ddc4d0504e5fc1/src/Neo/SmartContract/Native/NeoToken.cs#L162.

I agree that it's kind of optimisation and we can live without it, but not forever.

AnnaShaleva (commented):
The storage scheme changes are going in the right direction, now the storage scheme looks similar to Neo native contract in N3, and LastGasPerVote/LatestLastGasPerVote help a lot, but some issues are left.

Also, if needed, both OnPersist and PostPersist calls can be implemented on the backend side so that the contract can rely on these calls. Currently it's only OnPersist which is called before the in-block transactions processing, so let me know if PostPersist is needed.

genesis.json
```diff
@@ -34,21 +34,26 @@
...
+        "0x0000000000000000000000000000000000000000000000000000000000000004": "0x01d880",
+        "0x0000000000000000000000000000000000000000000000000000000000000005": "0x01",
+        "0x13649b2456f1b42fef0f0040b3aaeabcd21a76a0f3f5defd4f583839455116e8": "0x07",
+        "0x178c34144aa4f22f9dcdc9a6606881972cf65c835163e066ef62f7c270683763": "0x74f4effb0b538baec703346b03b6d9292f53a4cd",
+        "0x178c34144aa4f22f9dcdc9a6606881972cf65c835163e066ef62f7c270683764": "0x910ad1641b7125eff746accdca1f11148b22f472",
+        "0x178c34144aa4f22f9dcdc9a6606881972cf65c835163e066ef62f7c270683765": "0xfef5f250af14df73f983caab7b1f5002189c42e0",
+        "0x178c34144aa4f22f9dcdc9a6606881972cf65c835163e066ef62f7c270683766": "0xc51964013acbc6b271feecb0febd9e7a01202930",
+        "0x178c34144aa4f22f9dcdc9a6606881972cf65c835163e066ef62f7c270683767": "0xc5bbd9652546bc96be3dec97a38ee335f7873dfa",
+        "0x178c34144aa4f22f9dcdc9a6606881972cf65c835163e066ef62f7c270683768": "0x26f1794b81df2b832545b8b6bbca196b82e4feb1",
+        "0x178c34144aa4f22f9dcdc9a6606881972cf65c835163e066ef62f7c270683769": "0x0b51369d02e47ee3f143391b837aa08c31aaa19b"
```
AnnaShaleva: @chenquanyu, @chenquanyu, are you sure that this is the correct set of storage items? Because I've compiled the presented Governance contract, put the resulting binaries and ABI to 7-nodes privnet settings and got unexpected result from `getCurrentConsensus` call:
```
00000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000000
```
Whereas with the previous contract version (with validators set in the contract code) I've deployed the previous version to Remix and took all storage items from it to properly fill the genesis allocations. So call to `getCurrentConsensus` worked OK with the previous version and specified genesis allocations.

Could you attach the updated Governance contract code and ABI?

contracts/solidity/GovernanceV2.sol
AnnaShaleva: OK, maybe both cases are correct, since votes recalculation happens once per epoch. But users won't be happy with the fact that their votes are rewarded starting from the next epoch.

contracts/solidity/GovernanceV2.sol
txhsl: I'm using Hardhat, `import "@openzeppelin/contracts/utils/structs/EnumerableSet.sol";` is importing the contract from npm dependency. We can use `import "https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/structs/EnumerableSet.sol";` as well, since it is importing the contract through http. Both of them should work, but I think the current one is suitable because it already works in most of the cases (Hardhat, Solc and Remix).

genesis.json
chenquanyu: These storage items are out of date, I suggest we keep the `constructor()` method for testing purposes.

txhsl: Will get updated after I solve all above conversations.

contracts/solidity/GovernanceV2.sol
txhsl: Done. Well, I was thinking about distribute the reward of the last block of an epoch correctly, so called it `postPersist`. But it seems not reasonable to withdraw gas fees from `GovReawrd` after other txs were executed, because `onPersist` or `postPersist` itself is a tx as well.

contracts/solidity/GovernanceV2.sol
chenquanyu: If we only store latest gas per vote for candidate, then the user who votes both in this epoch and last epoch with different vote amount will lose the reward in last epoch voting.

contracts/solidity/GovernanceV2.sol
txhsl: No, rewarding votes from the next epoch is what we planned to. But I see you mean user need to gather reward immediately but not claimable until `onPersist`.

contracts/solidity/GovernanceV2.sol
txhsl: > Also, why do we ever need this general receive function in Governance contract? 

Just be easy to handle `msg.value`, `receive` will not work if we only settle in the end of epochs, going to be an empty method that only receives transfer.

contracts/solidity/GovernanceV2.sol
AnnaShaleva: > because onPersist or postPersist itself is a tx as well.

Well, it's actually not a "honest" tx. It's a magic executable message that doesn't pay any fees (at least in the current version of #147).

Again, we can do both `onPersist` and `postPersist` if needed. Do you need this?

contracts/solidity/GovernanceV2.sol
AnnaShaleva: > rewarding votes from the next epoch is what we planned to

It's a way to go, so let's keep this version.

txhsl: @AnnaShaleva There will be a problem if we only settle `candidateGasPerVote` when `onPersist`. The goal we have is to make voter stake and lock their GAS into governance and gather rewards. But I can vote in the end of an epoch and leave immediately when the epoch changes. I guess the block validators correctly so that I can share the whole reward of the last epoch, and only 1 block is necessary to do that.

contracts/solidity/GovernanceV2.sol
AnnaShaleva: > then the user who votes both in this epoch and last epoch 

During the second vote handling reward for the first vote will be sent to the user. Then latest GasPerVote for user's candidate should be updated and stored in the `user=>LatestGasPerVote` mapping. Once last epoch is ended, in onPersist candidate's GasPerVote will be updated wrt the received rewards per the whole last epoch. Then, after next epoch start, if candidate decide to claim, then proper reward will be send to the user for one fully completed epoch (but not for the currently going epoch). And I think it's OK.

contracts/solidity/GovernanceV2.sol
txhsl: No, what we need in the further will be an "honest" tx. A system call can be executed and change the statedb as we expected, but it may make it hard to prove the state change if we involve ZK to prove the EVM.

So if we can not distribute the reward of the last block of an epoch correctly with a tx, then for further concern, I don't prefer a `postPersist`.

contracts/solidity/GovernanceV2.sol
```diff
@@ -0,0 +1,322 @@
...
+interface IGovReward {
+    function withdraw() external;
+}
+
+contract GovernanceV2 is IGovernanceV2 {
+    using EnumerableSet for EnumerableSet.AddressSet;
+
+    // GovReward contract
+    address public constant govReward =
+        0x1212000000000000000000000000000000000003;
```
AnnaShaleva, line 49: Do we have a Proxy contract for GovernanceReward contract? Like GovernanceProxy contract for Governance contract implementation.

What happens if we'd like to update GovernanceReward contract implementation? Should we update Governance contract implementation after this and put the updated GovernanceReward contract implementation hash to this constant?

contracts/solidity/GovRewardV2.sol
```diff
@@ -0,0 +1,52 @@
...
+        require(msg.sender == governance, "Not governance");
+        _;
+    }
+
+    function getMiners() external view override returns (address[] memory) {
+        return IGovernance(governance).getCurrentConsensus();
+    }
+
+    function withdraw() external onlyGov {
+        TransferHelper.safeTransferETH(governance, address(this).balance);
```
AnnaShaleva: We send reward to Governance Proxy contract, and then Governance Proxy contract will send reward to Governance implementation contract with subsequent `receive()` method call. So as a result the final destination of reward funds is Governance implementation contract (not a Proxy), right?

contracts/solidity/GovernanceV2.sol
```diff
@@ -0,0 +1,322 @@
...
+
+    receive() external payable {
+        require(msg.sender == govReward, "side call not allowed");
+        address[] memory validators = currentConsensus;
+        uint length = validators.length;
+        for (uint i = 0; i < length; i++) {
+            candidateGasPerVote[validators[i]] +=
+                (msg.value * shareRateOf[validators[i]] * scaleFactor) /
+                consensusSize /
+                1000 /
```
AnnaShaleva, line 100: We should create at least basic documentation for reward calculations to show it to users. It should be documented that:
* `shareRateOf` is X/1000, otherwise it's not clear
* register fee amount
* minimum vote amount
* locking period for candidate registration fee
* etc.

Do we need a separate issue for documentation?

contracts/solidity/GovernanceV2.sol
```diff
@@ -0,0 +1,322 @@
...
+        require(
+            exitHeightOf[msg.sender] > 0 &&
+                block.number > exitHeightOf[msg.sender] + 2 * epochDuration,
+            "withdraw not allowed"
+        );
+
+        // send back balance
+        uint amount = candidateBalanceOf[msg.sender];
+        delete candidateBalanceOf[msg.sender];
+        delete exitHeightOf[msg.sender];
```
AnnaShaleva, line 148: Clear `shareRateOf` for this candidate?

contracts/solidity/GovernanceV2.sol
```diff
@@ -0,0 +1,322 @@
...
+    // voter=>height
+    mapping(address => uint) public voteHeight;
+    // candidate=>height=>number
+    mapping(address => mapping(uint => uint)) public epochStartGasPerVote;
+
+    receive() external payable {
+        require(msg.sender == govReward, "side call not allowed");
+        address[] memory validators = currentConsensus;
+        uint length = validators.length;
+        for (uint i = 0; i < length; i++) {
```
AnnaShaleva, line 96: ### Important issue

Add a short path before this cycle: if `msg.value` is 0, then no recalculations are needed, can safely skip this code below.

`msg.value` can be 0 in the following situation: two subsequent in-block transactions perform vote-related actions. First transaction will withdraw reward from GovernanceReward contract, and by the moment of second transaction invocation balance of GovernanceReward is 0.

AnnaShaleva: Or in GovernanceReward contract we may check that balance is not 0 and only then transfer. Change this function:
```
 function withdraw() external onlyGov {
        TransferHelper.safeTransferETH(governance, address(this).balance);
}
```

contracts/solidity/GovernanceV2.sol
```diff
@@ -0,0 +1,322 @@
...
+        currentConsensus = _computeConsensus();
+        emit Persist(currentConsensus);
+    }
+
+    function getCurrentConsensus() public view returns (address[] memory) {
+        return currentConsensus;
+    }
+
+    function _settleReward(address voter, address candidate) internal {
+        IGovReward(govReward).withdraw();
```
AnnaShaleva: ### Important issue
I'm still worried about the fact that we do this GasPerVote validator recalculations on *every user's transaction that changes votes balance*. I'm worried because users have to pay for that side contract-related recalculation actions. And it's unfair, because the first vote-related transaction in block will pay for all recalculations while the rest of transactions won't. Also, it leads to unpredictable transactions fees difference (between test invocation and real invocation).

I propose to perform GasPerVote validators recalculations once per block, in OnPersist, so that users don't have to pay for that. I'm OK with the fact we perform GasPerVote recalculations every block, but not in every vote-changing transaction, and not for users' funds.

AnnaShaleva: So what do you think about it?

AnnaShaleva: Just remove this single line from `_settleReward` and put it to the very beginning of OnPersist:
```solidity
    function onPersist() external {
        require(msg.sender == sysCall, "side call not allowed");

        // Withdraw miner reward from GovReward contract and recalculate validators' GasPerVote.
        IGovReward(govReward).withdraw();

        // Recalculate validators for the next epoch if it's the end of current epoch.
        if (block.number < currentEpochStartHeight + epochDuration) {
            return;
        }
        address[] memory candidates = candidateList.values();
        ...
```

contracts/solidity/GovernanceV2.sol
```diff
@@ -0,0 +1,322 @@
...
+            // record tag value
+            votedTo[msg.sender] = candidateTo;
+            voterGasPerVote[msg.sender] = candidateGasPerVote[candidateTo];
+        }
+
+        // update votes
+        votedAmount[msg.sender] += msg.value;
+        receivedVotes[candidateTo] += msg.value;
+        // NOTE: the left reward in current epoch will be unclaimable
+        voteHeight[msg.sender] = block.number;
```
AnnaShaleva: ### Important issue

```suggestion
        // NOTE: the left reward in the epoch of first vote will be unclaimable.
        if (votedCandidate == address(0)) {
                voteHeight[msg.sender] = block.number;
        }
```

We need to update `voteHeight` only if `votedCandidate == address(0)`. Otherwise voter won't receive reward if he performs second vote in the next epoch. Consider the situation:
1. Voter firstly votes in epoch X.
2. Same User votes in epoch X+1, `_settleReward` is called, no reward for X epoch will be transferred to candidate because for the end of epoch X "the left reward in current epoch will be unclaimable". But some reward will be transferred for the start of epoch X+1, and it's correct.
3. Voter waits for epoch X+2.
4. In epoch X+2 voter performs `claim`, but with the current code `voteHeight` is updated and has value of the block from epoch X+1 (block from step 2), and thus, zero reward will be transferred to candidate for the end of epoch X+1.

I suppose that in the described situation voter should be rewarded for full epoch X+1, not only for the start of epoch X+1.

Do you agree or I'm misunderstanding?

contracts/solidity/GovRewardV2.sol
chenquanyu: No, the `receive()` of Proxy contract will fall back to implement contract with delegate call. Please refer to https://github.com/OpenZeppelin/openzeppelin-contracts/blob/release-v3.4/contracts/proxy/Proxy.sol and
https://solidity-by-example.org/delegatecall.

contracts/solidity/GovernanceV2.sol
chenquanyu: Now the reward contract is not proxy because we need to get consensus list for proxy admin contract. And proxy admin contract can not call normal methods of governance proxy.  We can change the reward contract to proxy if we add another middle contract to call governance proxy contract.

contracts/solidity/GovernanceV2.sol
AnnaShaleva: I understand that this issue is directly related to https://github.com/bane-labs/go-ethereum/issues/152, but I think we'll find a way to properly execute system transactions.

contracts/solidity/GovernanceV2.sol
txhsl: Not yet.

contracts/solidity/GovernanceV2.sol
txhsl: Done.

contracts/solidity/GovernanceV2.sol
txhsl: If `onPersist()` happens at the beginning of every block, then it is a better solution to merge the `withdraw()` into system call. I agree.

contracts/solidity/GovernanceV2.sol
txhsl: > Or in GovernanceReward contract we may check that balance is not 0 and only then transfer. Change this function:

Done.

contracts/solidity/GovernanceV2.sol
txhsl: Yes..., My original idea is to prevent some vote-and-vote-again thing. Suppose a voter stakes the minimum value (in epoch X) to guess a candidate to be selected, and in the next epoch (epoch X+1) stakes much more to farm rewards. This is not user-friendly, I know.

But it seems reasonable to accept the suggestion, since the voter should wait for another epoch (X+2) to revote if he guesses wrong, and the reward remains the same if he guesses correctly. So the difference seems small between staking minimum and staking all.

I agree.

txhsl:
I've tried different settings for compilation (refer to [Optimizer Parameter Runs](https://learnblockchain.cn/docs/solidity/internals/optimizer.html#optimizer-parameter-runs)), and finally set it `2^32-1 = 4294967295`, but Hardhat tests told me there is no difference in execution cost.

Well, maybe someone who knows the reason can leave a comment.

contracts/solidity/GovernanceV2.sol
```diff
@@ -0,0 +1,325 @@
...
+            // record tag value
+            votedTo[msg.sender] = candidateTo;
+            voterGasPerVote[msg.sender] = candidateGasPerVote[candidateTo];
+        }
+
+        // update votes
+        votedAmount[msg.sender] += msg.value;
+        receivedVotes[candidateTo] += msg.value;
+        // NOTE: the left reward in the first epoch of first vote will be unclaimable.
+        if (votedCandidate == address(0)) {
```
chenquanyu, line 176: I suspect this may cause users to obtain rewards exceeding the total amount with the latest amount of X+1 rounds.

contracts/solidity/GovernanceV2.sol
```diff
@@ -0,0 +1,325 @@
...
+    }
+
+    function _settleReward(address voter, address candidate) internal {
+        // NOTE: suppose onPersist always happens at the beginning of every block, then latestGasPerVote is always the latest
+        uint height = voteHeight[voter];
+        uint lastGasPerVote = voterGasPerVote[voter];
+        uint latestGasPerVote = candidateGasPerVote[candidate];
+
+        // NOTE: suppose epoch change always happens at the beginning of a block, then vote in that block should wait another epoch to farm reward
+        uint voteEpochEndGasPerVote = epochStartGasPerVote[candidate][
```
chenquanyu, line 247: This value may be zero in current epoch, need to add check.

contracts/solidity/GovernanceV2.sol
```diff
@@ -0,0 +1,324 @@
...
+            (height - 1) / epochDuration + 1
+        ];
+        if (voteEpochEndGasPerVote > lastGasPerVote) {
+            lastGasPerVote = voteEpochEndGasPerVote;
+        }
+
+        uint reward = (votedAmount[voter] *
+            (latestGasPerVote - lastGasPerVote)) / scaleFactor;
+        voterGasPerVote[voter] = latestGasPerVote;
+        _safeTransferETH(voter, reward);
```
chenquanyu, line 257: Avoid transfering 0 reward.

txhsl: This is not the same case as https://github.com/bane-labs/go-ethereum/pull/146/files/d5992c25c209f4d9ea50710f2e119c21084dddbd#r1536606351, I think. Or if transfer 0 is dangerous.

txhsl:
Let's move further comments back to #7.

contracts/solidity/GovernanceV2.sol
chenquanyu: > This is not the same case as https://github.com/bane-labs/go-ethereum/pull/146/files/d5992c25c209f4d9ea50710f2e119c21084dddbd#r1536606351, I think. Or if transfer 0 is dangerous.

It's not dangerous, but cost more gas.

PR #147, opened 2024-03-11
Initial Governance contract integration

### LINKED ISSUE #14: dBFT stage 4: integrate voting

AnnaShaleva:
Here's a draft that I have right now, several questions need to be decided, they are marked as TODOs in the code. Close #14.

~~The good news is that the code is able to call Governance API and get the current set of validators. The bad news is that sometimes contract call doesn't work as expected, the following error occurs on attempt to unpack the result, I'm investigating it.~~
UPD: the described problem is solved, now Governance calls work as expected. Now I'm working on the rest of less critical TODOs:
- [x] Decide about final contract hash (ref. https://github.com/bane-labs/go-ethereum/pull/146#discussion_r1519684969).
- [x] ~~Automate contract build and genesis allocations pre-filling.~~ Will be done in #149.
- [x] Make contract code network-independent (ref. https://github.com/bane-labs/go-ethereum/pull/146#discussion_r1519655264 and https://github.com/bane-labs/go-ethereum/pull/146#discussion_r1519662507) and update the rest of privnet/testnet configurations afterwards.
- [x] Deal with genesis-based contract call processing.
- [x] Deal with next epoch validators calculations (ref. https://github.com/bane-labs/go-ethereum/pull/146#discussion_r1519970256).
- [x] Add validators cache.


AnnaShaleva:
@chenquanyu, @txhsl welcome to review, but please keep in mind that it's a very early version of integration.

@nicolegys, this PR is not ready for testing yet, I'll notify you as soon as it's ready.

AnnaShaleva:
@roman-khimov, first version is ready for review. The contract version should be updated (currently the very old one is used to test 7-nodes privnet), but the idea of the PR won't be changed if no issues found.

consensus/dbft/dbft.go
```diff
@@ -1589,56 +1610,81 @@ func (c *DBFT) APIs(chain consensus.ChainHeaderReader) []rpc.API {
...
+		return nil, fmt.Errorf("failed to unpack validators: %w", err)
+	}
+	var res = make([]common.Address, len(valSet))
+	for i := range valSet {
+		res[i] = valSet[i]
+	}
+
+	// TODO: roman-khimov, we've agreed that validators are sorted using lexicographic order, but in contract they are
+	// sorted by votes. Need to come to some single solution.
+	slices.SortFunc(res, common.Address.Cmp)
```
chenquanyu, line 1653: Do we need to sort validators in contract? It seems to be enough to sort them here.

core/systemcontracts/contracts.go
```diff
@@ -2,7 +2,10 @@ package systemcontracts
...
..."address"},{"internalType":"uint256","name":"","type":"uint256"}],"name":"unclaimedEpochsOf","outputs":[{"internalType":"uint256","name":"","type":"uint256"}],"stateMutability":"view","type":"function"},{"inputs":[{"internalType":"address","name":"candidateTo","type":"address"}],"name":"vote","outputs":[],"stateMutability":"payable","type":"function"},{"inputs":[{"internalType":"address","name":"","type":"address"},{"internalType":"uint256","name":"","type":"uint256"}],"name":"votedAmount","outputs":[{"internalType":"uint256","name":"","type":"uint256"}],"stateMutability":"view","type":"function"},{"inputs":[{"internalType":"uint256","name":"","type":"uint256"}],"name":"votedCandidates","outputs":[{"internalType":"uint256","name":"","type":"uint256"}],"stateMutability":"view","type":"function"},{"inputs":[{"internalType":"address","name":"","type":"address"},{"internalType":"uint256","name":"","type":"uint256"}],"name":"votedTo","outputs":[{"internalType":"address","name":"","type":"address"}],"stateMutability":"view","type":"function"},{"inputs":[],"name":"voterWithdraw","outputs":[],"stateMutability":"nonpayable","type":"function"},{"stateMutability":"payable","type":"receive"}]`
```
chenquanyu: May be we only need the interface abi here.

consensus/dbft/dbft.go
AnnaShaleva: > Do we need to sort validators in contract?

We don't. But I'd like to ensure, that it's OK for us to sort validators on dBFT engine side.

core/systemcontracts/contracts.go
AnnaShaleva: Yes, it will be updated once we have Governance finalized.

consensus/dbft/dbft.go
AnnaShaleva: It's OK for us, so let this sorting code to be present here. TODO is removed.

core/state_processor.go
```diff
@@ -189,3 +196,28 @@ func ProcessBeaconBlockRoot(beaconRoot common.Hash, vmenv *vm.EVM, statedb *stat
 	_, _, _ = vmenv.Call(vm.AccountRef(msg.From), *msg.To, msg.Data, 30_000_000, common.U2560)
 	statedb.Finalise(true)
 }
+
+// ProcessOnPersist applies a system call to the governance contract.
+func ProcessOnPersist(vmenv *vm.EVM, statedb *state.StateDB) error {
```
chenquanyu, line 201: Now we perform a call to OnPerist method of Governance contract via direct state modification (without fully-qualified transaction creation; for free) before in-block transactions processing. I think we can raise an issue first and we can implement free transactions in the future.

AnnaShaleva: Ref. #152.

AnnaShaleva:
@roman-khimov, ready for review & merge, @nicolegys is performing 7-nodes privnet testing. Should we wait for test results before the merge?

consensus/dbft/dbft.go
```diff
@@ -1593,7 +1601,8 @@ func (c *DBFT) APIs(chain consensus.ChainHeaderReader) []rpc.API {
 // to be sorted by bytes order (even if returned from governance contract).
 func (c *DBFT) getNextBlockValidators(blockHash common.Hash, blockNum uint64, compute bool) ([]common.Address, error) {
 	// Currently we don't have governance contract, thus, always return standby set.
-	if true {
+	// TODO: to be removed.
```
roman-khimov: TODOs in the final commit series?

consensus/dbft/dbft.go
```diff
@@ -1603,26 +1612,33 @@ func (c *DBFT) getNextBlockValidators(blockHash common.Hash, blockNum uint64, co
...
-	}
+	method := "getCurrentConsensus" // current epoch validators
+	/*if compute {
+		// TODO: given time-dependent `EPOCH_DURATION`, calculation of next consensus becomes complicated. We need
+		// `getNextConsensus` API to be presented in Governance contract since `getConsensus` is not enough.
+		// `getConsensus` calculates consensus based on the current state, even if epoch is not finished yet. What we need
+		// is `getNextConsensus` that returns current epoch validators for the middle of dBFT epoch and next epoch validators
+		// for the last block in the epoch. We can organize this at the code-level, but it complicates getNextBlockValidators call, so
+		// why not to move it to the contract level where it costs nothing.
+		// @roman-khimov, agree?
```
roman-khimov: TODOs in the final commit series?

consensus/dbft/dbft.go
```diff
@@ -1603,26 +1612,33 @@ func (c *DBFT) getNextBlockValidators(blockHash common.Hash, blockNum uint64, co
...
+		method = "getNextConsensus" // current epoch validators for the middle of dBFT epoch and next epoch validators for the last block in epoch
+	}*/
 
 	ctx, cancel := context.WithCancel(context.Background())
-	// Cancel when we are finished consuming integers
+	// Cancel when we are finished consuming integers.
```
roman-khimov: Mixed with other changes.

consensus/dbft/dbft.go
```diff
@@ -1632,9 +1648,19 @@ func (c *DBFT) getNextBlockValidators(blockHash common.Hash, blockNum uint64, co
 		return nil, err
 	}
 
-	var valSet []common.Address
+	var valSet [7]common.Address
```
roman-khimov: Can we get this number from the contract?

consensus/dbft/dbft.go
```diff
@@ -455,9 +451,16 @@ func New(config *params.DBFTConfig, _ ethdb.Database) (*DBFT, error) {
 				log.Crit("failed to finalize and assemble proposal",
 					"err", err)
 			}
-
```
roman-khimov, line 456: Whitespace.

consensus/dbft/dbft.go
```diff
@@ -573,6 +573,27 @@ func New(config *params.DBFTConfig, _ ethdb.Database) (*DBFT, error) {
...
 				return false
 			}
+
+			// Verify NextConsensus based on the state got after in-block transactions processing. Make a
+			// state copy in order to avoid state modifications potentially made by getValidators call.
+			// The original state will be committed if block is accepted.
+			// TODO: technically, the contract itself doesn't change the state in getCurrentValidators
+			// (at least, at the current version), and we don't need to create a copy. But I'm not sure
+			// whether processing engine makes changes in the state, it should be investigated in a separate
+			// issue. And if not, then state may not be copied.
```
roman-khimov: Number?

internal/ethapi/api.go
```diff
@@ -1138,6 +1138,12 @@ func DoCall(ctx context.Context, b Backend, args TransactionArgs, blockNrOrHash
 	return doCall(ctx, b, args, state, header, overrides, blockOverrides, timeout, globalGasCap)
 }
 
+func DoCallAtState(ctx context.Context, b Backend, args TransactionArgs, state *state.StateDB, header *types.Header, overrides *StateOverride, blockOverrides *BlockOverrides, timeout time.Duration, globalGasCap uint64) (*core.ExecutionResult, error) {
```
roman-khimov, line 1143: Undocumented?

core/systemcontracts/contracts.go
```diff
@@ -1,11 +1,18 @@
...
...tputs":[{"internalType":"uint256","name":"","type":"uint256"}],"stateMutability":"view","type":"function"},{"inputs":[{"internalType":"address","name":"candidateTo","type":"address"}],"name":"vote","outputs":[],"stateMutability":"payable","type":"function"},{"inputs":[{"internalType":"address","name":"","type":"address"}],"name":"voteHeight","outputs":[{"internalType":"uint256","name":"","type":"uint256"}],"stateMutability":"view","type":"function"},{"inputs":[{"internalType":"address","name":"","type":"address"}],"name":"votedAmount","outputs":[{"internalType":"uint256","name":"","type":"uint256"}],"stateMutability":"view","type":"function"},{"inputs":[{"internalType":"address","name":"","type":"address"}],"name":"votedTo","outputs":[{"internalType":"address","name":"","type":"address"}],"stateMutability":"view","type":"function"},{"inputs":[{"internalType":"address","name":"","type":"address"}],"name":"voterGasPerVote","outputs":[{"internalType":"uint256","name":"","type":"uint256"}],"stateMutability":"view","type":"function"},{"inputs":[],"name":"withdrawRegisterFee","outputs":[],"stateMutability":"nonpayable","type":"function"},{"stateMutability":"payable","type":"receive"}]`))
```
roman-khimov: You can do this in `init` and panic.

roman-khimov:
We need to test voting that changes the set of validators, btw.

nicolegys:
CNs failed when trying to change epoch.

```
CRIT [03-27|08:16:07.046] failed to process state from proposal    err="failed to process block: could not apply OnPersist [0x1774aa018de852e473c1d5cf332fc22ebfe7af960b8d7eb64370e9fa20c4354d]: onPersist call failed: execution reverted" number=12 "seal hash"=c98122..e91b71 "parent hash"=0x40f5b34ede642c91ea99563ef4f0bbe7d384604fa4c6d2db4c39df4e2e7bc517 "intermediate merkle root"=047e7d..96635e coinbase=0x1212000000000000000000000000000000000003 "gas limit"=30,000,000 "gas used"=0 difficulty=1 "mix digest"=0x0000000000000000000000000000000000000000000000000000000000000000 nonce=3 time=1,711,527,227 "uncle hash"=0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347 txs=0
```

txhsl:
> CNs failed when trying to change epoch.

Have you tried to register at least 7 different addresses as candidates before epoch change? The related contract method is https://github.com/bane-labs/go-ethereum/blob/governance/contracts/solidity/Governance.sol#L119.

AnnaShaleva:
> CNs failed when trying to change epoch.

Node expects the contract to properly handle `onPersist` call, and if it's not the case, then the node have to stop operation. The case is that it's Primary node who is trying to construct PrepareRequest, and if the construction fails, then we're toast.

> Have you tried to register at least 7 different addresses as candidates before epoch change?

So I'd expect the contract to handle the case when not enough candidates are registered. In this case StandBy validators should be returned. Because we will have such cases in the real network.

consensus/dbft/dbft.go
AnnaShaleva: These TODOs are resolved in the subsequent commits. Fixed.

Just FYI, we have a set of similar commits in NeoGo :D

consensus/dbft/dbft.go

consensus/dbft/dbft.go
AnnaShaleva: Moved to a separate commit.

consensus/dbft/dbft.go
AnnaShaleva: It was a kind of an intermediate contract version, it's fixed in the later commits. But now when we have a relatively stable contract, we don't need a code for intermediate version, so removed.

consensus/dbft/dbft.go
AnnaShaleva: Fixed.

consensus/dbft/dbft.go
AnnaShaleva: [Number](https://github.com/bane-labs/go-ethereum/issues/154).

internal/ethapi/api.go
AnnaShaleva: I followed the existing `DoCall` style, it also doesn't have a comment. But probably we may write better code. Fixed.

core/systemcontracts/contracts.go
AnnaShaleva: Fixed.

roman-khimov (approved):

PR #148, opened 2024-03-12
workflows: upload build artifacts only on release or manual trigger

AnnaShaleva:
We're hitting GitHub storage limits, so reduce the number of artifacts to be uploaded.


roman-khimov (approved):
https://github.com/nspcc-dev/gh-push-to-neofs/

PR #166, opened 2024-04-16
systemcontracts: apply baseFee from policy contract for burning

### LINKED ISSUE #163: governance:  burning fees function.
We have not yet added the function of burning fees. We can send part of fees to an empty address in the governance contract to burn and configure this burning ratio in the policy contract.

_Originally posted by @chenquanyu in https://github.com/bane-labs/go-ethereum/issues/7#issuecomment-2049059397_

### LINKED ISSUE #186: Implement `minGasTipCap` as minimum effective gas tip
# Rationale

Currently, we have the `minGasTipCap` setting in Policy and check it in both transaction pool and execution, but it is not fully working as we expect.

After the EIP-1559 upgrade, the `--miner.gasprice` setting actually became a `--min.gastipcap` thing, refer to https://github.com/bane-labs/go-ethereum/blob/bane-main/cmd/geth/main.go#L435. Due to a dynamic base fee, Geth is not able to get an accurate `effectiveGasTip` when accepting a new transaction into mempool, so it is now checking the `tx.GasTipCap`, which is the maximum gas tip that the miner can get. 

Unfortunately, this checking has obvious shortages:

1. A legacy transaction only declares `GasPrice` without `GasFeeCap` and `GasTipCap`, and Geth set the `GasPrice` value to `GasFeeCap` and `GasTipCap`. Refer to https://github.com/bane-labs/go-ethereum/blob/bane-main/core/types/tx_legacy.go#L101-L102. Since the base fee is always higher than 0, the `effectiveGasTip` is always lower than the transaction pool expects;
2. A dynamic fee transaction can declare a low `GasFeeCap` but high `GasTipCap` (e.g. set the same value), then refer to https://github.com/bane-labs/go-ethereum/blob/bane-main/core/state_transition.go#L462, it is still possible that the `effectiveGasTip` is lower than the transaction pool expects.

Miners will lose more when the base fee becomes higher, so it is encouraged in EIP-1559 to create a new block as soon as possible. This design could be helpful in PoW, but we don't need this in dBFT, because CNs has no reason to delay but generate a new block every 10s. After #166, we will not have the dynamic burning mechanism but a fixed burning value. So the `minGasTipCap` setting should become a `effectiveGasTip` checking.

# Implementation

After #166, we can get a long-term fixed `basefee` value from the Policy contract or the latest block, so we can compute an accurate `effectiveGasTip` when accepting a new transaction into mempool and prevent the exact final tip to be lower than the `minGasTipCap` policy (maybe we rename it to `minGasTip`).

chenquanyu:
Close #163
Add the `baseFee` reading from policy contract for `eip1559.CalcBaseFee()`. 

After this PR the `BaseFee`  field in block header will be the `baseFee` in policy contract, so this will cause hard fork for current testnet. I have added `neoxburnTime` field in genesis config, `neoxburnTime` should be set to a future timestamp and the nodes need to be re-init if we want to upgrade the testnet.

I have tested the upgrade process in privnet seven node, the `BaseFee` in block header changes after the `neoxburnTime` timestamp and transaction need to pay at least `baseFee` + `minGasTipCap` in policy contract.


steven1227:
Ready for reviewing and testing.

consensus/dbft/dbft.go
```diff
@@ -903,6 +903,12 @@ func (c *DBFT) verifyCascadingFields(chain consensus.ChainHeaderReader, header *
 	if header.GasUsed > header.GasLimit {
 		return fmt.Errorf("invalid gasUsed: have %d, gasLimit %d", header.GasUsed, header.GasLimit)
 	}
+
+	state, err := c.chain.StateAt(parent.Root)
+	if err != nil {
+		return fmt.Errorf("failed to get state: err %w, root %s", err, parent.Root)
+	}
+
```
AnnaShaleva: We can't do this, it's too invasive and it's not a proper place for this verification. `verifyCascadingFields` belongs to the common engine verification method (`VerifyHeader`), and should not access the node state. Consider running this code on a light node.

dBFT doesn't has its own snapshot, its verification mechanism is different from Clique. Ordinary nodes should just ensure that this block is signed by proper consensus nodes (this check is already present). And consensus nodes must ensure the base fee correctness during PrepareRequest state-dependent verification (inside the `WithVerifyBlock` callback).

consensus/misc/eip1559/eip1559.go
```diff
@@ -54,12 +56,20 @@ func VerifyEIP1559Header(config *params.ChainConfig, parent, header *types.Heade
...
 	}
 
+	// Read baseFee from policy if baseFee is set in policy contract
+	if state != nil {
+		baseFee := state.GetState(systemcontracts.PolicyProxyHash, systemcontracts.GetBaseFeeStateHash()).Big()
+		if baseFee.Cmp(common.Big0) > 0 {
+			return baseFee
+		}
+	}
+
```
AnnaShaleva: Also can't do this here since `CalcBaseFee` should be state-independent. We need to move it inside the state-dependent verification.

AnnaShaleva (commented):
I think we need to rework the code, there's a lot of places where the block state is being fetched although I'd say it shouldn't.

consensus/misc/eip1559/eip1559.go
chenquanyu: But now we get `baseFee` from Policy contract, we need `state` to verify the `baseFee`. Adding a method to get `baseFee` from `state` is no different from the current implementation.

consensus/dbft/dbft.go
chenquanyu: OK, I will move this check to `WithVerifyBlock`.

consensus/misc/eip1559/eip1559.go
AnnaShaleva: > we need state to verify the baseFee. 

That's why this check can't be present at `VerifyEIP1559Header`. Headers verification should be state-independent.

consensus/dbft/dbft.go
AnnaShaleva: > will move this check to WithVerifyBlock.

I think that this check should be moved to state-dependent transition verification. With this change no additional code is required in `WithVerifyBlock`.

core/blockchain.go
```diff
@@ -2471,6 +2472,12 @@ func (bc *BlockChain) ProcessState(block *types.Block) (*state.StateDB, types.Re
 		return nil, nil, nil, 0, fmt.Errorf("failed to retrieve state at %d, %s: %w", parent.NumberU64(), parent.Root(), err)
 	}
 
+	// verify baseFee
+	baseFee := statedb.GetState(systemcontracts.PolicyProxyHash, systemcontracts.GetBaseFeeStateHash()).Big()
+	if block.BaseFee().Cmp(baseFee) != 0 {
+		return nil, nil, nil, 0, fmt.Errorf("failed to verify policy baseFee, expected: %v, current: %v", baseFee, block.BaseFee())
+	}
+
```
AnnaShaleva: I think it'll be better to move this check to `(st *StateTransition) preCheck()` since this check is needed not only for dBFT blocks processing but also in the process of normal block persist cycle.

chenquanyu: `(st *StateTransition) preCheck()` is for message level check and only dBFT blocks need this check because others are check in `VerifyEIP1559Header`.

consensus/clique/clique.go
```diff
@@ -357,7 +357,7 @@ func (c *Clique) verifyCascadingFields(chain consensus.ChainHeaderReader, header
 		if err := misc.VerifyGaslimit(parent.GasLimit, header.GasLimit); err != nil {
 			return err
 		}
-	} else if err := eip1559.VerifyEIP1559Header(chain.Config(), parent, header); err != nil {
+	} else if err := eip1559.VerifyEIP1559Header(chain.Config(), parent, chain.CurrentHeader()); err != nil {
```
roman-khimov: Why? I thought we're checking `header` specifically here.

core/blockchain.go
roman-khimov: I don't think we have any other blocks but dBFT in Bane. And looks like this change follows the practice of "check during consensus, trust in use" for dBFT header fields (consider state-dependent `NextConsensus`, for example).

consensus/misc/eip1559/eip1559.go
```diff
@@ -54,12 +56,20 @@ func VerifyEIP1559Header(config *params.ChainConfig, parent, header *types.Heade
...
 // CalcBaseFee calculates the basefee of the header.
-func CalcBaseFee(config *params.ChainConfig, parent *types.Header) *big.Int {
+func CalcBaseFee(config *params.ChainConfig, parent *types.Header, state *state.StateDB) *big.Int {
 	// If the current block is the first EIP-1559 block, return the InitialBaseFee.
 	if !config.IsLondon(parent.Number) {
 		return new(big.Int).SetUint64(params.InitialBaseFee)
 	}
 
+	// Read baseFee from policy if baseFee is set in policy contract
+	if state != nil {
```
roman-khimov: I see it can be `nil` in your code, but I really doubt it should ever be this way.

core/txpool/legacypool/legacypool.go
```diff
@@ -1281,7 +1281,14 @@ func (pool *LegacyPool) runReorg(done chan struct{}, reset *txpoolResetRequest,
 		pool.demoteUnexecutables()
 		if reset.newHead != nil {
 			if pool.chainconfig.IsLondon(new(big.Int).Add(reset.newHead.Number, big.NewInt(1))) {
-				pendingBaseFee := eip1559.CalcBaseFee(pool.chainconfig, reset.newHead)
+				state, err := pool.chain.StateAt(reset.newHead.Root)
+				if err != nil {
+					state, err = pool.chain.StateAt(types.EmptyRootHash)
+				}
+				if err != nil {
+					log.Error("Get state error", "err", err, "root", reset.newHead.Root)
```
roman-khimov: Is it safe to continue here?

privnet/seven/genesis_privnet.json
```diff
@@ -15,6 +15,7 @@
     "arrowGlacierBlock": 0,
     "grayGlacierBlock": 0,
     "shanghaiTime": 0,
+    "neoxburnTime": 0,
```
roman-khimov: I hope it's testnet-only.

roman-khimov (commented):
Looks legit otherwise.

consensus/clique/clique.go
chenquanyu: Sorry, here is unnecessary change.

consensus/misc/eip1559/eip1559.go
chenquanyu: This file is changed in latest code, please refresh.

privnet/seven/genesis_privnet.json
chenquanyu: It's for t3 testnet hard fork, but for mainnet we should also include this to apply `baseFee` change.

core/txpool/legacypool/legacypool.go
chenquanyu: Sorry, I check the code again and find that the state is saved in `reset(oldHead, newHead *types.Header)`, so we can just use `pool.currentState` here.

privnet/seven/genesis_privnet.json
steven1227: It's better to keep this, same as `shanghaiTime` above. Can change a better name

consensus/misc/eip1559/eip1559.go
```diff
@@ -93,3 +95,32 @@ func CalcBaseFee(config *params.ChainConfig, parent *types.Header) *big.Int {
...
+	if err := misc.VerifyGaslimit(parentGasLimit, header.GasLimit); err != nil {
+		return err
+	}
+	// Verify the header is not malformed
+	if header.BaseFee == nil {
+		return errors.New("header is missing baseFee")
+	}
+	// Verifing the baseFee is moved to VerifyBlock.
+	return nil
+}
```
AnnaShaleva: I don't quite like the massive code duplication between `VerifyEIP1559HeaderDBFT` and `VerifyEIP1559Header`. Base fee verification and calculation is bound to the `IsNeoxburn` hardfork, thus can we keep the `VerifyEIP1559Header` and just move the last part of it under `if config.IsNeoxburn` condition? Like this:
```go
	...
	// Verify the baseFee is correct based on the parent header. If neoxburn hardfork enabled, then
	// base fee verificaiton is state-dependent and performed by state-dependent dBFT block verification.
	if config.IsNeoxburn {
		return nil
	}
	expectedBaseFee := CalcBaseFee(config, parent)
	if header.BaseFee.Cmp(expectedBaseFee) != 0 {
		return fmt.Errorf("invalid baseFee: have %s, want %s, parentBaseFee %s, parentGasUsed %d",
			header.BaseFee, expectedBaseFee, parent.BaseFee, parent.GasUsed)
	}
	...
```

consensus/misc/eip1559/eip1559.go
```diff
@@ -93,3 +95,32 @@ func CalcBaseFee(config *params.ChainConfig, parent *types.Header) *big.Int {
...
+
+// CalcBaseFeeDBFT calculates the basefee of the header.
+// if is neoxburn fork, get basefee from Policy contract.
+func CalcBaseFeeDBFT(config *params.ChainConfig, parent *types.Header, state *state.StateDB) *big.Int {
+	if !config.IsNeoxburn(parent.Number, parent.Time) {
+		return CalcBaseFee(config, parent)
+	}
+	return state.GetState(systemcontracts.PolicyProxyHash, systemcontracts.GetBaseFeeStateHash()).Big()
+}
```
AnnaShaleva: We're replacing all the usages of `CalcBaseFee` by `CalcBaseFeeDBFT`. The better way to me is to remove `CalcBaseFeeDBFT` function, add `state *state.StateDB` argument to the original `CalcBaseFee` and move this `config.IsNeoxburn`-dependent logic under `CalcBaseFee`. This approach allows us to easily rebase onto fresh Geth changes without unexpected missings. It also allows to ensure that all usages of `CalcBaseFee` are updated properly.

core/blockchain.go
```diff
@@ -2471,6 +2472,12 @@ func (bc *BlockChain) ProcessState(block *types.Block) (*state.StateDB, types.Re
 		return nil, nil, nil, 0, fmt.Errorf("failed to retrieve state at %d, %s: %w", parent.NumberU64(), parent.Root(), err)
 	}
 
+	// verify baseFee
+	baseFee := eip1559.CalcBaseFeeDBFT(bc.chainConfig, parent.Header(), statedb)
+	if block.BaseFee().Cmp(baseFee) != 0 {
+		return nil, nil, nil, 0, fmt.Errorf("failed to verify policy baseFee, expected: %v, current: %v", baseFee, block.BaseFee())
+	}
+
```
AnnaShaleva: To me, it's more a part of `(bc *BlockChain) VerifyBlock`, because this check is not related to the state processor itself.

Also, dBFT uses ProcessState during PrepareRequest creation where this verification code shouldn't be executed.

core/state_transition.go
```diff
@@ -331,6 +324,16 @@ func (st *StateTransition) preCheck() error {
 				return fmt.Errorf("%w: address %v, maxFeePerGas: %s, baseFee: %s", ErrFeeCapTooLow,
 					msg.From.Hex(), msg.GasFeeCap, st.evm.Context.BaseFee)
 			}
+			// Ensure the transaction is allowed by policy, only for neoxburn fork
+			if st.evm.ChainConfig().IsNeoxburn(st.evm.Context.BlockNumber, st.evm.Context.Time) {
+				// Apply policy minimum gas tip cap
+				// For LegacyTx, GasFeeCap and GasPrice are equal, so checking GasTipCap and GasFeeCap is enough
+				var minGasTipCap = st.state.GetState(systemcontracts.PolicyProxyHash, systemcontracts.GetMinGasTipCapStateHash()).Big()
+				if cmath.BigMin(msg.GasTipCap, new(big.Int).Sub(msg.GasFeeCap, st.evm.Context.BaseFee)).Cmp(minGasTipCap) < 0 {
+					return fmt.Errorf("%w: address %v, gastipcap %v, gasfeecap %v", ErrUnderpriced,
```
AnnaShaleva: We need to add `minGasTipCap` required by Policy to the error message.

core/txpool/validation.go
```diff
@@ -203,9 +204,11 @@ func ValidateTransactionWithState(tx *types.Transaction, signer types.Signer, op
 	}
 	// Ensure the transaction is allowed by policy
 	// Apply policy minimum gas tip cap
+	// For LegacyTx, GasFeeCap and GasPrice are equal, so checking GasTipCap and GasFeeCap is enough
 	var minGasTipCap = opts.State.GetState(systemcontracts.PolicyProxyHash, systemcontracts.GetMinGasTipCapStateHash())
-	if tx.GasTipCap().Cmp(minGasTipCap.Big()) < 0 {
-		return fmt.Errorf("%w: policy needed %v, tip permitted %v", ErrUnderpriced, minGasTipCap.Big(), tx.GasTipCap())
+	var baseFee = opts.State.GetState(systemcontracts.PolicyProxyHash, systemcontracts.GetBaseFeeStateHash()).Big()
+	if math.BigMin(tx.GasTipCap(), new(big.Int).Sub(tx.GasFeeCap(), baseFee)).Cmp(minGasTipCap.Big()) < 0 {
+		return fmt.Errorf("%w: policy needed %v, gastipcap %v, gasfeecap %v ", ErrUnderpriced, minGasTipCap.Big(), tx.GasTipCap(), tx.GasFeeCap())
```
AnnaShaleva: Need to add `baseFee` value to the error message, otherwise it may be not clear why transaction is declined.

eth/gasprice/feehistory.go
```diff
@@ -83,7 +83,11 @@ func (oracle *Oracle) processBlock(bf *blockFees, percentiles []float64) {
 		bf.results.baseFee = new(big.Int)
 	}
 	if chainconfig.IsLondon(big.NewInt(int64(bf.blockNumber + 1))) {
-		bf.results.nextBaseFee = eip1559.CalcBaseFee(chainconfig, bf.header)
+		state, _, err := oracle.backend.StateAndHeaderByNumber(context.Background(), rpc.BlockNumber(bf.blockNumber))
+		if err != nil {
+			log.Error("Failed to get state", "err", err, "header number", bf.blockNumber)
+		}
+		bf.results.nextBaseFee = eip1559.CalcBaseFeeDBFT(chainconfig, bf.header, state)
```
AnnaShaleva: We are not supposed to use node state inside Oracle, and it's a design problem. Have you tried to test `FeeHistory` call on the light-weighted (state-less) nodes? It likely won't work, because here we refer to the node state. Correct me if I'm wrong, please.

We need a BaseFee calculation scheme that allows to compute next block base fee without referring to the parent's state. Ideally, it would be nice to have have something similar to NextConsensus logic: it's included in the parent's block, and we don't need an access to state to verify the next block, we only need a parent's header for it.

miner/worker.go
```diff
@@ -951,7 +951,12 @@ func (w *worker) prepareWork(genParams *generateParams) (*environment, error) {
 	}
 	// Set baseFee and GasLimit if we are on an EIP-1559 chain
 	if w.chainConfig.IsLondon(header.Number) {
-		header.BaseFee = eip1559.CalcBaseFee(w.chainConfig, parent)
+		state, err := w.chain.StateAt(parent.Root)
+		if err != nil {
+			return nil, err
+		}
+		// Calculate baseFee with config
+		header.BaseFee = eip1559.CalcBaseFeeDBFT(w.chainConfig, parent, state)
```
AnnaShaleva: State fetching is a quite resource-consuming operation, we need to try to minimize it and reuse an existing state where possible. Miner has its own state that is created inside the `env, err := w.makeEnv(parent, header, genParams.coinbase)` a few lines below. Can we reuse it? And then update calculated BaseFee value where needed.

params/config.go
```diff
@@ -351,6 +351,7 @@ type ChainConfig struct {
 	CancunTime   *uint64 `json:"cancunTime,omitempty"`   // Cancun switch time (nil = no fork, 0 = already on cancun)
 	PragueTime   *uint64 `json:"pragueTime,omitempty"`   // Prague switch time (nil = no fork, 0 = already on prague)
 	VerkleTime   *uint64 `json:"verkleTime,omitempty"`   // Verkle switch time (nil = no fork, 0 = already on verkle)
+	NeoxburnTime *uint64 `json:"neoxburnTime,omitempty"` // Neoxburn switch time (nil = no fork, 0 = already on neoxburn)
```
AnnaShaleva: Please, rename it to `NeoXBurnTime` for the field and to `neoXBurnTime` for the JSON.

privnet/seven/genesis_privnet.json
AnnaShaleva: Please, add the corresponding entry to the Testnet config. Let's schedule this hardfork in ~2 weeks from now and then reschedule if needed. We need to review, test and merge all pending changes by this time.

AnnaShaleva: > I hope it's testnet-only.

For mainnet it will be enabled starting from 0, but we still need to include this entry to the config (given the fact that the current testnet should be supported with non-zero `neoxburnTime`). @roman-khimov, are you against of adding this entry to the mainnet config?

cmd/evm/internal/t8ntool/execution.go
```diff
@@ -156,6 +157,13 @@ func (pre *Prestate) Apply(vmConfig vm.Config, chainConfig *params.ChainConfig,
 		GasLimit:    pre.Env.GasLimit,
 		GetHash:     getHash,
 	}
+	env := &pre.Env
+	env.BaseFee = eip1559.CalcBaseFeeDBFT(chainConfig, &types.Header{
+		Number:   new(big.Int).SetUint64(env.Number - 1),
+		BaseFee:  env.ParentBaseFee,
+		GasUsed:  env.ParentGasUsed,
+		GasLimit: env.ParentGasLimit,
+	}, statedb)
```
AnnaShaleva: Not sure about this change, I need some more time to check it.

eth/gasprice/feehistory.go
AnnaShaleva: @chenquanyu, @chenquanyu, @roman-khimov this is the main issue I'd like to clarify (and may be fix).

>And looks like this change follows the practice of "check during consensus, trust in use" for dBFT header fields (consider state-dependent NextConsensus, for example).

That's true, so the only component that should *compute* the base fee from the state is a consensus engine. The problem is I see this state-dependent code in some other components like Oracle or Eth API which is not quite correct to me.

roman-khimov: IIUC there are cases when we need base fee for the next block outside consensus, like test executions.

roman-khimov: If you're proposing something like `NextBaseFee` added to the block header, it might be interesting.

core/blockchain.go
AnnaShaleva: Also we do retrieve parent state one more time a few lines below during state transition. We need to pay more attention to our computational resources, otherwise we'll significantly slow down the node.

eth/gasprice/feehistory.go
AnnaShaleva: > IIUC there are cases when we need base fee for the next block outside consensus, like test executions.

But the problem is not all of these cases are powered to access the blockchain state. Currently light-weighted node compatibility is not tested at all, and I'm just afraid that we'll break it completely.

> something like NextBaseFee added to the block header,

Yes, something like that, because I don't see another option to solve the mentioned problem.

consensus/misc/eip1559/eip1559.go
chenquanyu: I added `state *state.StateDB` argument before, but changed for your comment https://github.com/bane-labs/go-ethereum/pull/166#discussion_r1576134355

core/blockchain.go
chenquanyu: I add the verify here because the `statedb` is retrieved here, we will not  retrieve parent state one more time.

core/txpool/validation.go
chenquanyu: Fixed.

core/state_transition.go
chenquanyu: Fixed.

eth/gasprice/feehistory.go
chenquanyu: > If you're proposing something like `NextBaseFee` added to the block header, it might be interesting.

Do you mean adding the `NextBaseFee` field to the block header, or adding `NextBaseFee` to the `Extra` field of the block header? Currently we have not added fields to the block header, which may affect other tools to parse the block.

privnet/seven/genesis_privnet.json
chenquanyu: Will change the Testnet config when the Testnet is ready to update.

miner/worker.go
chenquanyu: Yes, we can reuse it. Fixed.

params/config.go
chenquanyu: Done.

eth/gasprice/feehistory.go
txhsl: Why don't we use `header.BaseFee` here?
https://github.com/bane-labs/go-ethereum/blob/fe076a097f83605e32961846309a0ecc8c8b141a/eth/gasprice/feehistory.go#L301-L306
What about change the length of `basefee` to `blocks`, and try to append the latest one in `abi_backend`.

core/blockchain.go
AnnaShaleva: > we will not retrieve parent state one more time.

That's true, seems that I misread the code, so the second problem is not actually a problem. However, the first problem still exist. The thing is: this check is not a part of state processing. When we do state processing, we assume that header has already passed verification and is valid. So I'd place this check to `VerifyBlock`, after body validation but before state processing. We can get the state to verify it and then pass the same state to `bc.ProcessState` as a second argument. At the same time, `bc.ProcessState` need to check whether provided state is nil, and if it is, then retrieve state by itself.

eth/gasprice/feehistory.go
AnnaShaleva: > What about change the length of basefee to blocks

I'm afraid we can't. It's user-facing behaviour and this extra value is documented to be returned, dependent applications rely on it.

cmd/evm/internal/t8ntool/execution.go
AnnaShaleva: `t8ntool` is an environment for Eth developers to simulate transactions/blocks processing. It uses configuration file with *some* contract storage data and *some* previous block data. Usually these data include only those contracts that a developer is interested in.

This change requires Policy contract and its storage items to be added to t8ntool configuration to work properly with our BaseFee-dependent logic. My concern is no one knows about this additional logic except us. I suggest two options:
1. Adjust documentation of t8ntool, add examples of exact values that need to be added to the t8ntool launch configuration and explain why developers ever need to take care of Policy contract storage values. Otherwise we break t8ntool and no one knows what exactly should be put into configuration to make it work.
2. Replace this code with a check that user-configured `pre.Env` value is set. Warn if it's not set and warn about the consequences of it. If it's set, then manually add this value to the state DB (to Policy storage). This code should likely be a part of `MakePreState`. This option also requires documentation adjustment.

The second option looks much more preferable to me. Bane developers do not need to put any Policy-cpecific values into their t8ntool configuration. Everything they need to do is to fill in the `currentBaseFee` field.

Either 1 or 2 requires t8ntool test (run it with some real config and try to process a transaction), we can't merge this change without any tests, otherwise there's a risk to break this tool.

consensus/misc/eip1559/eip1559.go
AnnaShaleva: Oh, I see, my bad. But the current way also doesn't seem to be right, we replacing all the usages of CalcBaseFee by CalcBaseFeeDBFT anyway. Let me take more detailed look at it.

core/blockchain.go
chenquanyu: OK, got it.

eth/gasprice/feehistory.go
chenquanyu: Light client has been removed from geth, so it's OK to get `state` from `oracle.backend` now.

cmd/evm/internal/t8ntool/execution.go
AnnaShaleva: The second option is implemented in #230.

consensus/misc/eip1559/eip1559.go
AnnaShaleva: In #230 I tried to reduce the number of usages of CalcBaseFeeDBFT, so there's not a lot of them left. One thing that prevent us from CalcBaseFee and CalcBaseFeeDBFT unification is testing code and `ChainMaker`, they still use an old version (CalcBaseFee) that is not capable of returning Policy-based value. We need to decide if we'd like to do something about it.

consensus/misc/eip1559/eip1559.go
AnnaShaleva: Implemented in #230.

AnnaShaleva (commented):
Once #230 is merged, we need to adjust Policy contract documentation, it's out-of-date: https://github.com/bane-labs/go-ethereum/tree/bane-main/contracts#policy

core/txpool/validation.go
```diff
@@ -203,9 +204,12 @@ func ValidateTransactionWithState(tx *types.Transaction, signer types.Signer, op
 	}
 	// Ensure the transaction is allowed by policy
 	// Apply policy minimum gas tip cap
-	var minGasTipCap = opts.State.GetState(systemcontracts.PolicyProxyHash, systemcontracts.GetMinGasTipCapStateHash())
-	if tx.GasTipCap().Cmp(minGasTipCap.Big()) < 0 {
-		return fmt.Errorf("%w: policy needed %v, tip permitted %v", ErrUnderpriced, minGasTipCap.Big(), tx.GasTipCap())
+	// For LegacyTx, GasFeeCap and GasPrice are equal, so checking GasTipCap and GasFeeCap is enough
+	var minGasTipCap = opts.State.GetState(systemcontracts.PolicyProxyHash, systemcontracts.GetMinGasTipCapStateHash()).Big()
+	var baseFee = opts.State.GetState(systemcontracts.PolicyProxyHash, systemcontracts.GetBaseFeeStateHash()).Big()
+	if math.BigMin(tx.GasTipCap(), new(big.Int).Sub(tx.GasFeeCap(), baseFee)).Cmp(minGasTipCap) < 0 {
```
AnnaShaleva, line 210: @txhsl, please, update Policy contract documentation wrt this new functionality.

songb2:
@AnnaShaleva @chenquanyu @txhsl  I've tested metamask and t8ntool with the latest commit on this pr https://github.com/bane-labs/go-ethereum/pull/230. No more issues found.

core/txpool/validation.go
txhsl: Ref #249. I also update it about #216 and #242.

AnnaShaleva (approved):

PR #173, opened 2024-04-17
systemcontract: add documents

txhsl:
Close #172. Add a document for contracts in this repo including GovernanceVote, GovProxyAdmin, Governance and Policy.

PR #182, opened 2024-04-19
systemcontract: add transferVote method

### LINKED ISSUE #165: Allow changing vote target in Governance contract
# Rationale

Current Governance requires 2 steps to change votes from one candidate to another, which needs 2 transactions and twice wallet interactions.

1. Revoke votes from current candidate, and withdraw the stake;
2. Vote to another candidate and stake again.

It can be simplified to improve user experience.

# Implementation

Add a method like `changeVote(address candidateTo)` to allow flexible adjustment on vote target.

txhsl:
Close #165. Need update before merge.


contracts/solidity/Governance.sol
```diff
@@ -29,6 +29,9 @@ interface IGovernance {
     // revoke votes and claim rewards
     function revokeVote() external;
 
+    // revoke votes, claim rewards and vote to another candidate
+    function changeVote(address candidateTo) external;
```
AnnaShaleva: We need to update the contract ABI in the `systemcontracts` package. And luckily for this change we don't need a hardfork to stay compatible with T3.

AnnaShaleva: Should be documented, ref. #173.

AnnaShaleva: Suggest to merge #173 firstly, and after that rebase this PR onto fresh bane-main and adjust documentation.

AnnaShaleva: Maybe s/changeVote/transferVote? Or something like that, naming is always hard.

contracts/solidity/Governance.sol
```diff
@@ -263,6 +266,28 @@ contract Governance is IGovernance, ReentrancyGuard, UUPSUpgradeable {
...
+
+        // settle reward here
+        uint unclaimedReward = _settleReward(msg.sender, candidateFrom);
+
+        // update votes
+        receivedVotes[candidateFrom] -= amount;
+        receivedVotes[candidateTo] += amount;
+        votedTo[msg.sender] = candidateTo;
+        voterGasPerVote[msg.sender] = candidateGasPerVote[candidateTo];
+        voteHeight[msg.sender] = block.number;
```
AnnaShaleva, line 279: I'd suggest not to change the height so that if voter changes the vote target, then it still receives the reward for the rest of the epoch. Exactly like if candidate performs the second vote for the same address.

AnnaShaleva: But it's a subject of discussion.

contracts/solidity/Governance.sol
txhsl: It seems better to update the ABI in this PR, I will do that later.

txhsl: Sure.

contracts/solidity/Governance.sol
txhsl: There is a case that a voter first vote to candidate A whose `shareRate` is low (e.g. 1), then change to another candidate B whose `shareRate` is high (e.g. 1000) after epoch change. That could happen if A is not select as a CN, so the voter leaves because there is no reward. This seems reasonable.

But if the voter change back to A before the next epoch change. There will be a fact that the voter puts its effective votes to A, but shares rewards from B. Is that what the candidate B expects to see?

I suggest that we delay this change till the future, e.g. when we reduce the epoch length.

AnnaShaleva: Agree, your second example is reasonable, we probably need to think about this case a bit more.

> I suggest that we delay this change till the future

Could you create an issue for that?

AnnaShaleva (dismissed):
With additional issue (https://github.com/bane-labs/go-ethereum/pull/182#discussion_r1619176166) this PR LGTM.

Luckily, `governanceABI` does not include method offsets and compatibility between the current Testnet is preserved. But we need to schedule system contracts updates on Testnet after v0.1.2 is released.

AnnaShaleva:
@chenquanyu, @roman-khimov, @nicolegys, ready for review and testing.

contracts/solidity/Governance.sol
```diff
@@ -263,6 +266,28 @@ contract Governance is IGovernance, ReentrancyGuard, UUPSUpgradeable {
         _safeTransferETH(msg.sender, amount + unclaimedReward);
     }
 
+    function transferVote(address candidateTo) external nonReentrant {
```
AnnaShaleva, line 264: @roman-khimov, I think we need an advice on system contracts versioning from your side.

v0.1.0 and v0.1.1 contain `config/genesis_testnet.json` that contains compiled version of system contracts. Now we update one of system contracts and so not change the genesis configuration because we need to be compatible with T3. We'll update Governance manually after v0.1.2 is released.

The problem is: current version of Governance does not match the compiled binary included into genesis config. Is it OK or do we need explicit versioning for system contracts? Like two different (V1 and V2) Governance contracts stored in the repo.

AnnaShaleva:
@txhsl, please, squash the commits into a single one.

contracts/solidity/Governance.sol
roman-khimov: Well, if you start a new network you'd have a new config with a new contract, isn't it?

AnnaShaleva: For new network there's no such problem. The question is only about the current T3 that is going to be long-term maintained.

roman-khimov: It has to have configuration that can be used with the network as is, which means _old_ contract in the genesis in our case.

txhsl: I think there is no need to update the genesis allocation.

AnnaShaleva: > I think there is no need to update the genesis allocation.

I don't suggest it, this will break the T3.

> It has to have configuration that can be used with the network as is

OK, everyone agrees that we don't need to keep an old contract code in the repo. So not a problem.

AnnaShaleva:
@txhsl, please, don't merge the unstructured changes into master. We'll end up in a git history mess in a few months.

txhsl:
Roger then, we only have #184, #216 and #217 left to merge to get a tag version for system contract after audit. I'm managing it.

contracts/solidity/Governance.sol
AnnaShaleva: Ref. #222.

PR #184, opened 2024-04-19
systemcontract: add native treasury for bridge

### LINKED ISSUE #183: Add Native Treasury
Based on some conversations in the Bridge Contract Repo (e.g., [here](https://github.com/bane-labs/bridge-evm-contracts/issues/37) and this following [issue](https://github.com/bane-labs/bridge-evm-contracts/issues/74)), it'd be preferred to have a treasury contract instead of the bridge contract holding most of the initial allocation in the genesis.

This treasury will be highly sensitive, and should have the single purpose of funding the bridge. I suggest the following for the treasury contract:

1. it'll be a native contract holding all* of the initial Gas allocation
2. it'll only have a single method that sends Gas to the bridge proxy
3. invoking this method requires a vote from the Neo X Governance
4. it should (not) be upgradeable (?)

*except for some other needed initial allocations, e.g., the bridge relayer, some starting allocation for the bridge, etc.

**Context from Bridge Contracts Development**
- The bridge management introduced a new role `funder` in https://github.com/bane-labs/bridge-evm-contracts/pull/79.
- The bridge now has a `receive` function that is restricted only to this `funder` role. It will be the only address that is allowed to send Gas to the bridge via simple transfer.
- This new role `funder` is intended to be set to the above described native treasury contract.

Please provide some feedback/input about the above suggestion and also whether it should be upgradeable. In the meantime, I'll draft an implementation for a native treasury contract and will open a PR for it.

@steven1227 @AnnaShaleva @roman-khimov @gsmachado

mialbu:
Closes #183


contracts/solidity/Treasury.sol
```diff
@@ -0,0 +1,33 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.20;
+
+import "./GovernanceVote.sol";
+
+/**
+ * @dev This is an auxiliary contract meant to be assigned as the Neo X treasury for funding the native bridge proxy.
+ * Use GovernanceVote to manage funding the bridge.
+ */
+contract Treasury is GovernanceVote {
```
AnnaShaleva: Quite a straightforward contract, the overall idea to move this logic to a separate contract LGTM, but:
1. Why don't you want it to be upgradeable? To ensure that all Neo X funds are well-protected? Overall I agree, if some additional logic is needed then it may be implemented on the Bridge contract itself.
2. This is an incompatible change, we can't merge it into T3, so we either need one more incompatible testnet or do this on the mainnet start. I'd avoid adding system contracts deploy in hard-forks since it's an essential part of the network and I'd like to have all set of system contracts to be deployed starting from the genesis block.

contracts/solidity/Treasury.sol
```diff
@@ -0,0 +1,33 @@
...
+    error FundingFailed();
+
+    /**
+     * @dev Sends `amount` of ether to the bridge proxy.
+     */
+    function fundBridge(
+        uint256 _amount
+    )
+        external
+        needVote(keccak256("fundBridge"), keccak256(abi.encode(_amount)))
```
AnnaShaleva: It requires more than 1/2 of the validators votes (4 votes out of 7 for T3 and mainnet). It's OK for the system-wide settings, but is it OK for the Bridge funding?

gsmachado: IMHO, it's in the validators' best interest to fund the bridge and organize themselves to make it happen.

Otherwise GAS will not be able to be migrated from N3 to X, and consequently GAS will become scarce in Neo X, which can bring some issues.

So, I assume validators would be (indirectly) incentivized to fund the bridge in tranches, based on Gov voting.

wdyt, @steven1227 @mialbu @txhsl @joeqian10 ?

contracts/solidity/Treasury.sol
gsmachado: > 1. Why don't you want it to be upgradeable? To ensure that all Neo X funds are well-protected? Overall I agree, if some additional logic is needed then it may be implemented on the Bridge contract itself.

Yes. The main goal here is to protect Neo X funds with **deadly simple** smart contract logic and use the same security assumption from the Neo X governance (consensus of the validators) to bring funds to the bridge -- which has more complex logic, and inherent higher risks.

This way we would create another layer of safety to protect against unforeseen situations concerning the bridge.

gsmachado: > 2. This is an incompatible change, we can't merge it into T3, so we either need one more incompatible testnet or do this on the mainnet start. I'd avoid adding system contracts deploy in hard-forks since it's an essential part of the network and I'd like to have all set of system contracts to be deployed starting from the genesis block.

About that ☝️, @mialbu will clarify what's the plan.

I remember we discussed that there's a way to introduce the concept of a "treasury" in testnet t3 without hard-forking. For mainnet we will need to allocate the GAS directly to the treasury, though, in the genesis.

mialbu: > 2\. This is an incompatible change, we can't merge it into T3, so we either need one more incompatible testnet or do this on the mainnet start. I'd avoid adding system contracts deploy in hard-forks since it's an essential part of the network and I'd like to have all set of system contracts to be deployed starting from the genesis block.

I know, you're right. We cannot test this contract as a "native" (active from genesis) contract in T3. However, we can still use and test it in T3. It just requires some additional steps before deploying it:

1. Precompute its contract hash (deployer's address + nonce)
2. Bridge _x_ amount of GAS from N3 testnet to use for this purpose
3. Send this _x_ amount to the precomputed hash
4. Deploy the treasury with the same deployer (can't use the deployer for any transaction inbetween - nonce must remain the same until deployment)

-> We have the treasury contract deployed with some funds that can be used to fund the bridge.

contracts/solidity/Treasury.sol
txhsl: > It's OK for the system-wide settings, but is it OK for the Bridge funding?

It's the same threshold for a governance vote, so at least I think it is safe.

> So, I assume validators would be (indirectly) incentivized to fund the bridge in tranches, based on Gov voting.

The funding itself is not expensive and rarely happens. It is not requiring validators to pay the fund but just a transaction fee, so I don't think it is an issue.

AnnaShaleva (approved):
Vote up from my side, let's track the T3 deploy in https://github.com/bane-labs/go-ethereum/issues/218, the code LGTM and may be included into v.0.1.2.

gsmachado:
Great!

I'd go ahead and merge, @mialbu

AnnaShaleva:
> I'd go ahead and merge

Need the final approve from @txhsl, @chenquanyu and @roman-khimov before the merge.

roman-khimov:
I'm OK with the concept, but can't comment much on the implementation.

contracts/solidity/Treasury.sol
```diff
@@ -0,0 +1,33 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.20;
```
txhsl: Update to `^0.8.25`.

contracts/solidity/Treasury.sol
```diff
@@ -0,0 +1,33 @@
...
+ * @dev This is an auxiliary contract meant to be assigned as the Neo X treasury for funding the native bridge proxy.
+ * Use GovernanceVote to manage funding the bridge.
+ */
+contract Treasury is GovernanceVote {
+    address public constant BRIDGE_PROXY =
+        0x1212000000000000000000000000000000000004;
+
+    event FundBridge(uint256 amount);
+
+    error FundingFailed();
```
txhsl: Move this to the `Errore.sol` library, or just use `TransferFailed`, maybe.

contracts/solidity/Treasury.sol
txhsl: Use a direct hash as method key.

mialbu: Done in [159dafe](https://github.com/bane-labs/go-ethereum/pull/184/commits/159dafe436001c07d8493017b7181b1900cfc1fa).

contracts/solidity/Treasury.sol
mialbu: Done in [a7b6a7d](https://github.com/bane-labs/go-ethereum/pull/184/commits/a7b6a7d0442323ef2565203459a979e9945118e8).

contracts/solidity/Treasury.sol
mialbu: Done in [b8674e0](https://github.com/bane-labs/go-ethereum/pull/184/commits/b8674e0f890d07ada51cbbbb81a1aee03ef82370).

PR #190, opened 2024-04-30
rpc: update fee estimation RPCs to support EIP-1559 tools e.g. MetaMask

### LINKED ISSUE #187: Solve the compatibility problem between EIP-1559 and Policies
# Rationale

Briefly speaking, we are not fully compatible with EIP-1559 after #6 and #166. The compatibility problem between EIP-1559 and Policies are caused by a different economic model. 

- As I've mentioned in #186, EIP-1559 proposed a dynamic fee market scheme with `baseFee` and `gasTip`. Generally, when there is a network traffic, the gas price increases. But if there is no transaction in the network, `baseFee` and `gasTip` may drop to nearly 0;
- But in N3, we don't have a dynamic price model. Instead, we maintain some price factors in the Neo Policy. As a result, the system fee remains the same regardless of traffic, and the only difference is the network fee. 

There are also several facts I've noticed on testnet:

1. It is frequent that CNs create empty blocks, so the `gasPrice`/`baseFee` is pretty low;
2. To achieve a cost about 0.01 GAS per native transfer transaction (costs 21000 gas in EVM), we need a `gasPrice` about `0.01 * 10 ^ 18 / 21000 ≈ 500 Gwei`, which is incredibly higher than other EVM chains, e.g. 7 Gwei on Ethereum, 1 Gwei on BNB Smart Chain, 100 Gwei on Polygon and 0.1 Gwei on Optimistic/Arbitrum;

Empty blocks and high gas price may cause problems in EIP-1559 compatible tools, e.g. MetaMask.

We've already modified some RPC APIs to suite Neo X Policies, like changing `eth_gasPrice` and `eth_maxPriorityFeePerGas` to provide a minimum acceptable value so that a transaction will not be rejected by txpool. But `eth_feeHistory` is somehow conflict with our Policy idea, it returns a fee list of recently accepted transactions. It would be confusing when the list is empty, for EIP-1559 compatible tools, this case equals to a zero `eth_maxPriorityFeePerGas`. Their logics don't consider a minimum value, so when I increase the `minGasTipCap` limit, MetaMask hits RPC errors.

# Implementation

For the MetaMask problem, I found it can be solved by disabling `eth_feeHistory` RPC API. But this compatibility issue may happen on other tools or custom codes. Maybe they use `0.8*eth_maxPriorityFeePerGas` and hit an error.

So we may need a discussion, are we good to continue with this EIP-1559 incompatible? Or we have other ideas about the price thing?

txhsl:
Close #187, After #166.

As mentioned in https://github.com/bane-labs/go-ethereum/pull/6#discussion_r1565128962, our price model (fixed `baseFee` + minimum `effectiveGasTip`) will touch off RPC errors in MetaMask. Because it analyses `eth_feeHistory` directly and suggests a `GasTipCap` lower than we expected. This fallback is removed in [MetaMask/core](https://github.com/MetaMask/core), but not yet the extension.

To prevent this kind of issues in similar tools and libraries, this PR increases `suggestGasTipCap` to 112.5% of the Policy limit, and sets it as the minimum return value of `eth_feeHistory` as well. (The corresponding multiplier of MetaMask ranges from 0.94 to 0.98, which is confusing but meanwhile the fact).

The 12.5% increase in `suggestGasTipCap` will allow us to perform a 12.5% increase in gas tip Policy without rejecting existing transactions in txpool. As for `baseFee`, a 100% increase is allowed with https://github.com/bane-labs/go-ethereum/blob/bane-main/accounts/abi/bind/base.go#L282.

I know it seems much more correct to return the minimum `effectiveGasTip` in `eth_feeHistory` instead of 112.5% of it, but MetaMask doesn't work with it. :(

### DIFF
--- a/eth/api_backend.go
+++ b/eth/api_backend.go
@@ -374,11 +374,25 @@ func (b *EthAPIBackend) SuggestGasTipCap(ctx context.Context) (*big.Int, error) 
 		return nil, err
 	}
 	minGasTipCap := stateDb.GetState(systemcontracts.PolicyProxyHash, systemcontracts.GetMinGasTipCapStateHash()).Big()
-	return cmath.BigMax(suggestTipCap, minGasTipCap), nil
+	denom := new(big.Int).SetUint64(b.eth.blockchain.Config().BaseFeeChangeDenominator())
+	return cmath.BigMax(suggestTipCap, minGasTipCap.Add(minGasTipCap, new(big.Int).Div(minGasTipCap, denom))), nil
 }
 
 func (b *EthAPIBackend) FeeHistory(ctx context.Context, blockCount uint64, lastBlock rpc.BlockNumber, rewardPercentiles []float64) (firstBlock *big.Int, reward [][]*big.Int, baseFee []*big.Int, gasUsedRatio []float64, err error) {
-	return b.gpo.FeeHistory(ctx, blockCount, lastBlock, rewardPercentiles)
+	oldestBlock, reward, baseFee, gasUsedRatio, err := b.gpo.FeeHistory(ctx, blockCount, lastBlock, rewardPercentiles)
+	if err != nil {
+		return nil, nil, nil, nil, err
+	}
+	suggestGasTipCap, err := b.SuggestGasTipCap(ctx)
+	if err != nil {
+		return nil, nil, nil, nil, err
+	}
+	for i, r := range reward {
+		for j, v := range r {
+			reward[i][j] = cmath.BigMax(suggestGasTipCap, v)
+		}
+	}
+	return oldestBlock, reward, baseFee, gasUsedRatio, nil
 }
 
 func (b *EthAPIBackend) ChainDb() ethdb.Database {


eth/api_backend.go
```diff
@@ -374,11 +374,25 @@ func (b *EthAPIBackend) SuggestGasTipCap(ctx context.Context) (*big.Int, error)
 		return nil, err
 	}
 	minGasTipCap := stateDb.GetState(systemcontracts.PolicyProxyHash, systemcontracts.GetMinGasTipCapStateHash()).Big()
-	return cmath.BigMax(suggestTipCap, minGasTipCap), nil
+	denom := new(big.Int).SetUint64(b.eth.blockchain.Config().BaseFeeChangeDenominator())
+	return cmath.BigMax(suggestTipCap, minGasTipCap.Add(minGasTipCap, new(big.Int).Div(minGasTipCap, denom))), nil
```
AnnaShaleva, line 378: > This fallback is removed in [MetaMask/core](https://github.com/MetaMask/core), but not yet the extension.

Do we have any chance to get back to minimum `effectiveGasTip` in future? When MetaMask is updated.

eth/api_backend.go
```diff
@@ -374,11 +374,25 @@ func (b *EthAPIBackend) SuggestGasTipCap(ctx context.Context) (*big.Int, error)
...
+	}
+	suggestGasTipCap, err := b.SuggestGasTipCap(ctx)
+	if err != nil {
+		return nil, nil, nil, nil, err
+	}
+	for i, r := range reward {
+		for j, v := range r {
+			reward[i][j] = cmath.BigMax(suggestGasTipCap, v)
+		}
+	}
```
AnnaShaleva, line 394: It's not quite correct because `b.SuggestGasTipCap` is based on the *latest* state. It won't work properly if I ask fee history for the latest 50 blocks and minGasTipCap was increased in some 25-th block before the current height. Then the whole reward history will include this increased value, which is not correct.

AnnaShaleva: At the same time I doubt we have other options. @txhsl, @chenquanyu, @roman-khimov any thoughts on this?

txhsl: I wish we have other options as well.

txhsl: MetaMask may have solved this problem in https://github.com/MetaMask/metamask-extension/pull/24520. Give me some time to check it.

eth/api_backend.go
txhsl: > Do we have any chance to get back to minimum `effectiveGasTip` in future? When MetaMask is updated.

ASA the new core codes get used in MetaMask's products.

eth/api_backend.go
txhsl: @AnnaShaleva I’ve tried the latest MetaMask explorer extension, and I think the gas tip estimation problem is solved.

I'll close this PR and related issue. If we have any great ideas to prevent similar issues, let's come back when it happens again.

AnnaShaleva: >  the gas tip estimation problem is solved.

It's really good!

PR #194, opened 2024-05-11
txpool: add regular txpool reannouncement for pending transactions

### LINKED ISSUE #193: Solve the broadcast problem of pending transactions under traffic and unstable connections
# Rationale

There is a widely-addressed issue of Geth, please refer to https://github.com/ethereum/go-ethereum/issues/22308. The problem is, some transactions may keep pending but never get mined when there is a traffic or the network connection is not stable.

Because the node will only broadcast a transaction once after the memory pool accepts it, so it is possible that the node mark a transaction as “broadcasted” but CNs don't receive it.

We don't allow local transactions on RPC nodes (now on T3), so there is no way to rebroadcast pending transactions (e.g. reboot).

This issue can be eased by increasing the amount of node connections, but it still exists.

# Implementation

Geth has never fixed this, but BSC tried with https://github.com/bnb-chain/bsc/pull/570, by allowing to set up a regular rebroadcast for local pending transactions. We may need a similar or the same thing.

txhsl:
Close #193. Ready for review.

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -346,6 +346,18 @@ var (
 		Value:    ethconfig.Defaults.TxPool.Lifetime,
 		Category: flags.TxPoolCategory,
 	}
+	TxPoolReannounceTimeThresholdFlag = &cli.DurationFlag{
+		Name:     "txpool.reannouncetimethreshold",
+		Usage:    "Threshold for announcing pending transactions again (default = 140 seconds, minimum = 1 minute)",
+		Value:    ethconfig.Defaults.TxPool.ReannounceTimeThreshold,
+		Category: flags.TxPoolCategory,
+	}
+	TxPoolReannounceRemotesFlag = &cli.BoolFlag{
+		Name:     "txpool.reannounceremotes",
+		Usage:    "Whether to reannounce remote transactions or not(default = true)",
+		Value:    ethconfig.Defaults.TxPool.ReannounceRemotes,
+		Category: flags.TxPoolCategory,
+	}
 	// Blob transaction pool settings
 	BlobPoolDataDirFlag = &cli.StringFlag{
 		Name:     "blobpool.datadir",
@@ -1492,6 +1504,12 @@ func setTxPool(ctx *cli.Context, cfg *legacypool.Config) {
 	}
 	if ctx.IsSet(TxPoolLifetimeFlag.Name) {
 		cfg.Lifetime = ctx.Duration(TxPoolLifetimeFlag.Name)
+	}
+	if ctx.IsSet(TxPoolReannounceTimeThresholdFlag.Name) {
+		cfg.ReannounceTimeThreshold = ctx.Duration(TxPoolReannounceTimeThresholdFlag.Name)
+	}
+	if ctx.IsSet(TxPoolReannounceRemotesFlag.Name) {
+		cfg.ReannounceRemotes = ctx.Bool(TxPoolReannounceRemotesFlag.Name)
 	}
 }
 
--- a/core/txpool/legacypool/legacypool.go
+++ b/core/txpool/legacypool/legacypool.go
@@ -51,6 +51,9 @@ const (
 	// more expensive to propagate; larger transactions also take more resources
 	// to validate whether they fit into the pool or not.
 	txMaxSize = 4 * txSlotSize // 128KB
+
+	// txReannounceMaxNum is the maximum number of transactions a reannounce action can include.
+	txReannounceMaxNum = 1024
 )
 
 var (
@@ -60,8 +63,9 @@ var (
 )
 
 var (
-	evictionInterval    = time.Minute     // Time interval to check for evictable transactions
-	statsReportInterval = 8 * time.Second // Time interval to report transaction pool stats
+	evictionInterval    = time.Minute              // Time interval to check for evictable transactions
+	statsReportInterval = 8 * time.Second          // Time interval to report transaction pool stats
+	reannounceInterval  = 7 * 2 * 10 * time.Second // Time interval to check for reannounce transactions
 )
 
 var (
@@ -134,6 +138,9 @@ type Config struct {
 	GlobalQueue  uint64 // Maximum number of non-executable transaction slots for all accounts
 
 	Lifetime time.Duration // Maximum amount of time non-executable transaction are queued
+
+	ReannounceTimeThreshold time.Duration // Threshold for announcing pending transactions again
+	ReannounceRemotes       bool          // Whether reannounce remote transactions or not
 }
 
 // DefaultConfig contains the default configurations for the transaction pool.
@@ -150,6 +157,9 @@ var DefaultConfig = Config{
 	GlobalQueue:  1024,
 
 	Lifetime: 3 * time.Hour,
+
+	ReannounceTimeThreshold: 7 * 2 * 10 * time.Second,
+	ReannounceRemotes:       true,
 }
 
 // sanitize checks the provided user configurations and changes anything that's
@@ -188,6 +198,10 @@ func (config *Config) sanitize() Config {
 		log.Warn("Sanitizing invalid txpool lifetime", "provided", conf.Lifetime, "updated", DefaultConfig.Lifetime)
 		conf.Lifetime = DefaultConfig.Lifetime
 	}
+	if conf.ReannounceTimeThreshold < time.Minute {
+		log.Warn("Sanitizing invalid txpool reannounce time threshold", "provided", conf.ReannounceTimeThreshold, "updated", DefaultConfig.ReannounceTimeThreshold)
+		conf.ReannounceTimeThreshold = DefaultConfig.ReannounceTimeThreshold
+	}
 	return conf
 }
 
@@ -199,13 +213,14 @@ func (config *Config) sanitize() Config {
 // current state) and future transactions. Transactions move between those
 // two states over time as they are received and processed.
 type LegacyPool struct {
-	config      Config
-	chainconfig *params.ChainConfig
-	chain       BlockChain
-	gasTip      atomic.Pointer[big.Int]
-	txFeed      event.Feed
-	signer      types.Signer
-	mu          sync.RWMutex
+	config       Config
+	chainconfig  *params.ChainConfig
+	chain        BlockChain
+	gasTip       atomic.Pointer[big.Int]
+	txFeed       event.Feed
+	reannoTxFeed event.Feed
+	signer       types.Signer
+	mu           sync.RWMutex
 
 	currentHead   atomic.Pointer[types.Header] // Current head of the blockchain
 	currentState  *state.StateDB               // Current state in the blockchain head
@@ -337,13 +352,15 @@ func (pool *LegacyPool) loop() {
 		prevPending, prevQueued, prevStales int
 
 		// Start the stats reporting and transaction eviction tickers
-		report  = time.NewTicker(statsReportInterval)
-		evict   = time.NewTicker(evictionInterval)
-		journal = time.NewTicker(pool.config.Rejournal)
+		report     = time.NewTicker(statsReportInterval)
+		evict      = time.NewTicker(evictionInterval)
+		journal    = time.NewTicker(pool.config.Rejournal)
+		reannounce = time.NewTicker(reannounceInterval)
 	)
 	defer report.Stop()
 	defer evict.Stop()
 	defer journal.Stop()
+	defer reannounce.Stop()
 
 	// Notify tests that the init phase is done
 	close(pool.initDoneCh)
@@ -393,6 +410,35 @@ func (pool *LegacyPool) loop() {
 				}
 				pool.mu.Unlock()
 			}
+
+		// Handle pending transaction reannouncement
+		case <-reannounce.C:
+			pool.mu.RLock()
+			reannoTxs := func() []*types.Transaction {
+				txs := make([]*types.Transaction, 0)
+				for addr, list := range pool.pending {
+					// Default ReannounceRemotes is true
+					if !pool.config.ReannounceRemotes && !pool.locals.contains(addr) {
+						continue
+					}
+
+					for _, tx := range list.Flatten() {
+						// Default ReannounceTimeThreshold is 140 seconds
+						if time.Since(tx.Time()) < pool.config.ReannounceTimeThreshold {
+							break
+						}
+						txs = append(txs, tx)
+						if len(txs) >= txReannounceMaxNum {
+							return txs
+						}
+					}
+				}
+				return txs
+			}()
+			pool.mu.RUnlock()
+			if len(reannoTxs) > 0 {
+				pool.reannoTxFeed.Send(core.ReannoTxsEvent{Txs: reannoTxs})
+			}
 		}
 	}
 }
@@ -425,6 +471,12 @@ func (pool *LegacyPool) SubscribeTransactions(ch chan<- core.NewTxsEvent, reorgs
 	// is because the new txs are added to the queue, resurrected ones too and
 	// reorgs run lazily, so separating the two would need a marker.
 	return pool.txFeed.Subscribe(ch)
+}
+
+// SubscribeReannoTransactions registers a subscription for reannounce transaction events,
+// supporting feeding only pending transactions.
+func (pool *LegacyPool) SubscribeReannoTransactions(ch chan<- core.ReannoTxsEvent) event.Subscription {
+	return pool.reannoTxFeed.Subscribe(ch)
 }
 
 // SetGasTip updates the minimum gas tip required by the transaction pool for a
--- a/eth/handler.go
+++ b/eth/handler.go
@@ -83,6 +83,9 @@ type txPool interface {
 	// can decide whether to receive notifications only for newly seen transactions
 	// or also for reorged out ones.
 	SubscribeTransactions(ch chan<- core.NewTxsEvent, reorgs bool) event.Subscription
+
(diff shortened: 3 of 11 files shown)


core/txpool/legacypool/legacypool.go
```diff
@@ -51,6 +52,12 @@ const (
 	// more expensive to propagate; larger transactions also take more resources
 	// to validate whether they fit into the pool or not.
 	txMaxSize = 4 * txSlotSize // 128KB
+
+	// txReannoMaxNum is the maximum number of transactions a reannounce action can include.
+	txReannoMaxNum = 1024
+
+	// reannoThreshold is a fixed number determines the gap of pending transaction reannouncement
+	reannoThreshold = 5
```
roman-khimov: We use a dynamic one in NeoGo, probably you've seen it. It depends on number of CNs and the presumption is that we expect consensus to handle it within two full rounds over CN list (different primaries can try including this transaction). But it's all heuristic and likely this value can be OK too.

txhsl:
May have compatibility problems, checking…
No, just something wrong with my local environment. :)

core/txpool/legacypool/legacypool.go
txhsl: Yes, I see, now trying to implement it if possible.

txhsl: Done.

txhsl:
@roman-khimov Finally the BSC solution, it works well in our tests. I think it is better than neo-go's strategy in this case, since a continuous `nonce` is required in Geth.

roman-khimov:
> it is better than neo-go's strategy

No. Can't be!

txhsl:
> No. Can't be!

Well, there is a coding mistake that deactivated the reannouncement operation. :(

https://github.com/bane-labs/go-ethereum/pull/194/commits/560ba4963b9900cb38beb63a46816a20d3d2eff7#diff-3b03ebf997a1699ce8e84a53a4b3981e403e74e2f90128eb188ad4226703970cL1449.

But can I merge the new codes? It works well. Reannouncing a series of continuous pending transactions seems to be a better choice in Geth.

roman-khimov:
@txhsl, just joking. @AnnaShaleva will take a look at it, don't worry. Maybe it really fits E-chain better.

cmd/utils/flags.go
```diff
@@ -346,6 +346,18 @@ var (
...
 	}
+	TxPoolReannounceTimeFlag = &cli.DurationFlag{
+		Name:     "txpool.reannouncetime",
+		Usage:    "Duration for announcing local pending transactions again (default = 10 years, minimum = 1 minute)",
+		Value:    ethconfig.Defaults.TxPool.ReannounceTime,
+		Category: flags.TxPoolCategory,
+	}
+	TxPoolReannounceRemotesFlag = &cli.BoolFlag{
+		Name:     "txpool.reannounceremotes",
+		Usage:    "Wether reannnounce remote transactions or not(default = false)",
```
roman-khimov: Whether to

cmd/utils/flags.go
```diff
@@ -346,6 +346,18 @@ var (
 		Value:    ethconfig.Defaults.TxPool.Lifetime,
 		Category: flags.TxPoolCategory,
 	}
+	TxPoolReannounceTimeFlag = &cli.DurationFlag{
```
roman-khimov: Ah, it's time-based, which is conceptually wrong to me. Time doesn't matter much, blocks do, since only new blocks can flush old transactions from the mempool.

roman-khimov (commented):
Well, now I know what BSC does and I also know NeoGo is still better which is quite satisfying. At the same time you may try it this way as well, at least we know that it's tested in some network for quite some time already.

cmd/utils/flags.go
txhsl: Yes, it's time-based, and the Geth transaction eviction is time-based as well.
https://github.com/bane-labs/go-ethereum/blob/d9c98489b2ca73ffed58a9ef3d3c6ce9d2b031cc/core/txpool/legacypool/legacypool.go#L353-L357
I prefer to put these similar things together. :)

roman-khimov: Makes sense this way.

cmd/utils/flags.go
```diff
@@ -346,6 +346,18 @@ var (
...
+		Name:     "txpool.reannouncetime",
+		Usage:    "Duration for announcing local pending transactions again (default = 10 years, minimum = 1 minute)",
+		Value:    ethconfig.Defaults.TxPool.ReannounceTime,
+		Category: flags.TxPoolCategory,
+	}
+	TxPoolReannounceRemotesFlag = &cli.BoolFlag{
+		Name:     "txpool.reannounceremotes",
+		Usage:    "Whether to reannnounce remote transactions or not(default = false)",
+		Value:    ethconfig.Defaults.TxPool.ReannounceRemotes,
```
AnnaShaleva: s/reannnounce/reannounce

Sorry for the nitpicking, but it's a user-facing thing.

core/txpool/legacypool/legacypool.go
```diff
@@ -188,6 +197,10 @@ func (config *Config) sanitize() Config {
 		log.Warn("Sanitizing invalid txpool lifetime", "provided", conf.Lifetime, "updated", DefaultConfig.Lifetime)
 		conf.Lifetime = DefaultConfig.Lifetime
 	}
+	if conf.ReannounceTime < time.Minute {
+		log.Warn("Sanitizing invalid txpool reannounce time", "provided", conf.ReannounceTime, "updated", time.Minute)
+		conf.ReannounceTime = time.Minute
```
AnnaShaleva:
```suggestion
		conf.ReannounceTime = DefaultConfig.ReannounceTime
```

core/txpool/legacypool/legacypool.go
```diff
@@ -188,6 +197,10 @@ func (config *Config) sanitize() Config {
 		log.Warn("Sanitizing invalid txpool lifetime", "provided", conf.Lifetime, "updated", DefaultConfig.Lifetime)
 		conf.Lifetime = DefaultConfig.Lifetime
 	}
+	if conf.ReannounceTime < time.Minute {
+		log.Warn("Sanitizing invalid txpool reannounce time", "provided", conf.ReannounceTime, "updated", time.Minute)
```
AnnaShaleva:
```suggestion
		log.Warn("Sanitizing invalid txpool reannounce time", "provided", conf.ReannounceTime, "updated", DefaultConfig.ReannounceTime)
```

cmd/utils/flags.go
```diff
@@ -346,6 +346,18 @@ var (
 		Value:    ethconfig.Defaults.TxPool.Lifetime,
 		Category: flags.TxPoolCategory,
 	}
+	TxPoolReannounceTimeFlag = &cli.DurationFlag{
+		Name:     "txpool.reannouncetime",
```
AnnaShaleva: How about renaming it to `txpool.reannounceinterval`? It looks to me more like an interval than a time.

AnnaShaleva: Also rename all related variable names.

cmd/utils/flags.go
```diff
@@ -346,6 +346,18 @@ var (
...
+		Usage:    "Duration for announcing local pending transactions again (default = 10 years, minimum = 1 minute)",
+		Value:    ethconfig.Defaults.TxPool.ReannounceTime,
+		Category: flags.TxPoolCategory,
+	}
+	TxPoolReannounceRemotesFlag = &cli.BoolFlag{
+		Name:     "txpool.reannounceremotes",
+		Usage:    "Whether to reannnounce remote transactions or not(default = false)",
```
AnnaShaleva: NeoGo's default behaviour is to always reannounce. And in general, I don't think reannounement is bad since reannouncement makes the mempools more synchronized (and we don't have a lot of nodes in the network). Thus, can we revert the flag meaning to something like `txpool.noremotesreannounce` with default value set to `false`?

@roman-khimov, what do you think?

core/txpool/legacypool/legacypool.go
```diff
@@ -51,6 +51,9 @@ const (
 	// more expensive to propagate; larger transactions also take more resources
 	// to validate whether they fit into the pool or not.
 	txMaxSize = 4 * txSlotSize // 128KB
+
+	// txReannoMaxNum is the maximum number of transactions a reannounce action can include.
+	txReannoMaxNum = 1024
```
AnnaShaleva: txReannounceMaxNum? It's long, but clear.

cmd/utils/flags.go
```diff
@@ -346,6 +346,18 @@ var (
 		Value:    ethconfig.Defaults.TxPool.Lifetime,
 		Category: flags.TxPoolCategory,
 	}
+	TxPoolReannounceTimeFlag = &cli.DurationFlag{
+		Name:     "txpool.reannouncetime",
+		Usage:    "Duration for announcing local pending transactions again (default = 10 years, minimum = 1 minute)",
+		Value:    ethconfig.Defaults.TxPool.ReannounceTime,
```
AnnaShaleva: An if we revert `txpool.reannounceremotes` value, than we can set the default to 7*2*10s=140s (should be sufficient for 7-CNs network with 10-seconds blocks).

core/txpool/legacypool/legacypool.go
```diff
@@ -393,6 +409,34 @@ func (pool *LegacyPool) loop() {
...
+							return txs
+						}
+					}
+				}
+				return txs
+			}()
+			pool.mu.RUnlock()
+			if len(reannoTxs) > 0 {
+				pool.reannoTxFeed.Send(core.ReannoTxsEvent{Txs: reannoTxs})
+			}
```
AnnaShaleva, line 441: A conceptual question: we reannounce only the first batch, but how about the others? We may send several batches to reannounce.

core/txpool/legacypool/legacypool.go
```diff
@@ -668,6 +718,7 @@ func (pool *LegacyPool) add(tx *types.Transaction, local bool) (replaced bool, e
 		invalidTxMeter.Mark(1)
 		return false, err
 	}
+
```
AnnaShaleva: Useless change.

eth/handler.go
```diff
@@ -666,6 +678,24 @@ func (h *handler) BroadcastTransactions(txs types.Transactions) {
...
+// ReannounceTransactions will announce a batch of local pending transactions
+// to a square root of all peers.
+func (h *handler) ReannounceTransactions(txs types.Transactions) {
+	hashes := make([]common.Hash, 0, txs.Len())
+	for _, tx := range txs {
+		hashes = append(hashes, tx.Hash())
+	}
+
+	// Announce transactions hash to a batch of peers
+	peersCount := uint(math.Sqrt(float64(h.peers.len())))
```
AnnaShaleva, line 690: I can't stop myself from pointing that out: this number is important. Take a look at https://github.com/nspcc-dev/neo-go/issues/608 and https://github.com/nspcc-dev/neo-go/pull/2743.

eth/peerset.go
```diff
@@ -234,6 +234,25 @@ func (ps *peerSet) allPeers() []*ethPeer {
 	return list
 }
 
+// headPeers retrieves a specified number list of peers.
+func (ps *peerSet) headPeers(num uint) []*ethPeer {
```
AnnaShaleva, line 238: How about something like `somePeers`?

eth/peerset.go
```diff
@@ -234,6 +234,25 @@ func (ps *peerSet) allPeers() []*ethPeer {
 	return list
 }
 
+// headPeers retrieves a specified number list of peers.
```
AnnaShaleva, line 237:
```suggestion
// somePeers returns not more than the specified number of peers as a list. The order of peers is undefined and may vary from call to call.
```

core/txpool/legacypool/legacypool.go
txhsl: I see.

cmd/utils/flags.go
txhsl: In fact, it is not the interval, I would like to call it a threshold.

txhsl: The naming is from BSC. The interval is fixed 1 minute here.

https://github.com/bane-labs/go-ethereum/blob/1d376fe89397cc580b4bd4aa235f8c80334b8e1e/core/txpool/legacypool/legacypool.go#L68

cmd/utils/flags.go
txhsl: `TxPoolReannounceRemotes` is a little bit different from what we are talking about here.

When we are testing and deploying, we disabled `nolocals` on RPC nodes. So, outside transactions sent through RPC are marked as `local` when being added to `txpool`.
https://github.com/bane-labs/go-ethereum/blob/1d376fe89397cc580b4bd4aa235f8c80334b8e1e/eth/api_backend.go#L293

But the other transactions, which are broadcasted by other nodes and downloaded by `txFetcher` are marked as `remote`.
https://github.com/bane-labs/go-ethereum/blob/1d376fe89397cc580b4bd4aa235f8c80334b8e1e/eth/handler.go#L304

So, the `TxPoolReannounceRemotes` setting actually determines “whether to reannounce transactions passed by”.  Well, it seems the same. :(

In our local tests, I set it `true` on CNs and set it `false` on RPC nodes. What do you think?

core/txpool/legacypool/legacypool.go
txhsl: Well, fine.

core/txpool/legacypool/legacypool.go
txhsl: The pending list guarantees at least 16 transactions per address. So at most transactions from 64 different senders will get reannounced. As for the others, yes, they have to wait until some of the reannounced transactions to get cleaned out.

But the reannounced pending transactions are verified and can be executed immediately, so it is just a consensus pipeline.

cmd/utils/flags.go
txhsl: Yes, but most of the transactions never get stuck in the pending list, it only happens under traffics and unstable connections. Most of them are recognized by CNs and just waiting to be packaged, so the amount of transactions we have to reannounce is not huge.

E.g. I set the `ReannounceTime` (or I mean the threshold) to 10 minutes on CNs, then the loop only reannounces transactions keep pending for more than 10 minutes. But I set it to 1 minutes on RPC nodes, then it has different usage. We don't have to do that very frequently.

eth/handler.go
txhsl: Yes, it is. Sqrt broadcasting is widely used in Geth. How do you think?
https://github.com/bane-labs/go-ethereum/blob/9062f7d6f5013953cb8039f06d8cd96219c54d2a/eth/handler.go#L612
https://github.com/bane-labs/go-ethereum/blob/9062f7d6f5013953cb8039f06d8cd96219c54d2a/eth/handler.go#L656

eth/peerset.go
txhsl: Strange.

eth/handler.go
AnnaShaleva: > Sqrt broadcasting is widely used in Geth.

Then let's stay aligned with the rest of the Geth code.

core/txpool/legacypool/legacypool.go
AnnaShaleva: Convincing, let's keep it this way.

cmd/utils/flags.go
AnnaShaleva: > I set it true on CNs

Agree with this.

> set it false on RPC nodes.

But not with this. Because, as I've said, there's nothing bad in the reannouncement, it allows to keep the network more synced. And I think we can allow to reannounce remote transactions because with dBFT we have unified rules of consensus (unlike with the PoW race). With dBFT there's actually almost no difference in processing between the "local" and "remote" transactions. So I think that all of them should be reannounced.

eth/peerset.go
AnnaShaleva: The point is that head sounds like the top of the list, which may be misleading since the returned value is not sorted in any way and may vary from call to call. But I'm OK with this name if we adjust the comment.

cmd/utils/flags.go
roman-khimov: In fact, RPC nodes is exactly where transactions are gonna be stuck in case of network problems. CNs will flush theirs into blocks eventually, but RPC nodes can only reannounce.

cmd/utils/flags.go
txhsl: Updated.

cmd/utils/flags.go
txhsl: Updated.

core/txpool/legacypool/legacypool.go
```diff
@@ -393,6 +410,34 @@ func (pool *LegacyPool) loop() {
...
+			pool.mu.RLock()
+			reannoTxs := func() []*types.Transaction {
+				txs := make([]*types.Transaction, 0)
+				for addr, list := range pool.pending {
+					if !pool.config.ReannounceRemotes && !pool.locals.contains(addr) {
+						continue
+					}
+
+					for _, tx := range list.Flatten() {
+						// Default ReannounceTimeThreshold is 10 years, won't announce by default.
```
AnnaShaleva: Please, adjust the comment wrt the updated code.

AnnaShaleva (commented):
Please, group the commits into logical blocks. Otherwise LGTM, tests pass.

core/txpool/legacypool/legacypool.go
txhsl: OK, I'll check it.

PR #216, opened 2024-05-29
systemcontract: add and apply candidate limit policy

### LINKED ISSUE #197: GVN-01 - High Resource Consumption in candidateList `[High]`
<p>A critical vulnerability related to excessive resource consumption has been identified in the <code>onPersist </code>method of the <strong>NeoX</strong> <code>Governance</code> contract, which uses the <code>EnumerableSet</code> library. Specifically, the <code>values()</code> method of this library is called multiple times: directly on line 313 and via the <code>_computeConsensus()</code> function on line 325. This method is known for its high gas demand, as it copies all values of the <code>candidateList</code> set in memory, which is described in the OpenZeppelin documentation as potentially expensive.</p>

image

<p>The main concern is that repeatedly using <code>values()</code> in a method that alters the state of the contract increases the gas cost significantly, especially as the <code>candidateList</code> set grows. This increase in gas consumption not only increases operating costs, but also represents a serious security vulnerability.</p>

image

<p>Since the registration of new candidates in <code>candidateList</code> is public, an attacker could exploit this feature to register a large number of candidates, causing the execution of <code>onPersist</code> to consume more resources than can be provided in a block, effectively blocking the function and causing a denial of service (DoS).</p>

**Remediation:**
<ul><li><p>Limit the use of <code>values()</code> in critical functions: Redesign the <code>onPersist</code> method to eliminate repeated use of the <code>values()</code> method within operations that change the state of the contract. Consider using iterators or alternative mechanisms that do not require copying the entire set into memory.</p></li><li><p>Perform load testing and contract optimization to ensure that key functions can handle a high number of interactions without exceeding resource limits. These tests will help identify and mitigate potential failure points due to excessive gas consumption.</p></li></ul>

**References:**
- https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/structs/EnumerableSet.sol#L211-L218

**Source references:**
<p><a target="_blank" rel="noopener noreferrer nofollow" href="https://github.com/bane-labs/go-ethereum/blob/ca856ce0ad5d20490d8e7304430f4db3e6331340/contracts/solidity/Governance.sol#L313">https://github.com/bane-labs/go-ethereum/blob/ca856ce0ad5d20490d8e7304430f4db3e6331340/contracts/solidity/Governance.sol#L313</a></p>
<p><a target="_blank" rel="noopener noreferrer nofollow" href="https://github.com/bane-labs/go-ethereum/blob/ca856ce0ad5d20490d8e7304430f4db3e6331340/contracts/solidity/Governance.sol#L325">https://github.com/bane-labs/go-ethereum/blob/ca856ce0ad5d20490d8e7304430f4db3e6331340/contracts/solidity/Governance.sol#L325</a></p>

txhsl:
Close #197.


txhsl:
The commit history seems a mess. :(

But the code is now rebased to master branch and ready for review.

contracts/solidity/Policy.sol
```diff
@@ -12,6 +12,7 @@ contract Policy is IPolicy, GovernanceVote, GovProxyUpgradeable {
     mapping(address => bool) public isBlackListed;
     uint256 public minGasTipCap;
     uint256 public baseFee;
+    uint256 public candidateLimit;
```
AnnaShaleva: We need a default value for it. Otherwise if there's no default, we'll end up in an unpleasant situation in T3, no registration will be possible until the first execution of `setCandidateLimit`.

contracts/solidity/Policy.sol
```diff
@@ -96,4 +97,22 @@ contract Policy is IPolicy, GovernanceVote, GovProxyUpgradeable {
...
     }
+
+    // set candidate limit (increase only)
+    function setCandidateLimit(
+        uint256 _candidateLimit
+    )
+        external
+        needVote(
+            bytes32(
+                0x172d358b638a8ee3e962dd73800c4025c48eb0f79c479bc2cdd1f63e72779efc
```
AnnaShaleva, line 113: Deserves a comment.

contracts/solidity/Governance.sol
```diff
@@ -111,6 +115,8 @@ contract Governance is IGovernance, ReentrancyGuard, GovProxyUpgradeable {
         if (tx.origin != msg.sender) revert Errors.OnlyEOA();
         if (msg.value < registerFee) revert Errors.InsufficientValue();
         if (shareRate > 1000) revert Errors.InvalidShareRate();
+        if (candidateList.length() >= IPolicy(POLICY).candidateLimit())
+            revert Errors.CandidateExceedLimit();
```
AnnaShaleva: Errors.CandidateLimitExceeded?

AnnaShaleva (commented):
Quite a straightforward API.

AnnaShaleva:
> The commit history seems a mess. :(

Just squash all of them and a add a pretty commit message, it will solve the problem!

contracts/solidity/Policy.sol
txhsl: 🤔The first execution of `setCandidateLimit` is necessary. When we upgrade the policy contract, we are replacing the implementation codes, but the storage remains 0.

@chenquanyu Should we add an allowance for self-invocation in `needVote`? Otherwise, we can not execute `setCandidateLimit` when upgrading.

chenquanyu: A zero value for this field doesn't make sense, so I recommend making the field private and adding a getter method that returns a default value when it is zero.

txhsl: If we add default value for `candidateLimit`, then we add default values for all the other polices. The way we deal with them should be consistent.

chenquanyu: I think they are different, zero values ​​for `minGasTipCap` and `baseFee` might be useful in the future, and we read the storage location directly in `geth`. For `candidateLimit` and other fields that may be added in the future, setting default values ​​can avoid breaking contract usage if needed.

txhsl: Added.

chenquanyu (approved):

PR #227, opened 2024-06-13
systemcontract: fix the issue of clearing the vote after voting is passed

### LINKED ISSUE #228: The voting history is not completely cleared in GovernanceVote contract
Now when the GovernanceVote contract vote passes, it will only clear the voting records from the current consensus list. If there are voters from the previous epoch who are not in the current consensus list, the voting history will not be cleared and may be used maliciously in the future.

chenquanyu:
Close #228 
This PR will change the storage structure of `GovernanceVote`, but this is compatible with the testnet. Tested on remix, the mapping in the first slot of the struct will remain in the same slot as the mapping of mapping.

### DIFF
--- a/contracts/solidity/base/GovernanceVote.sol
+++ b/contracts/solidity/base/GovernanceVote.sol
@@ -6,52 +6,72 @@ import {IGovReward} from "../interfaces/IGovReward.sol";
 
 abstract contract GovernanceVote {
     // events for voting
-    event Vote(address indexed voter, bytes32 indexed methodKey, bytes32 paramKey);
+    event Vote(
+        address indexed voter,
+        bytes32 indexed methodKey,
+        bytes32 paramKey
+    );
     event VotePass(bytes32 indexed methodKey, bytes32 paramKey);
 
     // governance reward contact
-    address public constant govReward =
+    address public constant GOV_REWARD =
         0x1212000000000000000000000000000000000003;
 
-    // vote mapping, method key -> (user address -> param key)
-    mapping(bytes32 => mapping(address => bytes32)) private voteMap;
+    struct VoteInfo {
+        // vote mapping, user address -> param key
+        mapping(address voter => bytes32) values;
+        // voter, used for clearing vote after voting is passed
+        address[] voters;
+    }
+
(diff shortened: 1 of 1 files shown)


contracts/solidity/base/GovernanceVote.sol
```diff
@@ -6,52 +6,71 @@ import {IGovReward} from "../interfaces/IGovReward.sol";
...
-    // vote mapping, method key -> (user address -> param key)
-    mapping(bytes32 => mapping(address => bytes32)) private voteMap;
+    struct VoteInfo {
+        address[] voters;
+        // vote mapping, user address -> param key
+        mapping(address voter => bytes32) values;
+    }
+
+    // vote mapping, method key -> voteInfo
+    mapping(bytes32 => VoteInfo) private voteMap;
```
txhsl, line 28: Will there be any storage compatible problem with T3? 🤔

chenquanyu: Yes, it's incompatible with T3, please read the comment https://github.com/bane-labs/go-ethereum/pull/227#issue-2350598008.

txhsl: Yeh I just heard we may have a new testnet if necessary.

chenquanyu: Sorry, I just found changing the order can keep compatible with T3 storage. Now it should be compatible with T3.

contracts/solidity/base/GovernanceVote.sol
```diff
@@ -6,52 +6,72 @@ import {IGovReward} from "../interfaces/IGovReward.sol";
...
         }
-        if (validVotes < (length + 1) / 2) return;
-
         // clear vote
-        emit VotePass(methodKey, paramKey);
         _clearVote(methodKey);
-
         // execute method
         _;
+        emit VotePass(methodKey, paramKey);
```
txhsl: Maybe emit this before execution?

chenquanyu: Fixed.

txhsl (approved):

PR #230, opened 2024-06-14
Adjust BaseFee burning mechanism

AnnaShaleva:
Please note that the target branch is policy-basefee (#166).

I've reviewed all the modules that become dependent from the node state after #166. It turns out that these modules are created with the full backend and does not work with the light backend anyway, thus it's not *such* critical to explicitly use the node state there. The only concern is about Oracle service: it didn't have any state (or DB) dependency before these BaseFee changes and it has this dependency now. And this service has an ability to process *batch* of blocks, which makes it extremely resource-consuming to retrieve state for every potential block where BaseFee change might have happen.

@txhsl, @chenquanyu, @nicolegys ready for review. We'll need tests for t8n tool and, which is more important, for Oracle service.


eth/gasprice/feehistory.go
```diff
@@ -83,11 +82,12 @@ func (oracle *Oracle) processBlock(bf *blockFees, percentiles []float64) {
 		bf.results.baseFee = new(big.Int)
 	}
 	if chainconfig.IsLondon(big.NewInt(int64(bf.blockNumber + 1))) {
-		state, _, err := oracle.backend.StateAndHeaderByNumber(context.Background(), rpc.BlockNumber(bf.blockNumber))
-		if err != nil {
-			log.Error("Failed to get state", "err", err, "header number", bf.blockNumber)
-		}
-		bf.results.nextBaseFee = eip1559.CalcBaseFeeDBFT(chainconfig, bf.header, state)
+		oracle.cacheLock.RLock()
+		// TODO: the idea is to reuse the cached value for block processing, but the cache is updated
```
txhsl: I'm not sure if we should use fee history for block processing, since this was designed for fee market analysis, to provide some advice through RPC APIs.

eth/gasprice/gasprice.go
```diff
@@ -216,12 +221,28 @@ func (oracle *Oracle) SuggestTipCap(ctx context.Context) (*big.Int, error) {
...
+		}
+		baseFee = eip1559.CalcBaseFeeDBFT(cfg, head, state) // OK
+		if cfg.IsNeoXBurn(head.Number, head.Time) {
+			minGasTipCap = state.GetState(systemcontracts.PolicyProxyHash, systemcontracts.GetMinGasTipCapStateHash()).Big()
+		}
+	}
 	oracle.cacheLock.Lock()
 	oracle.lastHead = headHash
 	oracle.lastPrice = price
+	oracle.lastBaseFee = baseFee
```
chenquanyu: `lastBaseFee` will only update after `SuggestTipCap` is called, which may be too inaccurate.

AnnaShaleva: Yep, I had TODO about it, ref. https://github.com/bane-labs/go-ethereum/pull/230#discussion_r1642123722. Now it's fixed.

eth/gasprice/feehistory.go
AnnaShaleva: I'm also not sure about it, but my primary concern is that we need to be compatible with the existing dApps and they expect this value to be present.

AnnaShaleva: To be honest, I'm thinking about adding an extension: some `getBaseFee` RPC API that will be exposed from the node for users and that will return a *single* value (the current baseFee stored in Policy). What do you think?

AnnaShaleva:
Ready for review, the PR comment is updated, please, take a look.

eth/gasprice/feehistory.go
roman-khimov: RPC extension is questionable, no software is ready for it and the primary thing Bane provides is compatibility.

txhsl: > RPC extension is questionable, no software is ready for it and the primary thing Bane provides is compatibility.

Yes, most of them use `eth_feeHistory`, which contains a `baseFeePerGas` field.

chenquanyu (approved):

txhsl (dismissed):

AnnaShaleva:
@nicolegys, could you please test this branch? Especially the Oracle service with his fee history call.

chenquanyu:
> @nicolegys, could you please test this branch? Especially the Oracle service with his fee history call.

@songb2 This is our new tester.

AnnaShaleva:
I just realized that unit-tests are failing, mostly due to the suitable mocks absence. I'll fix the UTs, but the code itself is ready for testing, @songb2.

songb2:
I've tested the following apis by changing baseFee and minGasTipCap in policy and also made some transactions, the reponses changes as expected:
1. eth_maxPriorityFeePerGas
2. eth_gasPrice
3. eth_feeHistory

AnnaShaleva:
@txhsl, @chenquanyu, please check out 6 latest commits, I haven't touched the previously reviewed changes because they are the target of an audit. Tests are fixed, but the implementation is also adjusted, I found a couple of bugs.

@songb2, could you please check that the updated branch works OK with Metamask and it's possible for Metamask to generate and properly submit transaction to the network if BaseFee and minGasTipCap values are changed in Policy. Also, we need a test for t8ntool: create an Env configuration that includes NeoXBurn fork and some `currentBaseFee` value set in the Env config. Try to execute some transaction with this Env t8ntool configuration and check whether the BaseFee limit is properly checked against the pre-configured value.

Otherwise ready to be merged.

eth/gasprice/gasprice.go
```diff
@@ -216,12 +230,29 @@ func (oracle *Oracle) SuggestTipCap(ctx context.Context) (*big.Int, error) {
 	if price.Cmp(oracle.maxPrice) > 0 {
 		price = new(big.Int).Set(oracle.maxPrice)
 	}
+	if cfg := oracle.backend.ChainConfig(); cfg.IsLondon(head.Number) {
+		state, _, err := oracle.backend.StateAndHeaderByNumber(ctx, rpc.BlockNumber(head.Number.Uint64()))
+		if err != nil {
+			return nil, nil, nil, fmt.Errorf("failed to get state at %d to calculate base fee: %w", head.Number.Uint64(), err)
+		}
+		lastBaseFee = eip1559.CalcBaseFeeDBFT(cfg, head, state)
+		if cfg.IsNeoXBurn(head.Number, head.Time) {
```
chenquanyu: `migGasTip` was added before the `NeoXBurn` fork, and `basefee` should be calculated here after `NeoXBurn` fork.

AnnaShaleva: Nice catch, will fix it.

AnnaShaleva: Fixed in a separate commit because the original commit is the target of audit.

chenquanyu (dismissed):

songb2:
@AnnaShaleva @chenquanyu @txhsl 
Still in testing, but I tested metamask and also some eth apis and found the following problems, please help confirm if they are issues:
1. go to Edit gas fee page, Low, Market, and Aggressive options don't show numbers
image
2. By default basefee and minGasTipCap are all 1 gwei, so the eth_maxPriorityFeePerGas is also 1 gwei, and eth_gasPrice is 2 gwei.
When I change minGasTipCap to 0.5 gwei, but eth_maxPriorityFeePerGas is still 1 gwei (I think it should change to 0.5 gwei), and eth_gasPrice is still 2 gwei (I think it should be 1.5 gwei)

| condition            | baseFee | minGasTipGap | eth_maxPriorityFeePerGas | eth_gasPrice |
|--------------------------|--------|--------|--------|--------|
| default value in policy | 1 | 1 | 1 | 2 |
| decrease minGasTipGap to 0.5 | 1 | 0.5 | 1 | 2 |

And the issue can also refected on metamask ui, the Priority Fee should be 0.5 gwei, Max base fee should be 1.5 gwei.
image

txhsl:
> 1. go to Edit gas fee page, Low, Market, and Aggressive options don't show numbers

I believe this is because our network ID is not supported by their gas API. [Metamask](https://github.com/MetaMask/core/blob/036edd310aba1001158aacd55f2e3cfcb6cebb59/packages/gas-fee-controller/src/gas-util.ts#L93) doesn't fetch these data from RPC, it sends request to https://gas.api.cx.metamask.io/networks/{networkid}/suggestedGasFees.

AnnaShaleva:
> I believe this is because our network ID is not supported by their gas API.

I initially was sure that it was something wrong on our side, but if it depends on network ID, then there's nothing we can do about it.

> When I change minGasTipCap to 0.5 gwei, but eth_maxPriorityFeePerGas is still 1 gwei (I think it should change to 0.5 gwei), and eth_gasPrice is still 2 gwei (I think it should be 1.5 gwei)

About this behaviour: to calculate `eth_maxPriorityFeePerGas` we use maximum from `minGasTipCap` and GAS price:
https://github.com/bane-labs/go-ethereum/blob/47d271238a77ddc36865ca697f3e2aff8e9d87e1/eth/api_backend.go#L366-L372

So I believe that if GAS price is still 1 gwei, then `eth_maxPriorityFeePerGas` returns the proper value.

And regarding `eth_gasPrice`: to calculate it we use suggested TipCap added to the latest block's BaseFee:
https://github.com/bane-labs/go-ethereum/blob/47d271238a77ddc36865ca697f3e2aff8e9d87e1/internal/ethapi/api.go#L77-L85
where suggested TipCap is a maximum from `minGasTipCap` and GAS price. I believe block's BaseFee is 1, and GAS price is still 1, so the described behaviour works according to the code. The question is: should we calculate suggested TipCap as a maximum from `minGasTipCap` and GAS price? @chenquanyu, @txhsl, what do you think?

songb2:
@AnnaShaleva @chenquanyu @txhsl 
For the tn8tool, I try to set neoXBurnTime to cmd/evm/testdata/29/env.json, but it doesn't take effect, seems that it wasn't read by config.go. I print some logs in IsNeoXBurn, it shows that the field is nil, as as result, it return false,. 
func isTimestampForked(s *uint64, head uint64) bool {
    if s == nil {
        return false
    }
    return *s <= head
}
I use the command to test ../../build/bin/evm t8n --input.alloc=./testdata/29/alloc.json --input.txs=./testdata/29/txs.json --input.env=./testdata/29/env.json --state.fork=Shanghai
image

AnnaShaleva:
> but it doesn't take effect, seems that it wasn't read by config.go

Will take a look at it, probably more modifications are needed to make t8n tool work with NeoXBurn.

AnnaShaleva:
> but it doesn't take effect, seems that it wasn't read by config.go

@songb2, turns out that fork should be specified via `state.fork` CLI flag, not via t8ntool env config. Please, check the last commit, I've extended the list of valid hardforks recognizable by t8ntool, so you have to specify `--state.fork=NeoXBurn`, it works like Shanghai+burn-related changes. Please, check this issue one more time and see how it works:
https://github.com/bane-labs/go-ethereum/blob/ba2663cb363aa543daca31ba7c216197f5f77c78/tests/state_test_util.go#L141

chenquanyu:
>  The question is: should we calculate suggested TipCap as a maximum from `minGasTipCap` and GAS price? @chenquanyu, @txhsl, what do you think?

It looks like `GAS price` comes from the 60% quantile of the historical block gasTip list. If we use the maximum value of `minGasTipCap` and `GAS price`, then if we reduce `minGasTipCap` in Policy, suggestTipCap will always remain the same. So I suggest we just use `minGasTipCap` as suggestTipCap.

AnnaShaleva:
> So I suggest we just use minGasTipCap as suggestTipCap.

@chenquanyu, @txhsl, implemented, check out the last commit please.

@songb2 we need to test this version with Metamask.

chenquanyu (approved):

AnnaShaleva:
I'm merging this PR into policy-basefee so that it's possible to update the base branch. However, final test with Metamask is still needed, we're not going to release without it. Thus, @songb2, please, write the test results to #166. Any further discussions are moved to the original PR.

PR #235, opened 2024-06-28
antimev: allow skip nonce by 1 for encryption external tx

chenquanyu:
For antimev, the encryption external transaction will be a normal transfer to GovReward contract with encrypted internal transaction in data. We need to allow encryption external transaction can skip nonce by 1 to execute encrypted internal transaction first to support browser wallet plugin like metamask.

Close this PR as we decided to use the same nonce for both the envelope and the encrypted transaction, and only one of them will be on-chain.

### DIFF
--- a/core/encryption/encryption.go
+++ b/core/encryption/encryption.go
@@ -2,15 +2,14 @@ package encryption
 
 import (
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/systemcontracts"
 	"github.com/ethereum/go-ethereum/core/types"
 )
 
-const ENC_RECEIVER = "0x1212000000000000000000000000000000000003"
-
 func IsEncReceiver(to *common.Address) bool {
-	return to != nil && *(to) == common.HexToAddress(ENC_RECEIVER)
+	return to != nil && *(to) == systemcontracts.GovernanceRewardHash
 }
 
 func IsEncTx(tx *types.Transaction) bool {
-	return tx != nil && tx.To() != nil && *(tx.To()) == common.HexToAddress(ENC_RECEIVER)
+	return tx != nil && IsEncReceiver(tx.To())
 }
--- a/core/encryption/encryption.go
+++ b/core/encryption/encryption.go
@@ -1,0 +1,16 @@
+package encryption
+
+import (
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+)
+
+const ENC_RECEIVER = "0x1212000000000000000000000000000000000003"
+
+func IsEncReceiver(to *common.Address) bool {
+	return to != nil && *(to) == common.HexToAddress(ENC_RECEIVER)
+}
+
+func IsEncTx(tx *types.Transaction) bool {
+	return tx != nil && tx.To() != nil && *(tx.To()) == common.HexToAddress(ENC_RECEIVER)
+}
--- a/core/state_transition.go
+++ b/core/state_transition.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/ethereum/go-ethereum/common"
 	cmath "github.com/ethereum/go-ethereum/common/math"
+	"github.com/ethereum/go-ethereum/core/encryption"
 	"github.com/ethereum/go-ethereum/core/systemcontracts"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/core/vm"
@@ -278,7 +279,8 @@ func (st *StateTransition) preCheck() error {
 	if !msg.SkipAccountChecks {
 		// Make sure this transaction's nonce is correct.
 		stNonce := st.state.GetNonce(msg.From)
-		if msgNonce := msg.Nonce; stNonce < msgNonce {
+		// add encryption external tx check
+		if msgNonce := msg.Nonce; stNonce < msgNonce && !(encryption.IsEncReceiver(msg.To) && stNonce+1 == msgNonce) {
 			return fmt.Errorf("%w: address %v, tx: %d state: %d", ErrNonceTooHigh,
 				msg.From.Hex(), msgNonce, stNonce)
 		} else if stNonce > msgNonce {
@@ -445,7 +447,7 @@ func (st *StateTransition) TransitionDb() (*ExecutionResult, error) {
 		ret, _, st.gasRemaining, vmerr = st.evm.Create(sender, msg.Data, st.gasRemaining, value)
 	} else {
 		// Increment the nonce for the next transaction
-		st.state.SetNonce(msg.From, st.state.GetNonce(sender.Address())+1)
+		st.state.SetNonce(msg.From, st.msg.Nonce+1)
 		ret, st.gasRemaining, vmerr = st.evm.Call(sender, st.to(), msg.Data, st.gasRemaining, value)
 	}
 
--- a/core/txpool/legacypool/legacypool.go
+++ b/core/txpool/legacypool/legacypool.go
@@ -30,6 +30,7 @@ import (
 	"github.com/ethereum/go-ethereum/common/prque"
 	"github.com/ethereum/go-ethereum/consensus/misc/eip1559"
 	"github.com/ethereum/go-ethereum/core"
+	"github.com/ethereum/go-ethereum/core/encryption"
 	"github.com/ethereum/go-ethereum/core/state"
 	"github.com/ethereum/go-ethereum/core/txpool"
 	"github.com/ethereum/go-ethereum/core/types"
@@ -1714,7 +1715,8 @@ func (pool *LegacyPool) demoteUnexecutables() {
 			localGauge.Dec(int64(len(olds) + len(drops) + len(invalids)))
 		}
 		// If there's a gap in front, alert (should never happen) and postpone all transactions
-		if list.Len() > 0 && list.txs.Get(nonce) == nil {
+		// Skip the gap if an encrypt external tx is at nonce+1
+		if list.Len() > 0 && list.txs.Get(nonce) == nil && (!encryption.IsEncTx(list.txs.Get(nonce + 1))) {
 			gapped := list.Cap(0)
 			for _, tx := range gapped {
 				hash := tx.Hash()
--- a/core/txpool/legacypool/list.go
+++ b/core/txpool/legacypool/list.go
@@ -26,6 +26,7 @@ import (
 	"time"
 
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/encryption"
 	"github.com/ethereum/go-ethereum/core/types"
 )
 
@@ -209,7 +210,12 @@ func (m *sortedMap) Remove(nonce uint64) bool {
 // happen but better to be self correcting than failing!
 func (m *sortedMap) Ready(start uint64) types.Transactions {
 	// Short circuit if no transactions are available
-	if m.index.Len() == 0 || (*m.index)[0] > start {
+	if m.index.Len() == 0 {
+		return nil
+	}
+	// Check first nonce
+	isEncTx := encryption.IsEncTx(m.items[(*m.index)[0]])
+	if ((*m.index)[0] > start && !isEncTx) || ((*m.index)[0] > start+1 && isEncTx) {
 		return nil
 	}
 	// Otherwise start accumulating incremental transactions
@@ -218,6 +224,10 @@ func (m *sortedMap) Ready(start uint64) types.Transactions {
 		ready = append(ready, m.items[next])
 		delete(m.items, next)
 		heap.Pop(m.index)
+		// Check if next tx is encrypt external tx and nonce is next +2, increase next by 1 first
+		if m.index.Len() > 0 && encryption.IsEncTx(m.items[(*m.index)[0]]) && (*m.index)[0] == next+2 {
+			next++
+		}
 	}
 	m.cacheMu.Lock()
 	m.cache = nil
--- a/core/systemcontracts/contracts.go
+++ b/core/systemcontracts/contracts.go
@@ -30,6 +30,8 @@ var (
 	PolicyProxyHash = common.HexToAddress("0x1212000000000000000000000000000000000002")
 	// PolicyABI is a compiled ABI of Policy contract.
 	PolicyABI abi.ABI
+	// GovernanceRewardHash is a contract hash used to receive governance rewards.
+	GovernanceRewardHash = common.HexToAddress("0x1212000000000000000000000000000000000003")
 )
 
 func init() {


core/txpool/legacypool/list.go
```diff
@@ -218,6 +224,10 @@ func (m *sortedMap) Ready(start uint64) types.Transactions {
 		ready = append(ready, m.items[next])
 		delete(m.items, next)
 		heap.Pop(m.index)
+		// Check if next tx is encrypt external tx and nonce is next +2, increase next by 1 first
+		if m.index.Len() > 0 && encryption.IsEncTx(m.items[(*m.index)[0]]) && (*m.index)[0] == next+2 {
```
roman-khimov, line 228: What if I'm to send an envelope tx with `N+1` nonce and then replace it with a regular one by fee?

roman-khimov (commented):
Not exactly this PR subject, but what if I'm to send two transactions (current nonce is N):
 * envelope with N outside and N+1 inside
 * envelope with N+1 outside and N inside
?

Or
 * regular tx with N
 * envelope with N+1 and N inside

Or
 * envelope with N+1 outside and N+2 inside

Or
 * envelope with N+1 outside and an envelope with N+3 inside

Nonces are important in E-chain, I'm not sure we know all potential side-effects of trying to allow N+1 currently.

core/encryption/encryption.go
```diff
@@ -0,0 +1,16 @@
+package encryption
+
+import (
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+)
+
+const ENC_RECEIVER = "0x1212000000000000000000000000000000000003"
```
AnnaShaleva: I think we just have to reuse an existing address of the Governance contract specified in the `systemcontracts` package. Thus, please, move this constant to the `systemcontracts` and reuse it here.

AnnaShaleva: What is the etymology of ENC? Is it a short of "encrypted"?

core/encryption/encryption.go
```diff
@@ -0,0 +1,16 @@
...
+
+func IsEncReceiver(to *common.Address) bool {
+	return to != nil && *(to) == common.HexToAddress(ENC_RECEIVER)
```
AnnaShaleva: And we need to store `common.HexToAddress` in a package-level variable in order not to perform deserialization every time this code is invoked, because it's going to be a hot-path. Exactly like for `systemcontracts` hashes.

core/state_transition.go
```diff
@@ -445,7 +447,7 @@ func (st *StateTransition) TransitionDb() (*ExecutionResult, error) {
 		ret, _, st.gasRemaining, vmerr = st.evm.Create(sender, msg.Data, st.gasRemaining, value)
 	} else {
 		// Increment the nonce for the next transaction
-		st.state.SetNonce(msg.From, st.state.GetNonce(sender.Address())+1)
+		st.state.SetNonce(msg.From, st.msg.Nonce+1)
```
AnnaShaleva, line 450: Are you sure it's safe to refer to the message's nonce instead of the state's nonce? Is it possible that state's nonce differ from the message's nonce?

AnnaShaleva: Also, there are other usages of `st.state.GetNonce` that should likely be fixed.

core/txpool/legacypool/list.go
chenquanyu: If the regular one comes before the envelope tx is filled in block, then the envelope tx will be replaced and the regular one will be pending or queued until the tx with nonce `N` comes.

core/txpool/legacypool/legacypool.go
```diff
@@ -1714,7 +1715,8 @@ func (pool *LegacyPool) demoteUnexecutables() {
 			localGauge.Dec(int64(len(olds) + len(drops) + len(invalids)))
 		}
 		// If there's a gap in front, alert (should never happen) and postpone all transactions
-		if list.Len() > 0 && list.txs.Get(nonce) == nil {
+		// Skip the gap if an encrypt external tx is at nonce+1
+		if list.Len() > 0 && list.txs.Get(nonce) == nil && (!encryption.IsEncTx(list.txs.Get(nonce + 1))) {
```
AnnaShaleva, line 1719: How about the blobpool modification? What is the encrypted transaction is a blob transaction?

AnnaShaleva (commented):
We can't merge it until T4-compatible release is out.

chenquanyu:
> Not exactly this PR subject, but what if I'm to send two transactions (current nonce is N):

We first confirm whether the transaction in the envelope will be packaged, regardless of whether the transaction inside is normal. If the transaction nonce inside is wrong or causes insufficient fees, it will be discarded.

core/txpool/legacypool/list.go
roman-khimov: Well, if envelope one is to be evicted then I can cancel transactions for free, because regular one can't enter the mempool with this nonce.

core/state_transition.go
chenquanyu: > Are you sure it's safe to refer to the message's nonce instead of the state's nonce? Is it possible that state's nonce differ from the message's nonce?

This was checked at the begining of the method. 
> Also, there are other usages of st.state.GetNonce that should likely be fixed.

We only change it here because we don't want to change other behavior of state nonce.

core/txpool/legacypool/legacypool.go
chenquanyu: I don't think we should allow blob transaction as encrypted transaction.

roman-khimov:
Also, in regular E-chain, I can send multiple consecutive transactions like N, N+1, N+2. Potentially, all of them can end up in the same block. With envelopes it becomes like N+1, N+3, N+5 with potentially valid regular transactions in N, N+2, N+4 slots floating around. Not that I can think of some particular problem/attack here in 15m, but something to think about.

roman-khimov:
@chenquanyu, have you considered inner N == outer N? I know payment guarantee becomes a problem, but we're always considering this in the context of special payment that goes to some system contract while in fact a regular transaction fee can be sufficient (we can impose any restricitions on it), after all it's validators that do all of these dances, so it's OK for them to get some compensation and if anything is wrong with the payload then we just choose the outer tx that sends some fee to a system contract.

core/txpool/legacypool/list.go
chenquanyu: > Well, if envelope one is to be evicted then I can cancel transactions for free, because regular one can't enter the mempool with this nonce.

In fact, the future nonce regular tx can enter mempool, they are queued but not processable(in pending map).

core/encryption/encryption.go
chenquanyu: Moved to `systemcontracts `.

chenquanyu:
> @chenquanyu, have you considered inner N == outer N? I know payment guarantee becomes a problem, but we're always considering this in the context of special payment that goes to some system contract while in fact a regular transaction fee can be sufficient (we can impose any restricitions on it), after all it's validators that do all of these dances, so it's OK for them to get some compensation and if anything is wrong with the payload then we just choose the outer tx that sends some fee to a system contract.

I think the inner tx nonce must be `N-1` if the envelope tx nonce is `N`, otherwise the inner tx should be discarded.

roman-khimov:
I mean making it so that nonces are the same by design for the normal use case. This can simplify a lot of things node-side (envelope is just replaced by the contents).

chenquanyu:
> I mean making it so that nonces are the same by design for the normal use case. This can simplify a lot of things node-side (envelope is just replaced by the contents).

I think this design of using the same nonce is better, but the issue of fees needs further discussion. Do you want to charge a fee for decryption by increasing the gas price?

roman-khimov:
Yes, we can have requirements for gas price and/or priority fees of the inner transaction. If it's OK --- it replaces the envelope, if it's not --- envelope goes in.

core/encryption/encryption.go
```diff
@@ -0,0 +1,15 @@
...
+	"github.com/ethereum/go-ethereum/core/types"
+)
+
+func IsEncReceiver(to *common.Address) bool {
+	return to != nil && *(to) == systemcontracts.GovernanceRewardHash
+}
+
+func IsEncTx(tx *types.Transaction) bool {
+	return tx != nil && IsEncReceiver(tx.To())
+}
```
AnnaShaleva, line 15: Can we think about more reliable criteria for Envelopes detection? I can send funds to GovernanceReward contract and this transaction will be recognized as Envelope, although not intended to be so.

To be honest, I thought that we''ll create a fully-qualified separate transaction type for this (some `transaction.Envelope`), like `transaction.DynamicFeeTxType` or `transaction.BlobTxType`. Have you considered this idea?

core/txpool/legacypool/legacypool.go
AnnaShaleva: We need at least some basic documentation for Envelopes.

AnnaShaleva, line 1719: IIUC, we've agreed that Envelope nonce should be the same as inner nonce. If so, then let's update this PR wrt this statement.

core/encryption/encryption.go
AnnaShaleva: Oh, my bad, I see https://github.com/bane-labs/go-ethereum/pull/285. So this PR depends on #285.

AnnaShaleva: Although I see that `EncryptedTx` type is for inner encrypted transaction, and Envelope itself doesn't have a separate type, so the question is still relevant.

core/txpool/legacypool/legacypool.go
chenquanyu: I'm going to close this PR since we decided to use the same nonce for internal encrypted tx and external Envelope tx.

core/encryption/encryption.go
chenquanyu: The external transaction we designed now is a normal transfer transaction to the reward contract. The length of the data should be checked to ensure that it is greater than or equal to the minimum legal transaction length (assuming that the length remains unchanged after encryption).

core/txpool/legacypool/list.go
chenquanyu: Resolve this because we decide to use same nonce.

core/txpool/legacypool/legacypool.go
chenquanyu: Blob transaction has a different type.

AnnaShaleva:
@chenquanyu, this PR contains some useful changes regarding Envelopes detection, thus I'd suggest to adjust this PR instead of closing it. The point is that we need some additional fee management for Envelope transactions, because Envelope is a kind of "fallback" transaction that will be accepted to the chain if EncryptedTx fails to pass decryption/verification. And thus ,Envelope fees should be greater than standard transaction fees. Let's adjust this PR to properly check fees for Envelopes, we may use some constant encryption service fee for the first implementation and then adjust it to be Policy-based value.

What do you think?

core/encryption/encryption.go
AnnaShaleva: Discussed with Roman, so we'd better to avoid introducing yet another transaction type for Envelopes. I'm OK with it.

txhsl: > So this PR depends on https://github.com/bane-labs/go-ethereum/pull/285.

No, They are not relevant, this #235 is going to be closed, since @chenquanyu is now trying to stick to https://github.com/bane-labs/go-ethereum/pull/235#issuecomment-2230227119, and it is not to allow skip nonce by 1.

IMO, we can have a new transaction type for envelopes, since the mempool ordering needs to separate them for different space.

AnnaShaleva: > this https://github.com/bane-labs/go-ethereum/pull/235 is going to be closed anyway

Please, consider https://github.com/bane-labs/go-ethereum/pull/235#issuecomment-2298302411.

> since the mempool ordering needs to separate them for different space.

Can't it be done without additional type? Roman is against of yet another transaction type because our goal is to make as less changes in the original Geth codebase as possible. Envelopes should be handled in the same legacy pool as usual transactions, and the only difference for them is fee management (Envelope should pay some additional service fee), and from the first glance it seems possible to implement it without additional transaction type.

chenquanyu:
> @chenquanyu, this PR contains some useful changes regarding Envelopes detection, thus I'd suggest to adjust this PR instead of closing it. The point is that we need some additional fee management for Envelope transactions, because Envelope is a kind of "fallback" transaction that will be accepted to the chain if EncryptedTx fails to pass decryption/verification. And thus ,Envelope fees should be greater than standard transaction fees. Let's adjust this PR to properly check fees for Envelopes, we may use some constant encryption service fee for the first implementation and then adjust it to be Policy-based value.
> 
> What do you think?

We should charge encryption fees according to the length of encrypted transactions, so the encryption fee encryptFee should be known when obtaining the outer transaction data, and should be placed in the value. However, now only one of the outer and inner transactions will be retained. If the inner transaction replaces the outer transaction, then it is necessary to ensure that the gasPrice * gasUsed of the inner transaction >= minGasPrice * gasUsed + encryptFee. The main problem with this design is how users can set a reasonable gasPrice to ensure that the encryption fee is sufficient.

core/encryption/encryption.go
txhsl: >  I thought that we''ll create a fully-qualified separate transaction type for this.

Is it the question ↑ and answer ↓?

> Roman is against of yet another transaction type because our goal is to make as less changes in the original Geth codebase as possible.

AnnaShaleva:
> the main problem with this design is how users can set a reasonable gasPrice to ensure that the encryption fee is sufficient.

I see, and this problem is relevant for Metamask integration, right? And we can't really modify the `SuggestTipCap` RPC handler to return some increased value for encrypted transactions because this method doesn't take any additional arguments.

Could you please clarify why do we need `encryptFee` to be dependent from the encrypted tx length? May be we can simplify this scheme so that it's possible for the user to add some pre-defined fee before the transaction construction.

core/encryption/encryption.go
AnnaShaleva: Yes, exactly. Earlier I thought about separate type for Envelope, but then we discussed it with Roman, and seems that it's better without this separate type.

txhsl:
> Could you please clarify why do we need `encryptFee` to be dependent from the encrypted tx length?

Ref #301. Normally, an amev transaction needs 2 steps to get encrypted.
1. Generate a random `bls12_381` G1 point, and use it as an AES key to encrypt the transaction data;
2. Generate a random `big.int` number, and use it to encrypt the above `bls12_381` point, with the global public key.

And correspondingly, it takes 2 steps to decrypt it.
1. Request dBFT to share messages about the `big.int` number, and decrypt the `bls12_381` point;
2. Use this `bls12_381` point to decrypt the transaction data in AES.

So the cost of the decryption process consists of 2 parts.
1. A TPKE decryption on `bls12_381` curve;
2. An AES decryption about the transaction data.

The cost of Part 1 is constant, but the cost of Part 2 is dynamic, and is based on the size of the raw transaction to be encrypted.

Well, I know the AES computation is much quicker and the transaction might not be too huge, but a dynamic `encryptFee` seems reasonable to me.

roman-khimov:
> AES computation is much quicker

I think that within the limits of transaction size it should be orders of magnitude lower than any of the BLS math (please check though) and if that's the case then a simpler model with static fee is preferable.

txhsl:
> within the limits of transaction size

Well, I'm afraid there is no direct size limit.

The price of zero byte in Ethereum protocol is `TxDataZeroGas = 4`, while the price of non-zero byte is `TxDataNonZeroGasEIP2028 = 16`. The block gas limit we have now is `30,000,000`, so there could be at most `7 MB` data per block.

Normally, the cost of Part 2 is significantly lower than Part 1, but in the worst case (only 1 tx with 7 MB zero bytes), the cost of Part 2 will be higher than Part 1 (but the total time is the shortest).

Anyway, a simpler model with static fee is acceptable if the network security can be protected from DoS.

roman-khimov:
> 7 MB data per block

Not a very small number, but then we need to check this worst case against TPKE decryption. I'd still bet on TPKE being slower, but at the same time if we're to assume 1 GB/s speed of AES (pretty conservative for modern CPUs with AESNI to me) that'd be ~7ms and the numbers that I roughly recall for something BLSish are https://github.com/neo-project/neo/issues/1085#issuecomment-563959878. But TPKE is a bit more fancy than that to me. Let's test it and then we could make some decision on that.

chenquanyu:
> > within the limits of transaction size
> 
> Well, I'm afraid there is no direct size limit.
> 
> The price of zero byte in Ethereum protocol is `TxDataZeroGas = 4`, while the price of non-zero byte is `TxDataNonZeroGasEIP2028 = 16`. The block gas limit we have now is `30,000,000`, so there could be at most `7 MB` data per block.
> 
> Normally, the cost of Part 2 is significantly lower than Part 1, but in the worst case (only 1 tx with 7 MB zero bytes), the cost of Part 2 will be higher than Part 1 (but the total time is the shortest).
> 
> Anyway, a simpler model with static fee is acceptable if the network security can be protected from DoS.

The `legacypool` has a transaction size limit of 128KB.  https://github.com/bane-labs/go-ethereum/blob/1acd8231f6dfde010afe07b12f64d8d3b2f56114/core/txpool/legacypool/legacypool.go#L54

txhsl:
> The `legacypool` has a transaction size limit of 128KB.

Good to know, then the AES performance won't be a potential risk.

PR #236, opened 2024-06-28
Prepare to 0.2.0 release

### LINKED ISSUE #169: Remove privnet genesis templates
# Rationale

We have a set of static validator wallets for privnet, they likely will never be regenerated. At the same time it's just useless extra work to keep the genesis templates up-to-date and relevant wrt privnet genesis configuration files.

# Implementation

Refactor Makefile, remove privnet wallets generation part from it. Remove privnet genesis template files.

### LINKED ISSUE #175: Add documentation about NeoX node start
# Rationale

Although almost no changes were made in the node configuration flags, we're currently missing NeoX node start guide.

# Implementation

We need to adjust the README.md to reflect the relevant NeoX node information. We also need to add a documentation about how to start the NeoX node and join the testnet (and mainnet in future).

### LINKED ISSUE #224: Ensure Solidity compiler ^0.8.25 is compatible with dBFT-supported hardforks
Ref. https://github.com/bane-labs/go-ethereum/pull/208#discussion_r1620928382, ref. https://github.com/bane-labs/go-ethereum/issues/206. We can't update compiler version without the dBFT compatibility check.

### LINKED ISSUE #226: [Documentation] Integrate Treasury Contract in Neo X Architecture
The treasury introduced in #184 should be added to the architecture in the `/contracts/README.md`.

### LINKED ISSUE #231: Allocate empty upgradable system contracts in the genesis file for future usage
# Rationale

Currently, we already have several system contracts like GovReward, Governance and Policy, and they all have pre-allocated addresses for the proxy storage. But when we want to add new contracts like Treasury, it's hard to give them similar addresses since we can't change genesis allocation for a deployed network. So they may have to be deployed to normal smart contract addresses.

# Implementation

Since our system contract are upgradable and controlled by governance, we can allocate enough empty contracts in the genesis file, so that we can use these addresses in the future.

AnnaShaleva:
Depends on #230, #166.

Close #169, close #226, close #224, close #231, close #175.

This PR needs to be merged before T4-compatible release.


config/genesis_testnet.json
```diff
@@ -85,7 +85,7 @@
       }
     },
     "0x1212000000000000000000000000000000000004": {
-      "balance": "60000000000000000000000000",
+      "balance": "0", // TODO: do we need at least *some* initial balance for native BridgeProxy?
```
AnnaShaleva: @mialbu, @gsmachado, @steven1227 for T4 all native Bridge contract funds are moved to native Treasury contract. Do we need to supply Bridge contract with at least *some* initial amount of GAS in the genesis allocations?

config/genesis_testnet.json
```diff
@@ -121,6 +121,13 @@
         "0x360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbc": "0x0000000000000000000000001212100000000000000000000000000000000005"
       }
     },
+    "0x1212000000000000000000000000000000000006": {
+      "balance": "60000000000000000000000000", // TODO: need more?
```
AnnaShaleva: @mialbu, @gsmachado, @steven1227 what is the desired balance of native Treasury contract for T4?

config/genesis_testnet.json
```diff
@@ -121,6 +121,13 @@
...
+      "code": "0x6080604052600a600c565b005b60186014601a565b6051565b565b6000604c7f360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbc546001600160a01b031690565b905090565b3660008037600080366000845af43d6000803e808015606f573d6000f35b3d6000fdfea26469706673582212205473baeda08227b037acf1d10bff816015a11f298917e2465da958f40bcefac664736f6c63430008180033",
+      "storage": {
+        // TODO: fill the storage
```
AnnaShaleva: @chenquanyu, @txhsl, @mialbu I'm not sure how to get the initial storage items for Treasury proxy contract. Could you please, explain or provide them?

config/genesis_testnet.json
```diff
@@ -121,6 +121,13 @@
         "0x360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbc": "0x0000000000000000000000001212100000000000000000000000000000000005"
       }
     },
+    "0x1212000000000000000000000000000000000006": {
+      "balance": "60000000000000000000000000", // TODO: need more?
+      "code": "0x6080604052600a600c565b005b60186014601a565b6051565b565b6000604c7f360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbc546001600160a01b031690565b905090565b3660008037600080366000845af43d6000803e808015606f573d6000f35b3d6000fdfea26469706673582212205473baeda08227b037acf1d10bff816015a11f298917e2465da958f40bcefac664736f6c63430008180033",
```
AnnaShaleva: @chenquanyu, @txhsl all our native contracts are upgradable and supposed to have the same Proxy code, right? I noticed that some of our Proxy contracts have different code than the others. It's probably due to different compiler versions or something like that, but we need to unify this point. What is the desired code for the Proxy contracts?

config/genesis_testnet.json
```diff
@@ -140,6 +147,10 @@
...
...6575f83815260208190526040812083519091908490849081106104d9576104d9610664565b6020908102919091018101516001600160a01b031682528101919091526040015f908120556001016104ac565b505f828152602081905260408120906105226001830182610528565b50505050565b5080545f8255905f5260205f20908101906105439190610546565b50565b5b8082111561055a575f8155600101610547565b5090565b5f6020828403121561056e575f80fd5b5035919050565b634e487b7160e01b5f52604160045260245ffd5b80516001600160a01b038116811461059f575f80fd5b919050565b5f60208083850312156105b5575f80fd5b825167ffffffffffffffff808211156105cc575f80fd5b818501915085601f8301126105df575f80fd5b8151818111156105f1576105f1610575565b8060051b604051601f19603f8301168101818110858211171561061657610616610575565b604052918252848201925083810185019188831115610633575f80fd5b938501935b828510156106585761064985610589565b84529385019392850192610638565b98975050505050505050565b634e487b7160e01b5f52603260045260245ffd5b634e487b7160e01b5f52601160045260245ffd5b5f6001820161069d5761069d610678565b5060010190565b808201808211156102fd576102fd610678565b5f826106d157634e487b7160e01b5f52601260045260245ffd5b50049056fea2646970667358221220c8319d89dc3115c33f4137f9256b9f14df89cbc2e3495401a37dcb408a576a1864736f6c63430008190033"
```
AnnaShaleva: @chenquanyu, @txhsl, @mialbu could you please check that this Treasury implementation code matches the expectations? I compiled it using our hardhat config from `./contracts/hardhat.config.ts` with `evmVersion: "shanghai"`.

config/genesis_testnet.json
```diff
@@ -128,6 +128,34 @@
         // TODO: fill the storage
       }
     },
+    "0x1212000000000000000000000000000000000007": {
+      "balance": "0",
+      "code": "0x6080604052600a600c565b005b60186014601a565b6051565b565b6000604c7f360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbc546001600160a01b031690565b905090565b3660008037600080366000845af43d6000803e808015606f573d6000f35b3d6000fdfea26469706673582212205473baeda08227b037acf1d10bff816015a11f298917e2465da958f40bcefac664736f6c63430008180033",
+      "storage": {
+        // TODO
```
AnnaShaleva: @chenquanyu, @txhsl I need your help in storage allocations for systemcontracts stubs. Also, please check that Proxy/implementation code matches the expectations.

contracts/solidity/interfaces/IStub0.sol
```diff
@@ -0,0 +1,8 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.25;
+
+interface IStub0 {
+
+    // a stub function
+    function stub() external;
```
AnnaShaleva: @txhsl, @chenquanyu is this contract a valid Eth contract? We need to ensure that there will be no problem once we need to upgrade it.

AnnaShaleva (commented):
Please, consider answering these questions, it will help to finish this PR.

contracts/solidity/interfaces/IStub0.sol
txhsl: It is, but to make the upgrade possible, we need it to inherit `GovProxyUpgradeable.sol`, and contain the following codes. :)

https://github.com/bane-labs/go-ethereum/blob/7017f7aa5fe678b90496564ebb0d155d51095976/contracts/solidity/Governance.sol#L15

And.

https://github.com/bane-labs/go-ethereum/blob/7017f7aa5fe678b90496564ebb0d155d51095976/contracts/solidity/Governance.sol#L69-L87

The `SELF` address is the corresponding implementation address, which is **different** in each implementation file.

config/genesis_testnet.json
txhsl: Only a key `0x360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbc` with its value `0x1212100000000000000000000000000000000006`.

config/genesis_testnet.json
txhsl: Codes for addresses from `0x1212000000000000000000000000000000000001` to `0x1212000000000000000000000000000000000003` are provided by @chenquanyu and me, while from `0x1212000000000000000000000000000000000004` to `0x1212000000000000000000000000000000000005` are provided by AxLabs, so they could be different.

From my point of view, I will recommend the code of `0x1212000000000000000000000000000000000001`.

config/genesis_testnet.json
txhsl: The proxy code should be the same as other proxies', and the storage allocation needs only a key `0x360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbc` with its corresponding implementation address, just like the Treasury allocates.

contracts/solidity/interfaces/IStub0.sol
chenquanyu: Maybe we can write a new contract to store `SELF` in a specific storage key. Then we only need one contract for all the empty upgradable system contracts.

txhsl: These codes and value is temporary usage, because UUPS proxy can't get correctly initialized if allocated in genesis file. After the first upgrade, new implementation contracts will get `natural` addresses and don't need these any more.

config/genesis_testnet.json
AnnaShaleva: Implementation storage item is added. @mialbu, could you please confirm that no additional storage items are needed?

config/genesis_testnet.json
AnnaShaleva: > I will recommend the code of 0x1212000000000000000000000000000000000001

Fixed, let's use it for new contracts.

config/genesis_testnet.json
AnnaShaleva: Fixed, could you check please?

contracts/solidity/interfaces/IStub0.sol
AnnaShaleva: > to make the upgrade possible, we need it to inherit GovProxyUpgradeable.sol

Oh gosh, I was just copy-pasting the Treasury contract and completely forgot that Treasury is not upgradeable. Fixed, @txhsl, could you please check one more time?

config/genesis_testnet.json
txhsl: LGTM.

contracts/solidity/interfaces/IStub0.sol
txhsl: Yeh, if Treasury is not upgradeable, then things are similar with GovProxyAdmin at `0x1212000000000000000000000000000000000000`. There will not be an implementation address, but only the implementation code in `0x1212000000000000000000000000000000000006`, and no need to allocate `0x360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbc`.

txhsl: BTW, we have updated nearly most of the contracts (including the bridge), due to the audit and some new features, so their implementation codes need to be compiled and updated again as well.

AnnaShaleva: > Yeh, if Treasury is not upgradeable, then things are similar with GovProxyAdmin

Fixed, check it please.

> so their implementation codes need to be compiled and updated again as well.

Implementations are updated for GovernanceProxyAdmin, Governance, Policy and GovernanceReward contracts. But storage items are still not updated, I need your help to update storage items.

To update Bridge and BridgeManagement implementaitons I need the final version of these contracts for T4 from @mialbu.

config/genesis_testnet.json
AnnaShaleva: UPD: Treasury is not upgradeable, thus we use a single genesis allocation for it with `0x1212000000000000000000000000000000000006` address and an empty storage items set.

config/genesis_testnet.json
mialbu: Imo there should be some initial allocation to the bridge. Otherwise, we'd require a governance vote before the bridge can be used.

config/genesis_testnet.json
mialbu: I'm fixing some things from the audit still. It'd be great if we can use that new contract for the start of T4. The medium issues are already fixed (unmerged PRs), and there's one or two low severity issues that I'll implement tomorrow. After that I'll need to have another look at the storage entries anyway.

contracts/solidity/interfaces/IStub0.sol
mialbu: > To update Bridge and BridgeManagement implementaitons I need the final version of these contracts for T4 from @mialbu.

https://github.com/bane-labs/go-ethereum/pull/236#discussion_r1662813824

txhsl: >  But storage items are still not updated, I need your help to update storage items.

Since I made the contract storage compatible, we only need to add a new field for `candidateLimit`. Also, @steven1227 wants to increase both the `baseFee` and `minGasTipCap` to `20 Gwei`.
```json
"0x1212000000000000000000000000000000000002": {
      "storage": {
        "0x0000000000000000000000000000000000000000000000000000000000000002": "0x04a817c800",
        "0x0000000000000000000000000000000000000000000000000000000000000003": "0x04a817c800",
        "0x0000000000000000000000000000000000000000000000000000000000000004": "0x07d0",
        "0x360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbc": "0x1212100000000000000000000000000000000002"
      }
    },
```

contracts/solidity/interfaces/IStub1.sol
```diff
@@ -0,0 +1,8 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.25;
+
+interface IStub1 {
```
chenquanyu: Why do we need interface and method for stub contracts?

config/genesis_testnet.json
AnnaShaleva: Yes, it's the same chicken-and-egg problem. So how much funds do you think we should give to the bridge?

config/genesis_testnet.json
AnnaShaleva: It's OK, just write back once the contracts are in their final state.

contracts/solidity/interfaces/IStub1.sol
AnnaShaleva: It's just a template for further extension, but if you'd like then I can remove interface and remove this method from this contract. Agree?

contracts/solidity/interfaces/IStub0.sol
AnnaShaleva: Updated. And just to clarify: we don't need to change storage items for Governance or GovernanceVote contracts, right? There were changes in these contracts, but these changes did not affect the initializable variables.

txhsl: > There were changes in these contracts, but these changes did not affect the initializable variables.

Yes.

contracts/solidity/interfaces/IStub1.sol
chenquanyu: Maybe we don't need so many new empty implementation contracts, we can just point all these proxy contracts to the same implementation contract.

AnnaShaleva: > I can remove interface and remove this method from this contract

Done.

> we can just point all these proxy contracts to the same implementation contract.

Can we? Implementation contract should be able to control the upgrade and all implementation contracts have different `SELF`:
```
address public constant SELF = 0x1212100000000000000000000000000000000009;
```

AnnaShaleva: And I don't think that it's correct to point to the shared implementation from different proxies because we may have problems once we're going to update the implementation for one of the proxies. It's not quite clear to me what will happen with the rest of proxies in this case.

Could you please clarify? May be I'm wrong.

chenquanyu: > And I don't think that it's correct to point to the shared implementation from different proxies because we may have problems once we're going to update the implementation for one of the proxies. It's not quite clear to me what will happen with the rest of proxies in this case.
> 
> Could you please clarify? May be I'm wrong.

Nothing will happen to other proxies  when you upgrade one proxy, you just change the implementation slot value when upgrading.

config/genesis_testnet.json
gsmachado: I'd say we need just a bit of GAS for this. Like, 50 GAS to be on the safe side? wdyt?

steven1227: Maybe a tiny amount is enough ?

gsmachado: Let's set 5 GAS, then. Or, I'm also fine with 1 GAS, if it's enough.

@mialbu will this be more than enough for the initial steps, right?

mialbu: Actually, if it's a tiny amount, we need to fund it anyway rather soon already. Then I'm not sure if there's any point in allocating any funds to the bridge in the genesis.

AnnaShaleva: Initial transfer from Treasury (where all Bridge funds are stored) requires votes from CN accounts. CN accounts should be supplied with GAS before they can vote. We don't put genesis allocations for CN accounts, they are supposed to bridge initial amount of funds to themselves. So the question is: can we perform these initial bridging to CN accounts with no GAS on the Bridge contract balance? I suppose we can't, that's why we need at least *some* amount for this genesis allocation.

config/genesis_testnet.json
AnnaShaleva: As discussed, changed to `100,000,000`.

contracts/solidity/interfaces/IStub1.sol
AnnaShaleva: @txhsl, agree to use a single stub *implementation*?

config/genesis_testnet.json
mialbu: Oh okay, I wasn't aware of that. Yes, in that case the bridge needs some initial allocation, absolutely.

There is a minimum deposit amount on the N3 side. It's been 1 Gas so far, but it can be modified as well. So, the initial allocation should be at least that amount, and it also needs to be enough to cover all the fees for the necessary transactions to fund the bridge. If 1 Gas is sufficient for all required transactions, then 1 Gas is sufficient as initial allocation to the bridge.

contracts/solidity/interfaces/IStub1.sol
AnnaShaleva: @chenquanyu fixed, please check.

config/genesis_testnet.json
```diff
@@ -47,7 +47,7 @@
         "0x0000000000000000000000000000000000000000000000000000000000000001": "0x07",
         "0x0000000000000000000000000000000000000000000000000000000000000002": "0x0de0b6b3a7640000",
         "0x0000000000000000000000000000000000000000000000000000000000000003": "0x027b46536c66c8e3000000",
-        "0x0000000000000000000000000000000000000000000000000000000000000004": "0x3635c9adc5dea00000",
+        "0x0000000000000000000000000000000000000000000000000000000000000004": "0x43c33c1937564800000",
```
AnnaShaleva: @txhsl, registration fee is set to 20K GAS, please check it's the right storage item.

txhsl: It is the expected key and number value, but for a correct format without error, we may need `0x043c33c1937564800000` here.

contracts/solidity/interfaces/IStub1.sol
txhsl: > @txhsl, agree to use a single stub _implementation_?

I'm OK with that if it works.

config/genesis_testnet.json
txhsl: If this value changes before launch, I will let @steven1227 inform here directly.

config/genesis_testnet.json
txhsl: I'm not sure, but it is somehow different with my deploy code. @chenquanyu Can you check again as well?

config/genesis_testnet.json
AnnaShaleva: > but for a correct format without error, we may need 0x043c33c1937564800000 here.

Fixed.

contracts/solidity/interfaces/IStub1.sol
AnnaShaleva: @chenquanyu could you please check that at least one stub can be correctly upgraded on privnet?

config/genesis_testnet.json
AnnaShaleva: I used the updated hardhat config from this PR to compile:
```
import { HardhatUserConfig } from "hardhat/config";
import "@nomicfoundation/hardhat-toolbox";

const config: HardhatUserConfig = {
  solidity: {
    version: "0.8.25",
    settings: {
      evmVersion: "shanghai",
      optimizer: {
        enabled: true,
        runs: 500,
      },
    },
  },
  paths: {
    sources: "./solidity",
    tests: "./test",
  }
};

export default config;
```

Then did `npx hardhat compile` and then took the code from `deployedBytecode` field of `go-ethereum/contracts/artifacts/solidity/Treasury.sol/Treasury.json`. Am I doing something wrong?

config/genesis_testnet.json
AnnaShaleva: 1GAS for BridgeProxy contract is allocated.

config/genesis_testnet.json
txhsl: No, not. I reset the env and get the same result as this line.

contracts/solidity/interfaces/IStub1.sol
chenquanyu: > @chenquanyu could you please check that at least one stub can be correctly upgraded on privnet?

I have checked that all stub contracts can be upgraded correctly on privnet. Maybe we should add more stub contracts to meet future needs.

AnnaShaleva: What is the desired number? Do you have exact contract examples that should be added in future (except DKG)?

AnnaShaleva: OK, how about 10 stub contracts? It's a lot, and I think we can fit our future needs into these contracts for sure.

txhsl: Yes, it's a lot. Or just from `0x1212000000000000000000000000000000000000` to `0x121200000000000000000000000000000000000f`.

AnnaShaleva: Added.

AnnaShaleva:
@txhsl @chenquanyu waiting for you approval before the merge, Roman is on vacation.

txhsl (approved):

PR #238, opened 2024-07-03
systemcontract: check new candidate limit with network settings

### LINKED ISSUE #237: Governance candidate limit should be checked to be not lower than consensus size
#### System information

Geth version: `geth version`
CL client & version: e.g. lighthouse/nimbus/prysm@v1.0.0
Commit hash : (if `develop`)

#### Expected behaviour
There are already 12 candidates, and the consensus node changes to limit the number of registered candidates to 7. At this time, the program should report an error

#### Actual behaviour
The program setCandidateLimit succeeded

#### Backtrace

````
[backtrace]
````

When submitting logs: please submit them as text and not screenshots.

### PR DESCRIPTION
**Pull request by txhsl**
Close #237.

### DIFF
--- a/contracts/solidity/Policy.sol
+++ b/contracts/solidity/Policy.sol
@@ -2,12 +2,15 @@
 pragma solidity ^0.8.25;
 
 import {Errors} from "./libraries/Errors.sol";
+import {IGovernance} from "./interfaces/IGovernance.sol";
 import {IPolicy} from "./interfaces/IPolicy.sol";
 import {GovernanceVote} from "./base/GovernanceVote.sol";
 import {ERC1967Utils, GovProxyUpgradeable} from "./base/GovProxyUpgradeable.sol";
 
 contract Policy is IPolicy, GovernanceVote, GovProxyUpgradeable {
     address public constant SELF = 0x1212100000000000000000000000000000000002;
+    // governance contact
+    address public constant GOV = 0x1212000000000000000000000000000000000001;
     uint256 public constant DEFAULT_CANDIDATE_LIMIT = 2000;
 
     mapping(address => bool) public isBlackListed;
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: contracts/solidity/Policy.sol [hunk 1]
```diff
@@ -115,7 +118,7 @@ contract Policy is IPolicy, GovernanceVote, GovProxyUpgradeable {
             keccak256(abi.encode(_candidateLimit))
         )
     {
-        if (_candidateLimit <= 0) revert Errors.InvalidCandidateLimit();
+        if (_candidateLimit < IGovernance(GOV).consensusSize()) revert Errors.InvalidCandidateLimit();
```
AnnaShaleva, line 121: The fix itself is legit, but I have a question: currently Governance contract reverts `registerCandidate` execution if candidates limit is reached:
https://github.com/bane-labs/go-ethereum/blob/971cac59f17c2c4673f8170daf2a9fc94acee74b/contracts/solidity/Governance.sol#L120-L121
But nothing happens with *existing* candidates if candidate limit is decreased in Policy. From one hand, it's not quite correct since Governance has more candidates than allowed by Policy. But from the other hand, it's not critical since no more candidates are allowed to be registered. The question is: are we OK with it?

AnnaShaleva (commented):
Otherwise LGTM, it's really nice that we have unit-tests for contracts.

contracts/solidity/Policy.sol
txhsl: Yes, it is not a critical problem, Governance can work safely without this fix.

> From one hand, it's not quite correct since Governance has more candidates than allowed by Policy.

Unfortunately, we can't delete candidates when decreasing the limit, which may require a huge gas cost for ordering, so I'm OK with that.

The issue was raised by our test team, and they think it is a logic concern, which means Policy should be consistent with network settings.

I'm open to any opinions, if we don't need this fix, then we can close.

AnnaShaleva: We definitely need this fix.

AnnaShaleva (approved):

PR #241, opened 2024-07-03
systemcontract: multiply before divide

txhsl:


dusmart:
the values of (𝑎÷𝑏)÷𝑐 and 𝑎÷(𝑏×𝑐) are equal.

what `multiply before divide` principle state is that use (a*b)/c instead of (a/c)*b

https://math.stackexchange.com/questions/2557458/is-a-b-c-equal-to-a-bc-for-integer-division#:~:text=This%20corresponds%20to%20the%20integer,b%C3%97c)%20are%20equal.


By the way, the price for `*` and `/` are the same.

https://github.com/bane-labs/go-ethereum/blob/cada98a3fa21eda3d8295a3e0f491c1b6674014f/core/vm/jump_table.go#L267-L272

https://github.com/bane-labs/go-ethereum/blob/cada98a3fa21eda3d8295a3e0f491c1b6674014f/core/vm/jump_table.go#L279-L284

I'm Ok if the auditing team insist this change. This won't bring side effect anyway.

AnnaShaleva (commented):
Well, I'm also OK with these changes. Precision loss is inevitable anyway.

AnnaShaleva:
Should anyone else review this PR?

PR #242, opened 2024-07-04
systemcontract: apply blacklist policy to governance election

### LINKED ISSUE #239: Black listed Policy feature is not used by Governance
#### Expected behaviour

Addresses marked as "blackListed" in Policy system contract should be banned from:
* calling contracts, deploying contracts, transferring funds (this may be unified in the implementation by blocking transactions with blocked sender)
* elections in consensus members (this may be implemented exactly like in N3: all addresses are allowed to be registered as a candidate, but only non-blocked addresses are returned from `getCandidates` call and be elected as CN, see https://github.com/nspcc-dev/neo-go/blob/0f2229d6a5e65f1e5608e1ed2ebb1e9e59c5c6c3/pkg/core/native/native_neo.go#L1024).

#### Actual behaviour

Address may be blocked/unblocked via Policy system contract interface. But it does not affect any operations on the network. Right now Policy's "blackListed" addresses are no more than records in the contract.

txhsl:
Close #239. Testing with 2000 `candidateLimit`...

### DIFF
--- a/contracts/solidity/Governance.sol
+++ b/contracts/solidity/Governance.sol
@@ -14,8 +14,7 @@ contract Governance is IGovernance, ReentrancyGuard, GovProxyUpgradeable {
 
     address public constant SELF = 0x1212100000000000000000000000000000000001;
     // Policy contract
-    address public constant POLICY = 
-        0x1212000000000000000000000000000000000002;
+    address public constant POLICY = 0x1212000000000000000000000000000000000002;
     // GovReward contract
     address public constant GOV_REWARD =
         0x1212000000000000000000000000000000000003;
@@ -65,6 +64,8 @@ contract Governance is IGovernance, ReentrancyGuard, GovProxyUpgradeable {
     mapping(address => uint) public voteHeight;
     // candidate=>height=>number
     mapping(address => mapping(uint => uint)) public epochStartGasPerVote;
+    // blacklisted candidate amount
+    uint public blacklistedCandidates;
 
     // Only for precompiled uups implementation in genesis file, need to be removed when upgrading the contract.
     // This override is added because "immutable __self" in UUPSUpgradeable is not avaliable in precompiled contract.
@@ -117,29 +118,21 @@ contract Governance is IGovernance, ReentrancyGuard, GovProxyUpgradeable {
         if (tx.origin != msg.sender) revert Errors.OnlyEOA();
         if (msg.value < registerFee) revert Errors.InsufficientValue();
         if (shareRate > 1000) revert Errors.InvalidShareRate();
-        if (candidateList.length() >= IPolicy(POLICY).getCandidateLimit())
-            revert Errors.RegisterDisabled();
+        if (
+            candidateList.length() + blacklistedCandidates >=
+            IPolicy(POLICY).getCandidateLimit()
+        ) revert Errors.RegisterDisabled();
         if (exitHeightOf[msg.sender] > 0) revert Errors.LeftNotClaimed();
-        if (!candidateList.add(msg.sender)) revert Errors.CandidateExists();
-        if (receivedVotes[msg.sender] > 0) {
-            totalVotes += receivedVotes[msg.sender];
-        }
+        if (!_activateCandidate(msg.sender)) revert Errors.CandidateExists();
 
         // record share rate and balance
         shareRateOf[msg.sender] = shareRate;
         candidateBalanceOf[msg.sender] = msg.value;
-        emit Register(msg.sender);
     }
 
     function exitCandidate() external {
-        if (!candidateList.remove(msg.sender))
+        if (!_deactivateCandidate(msg.sender))
             revert Errors.CandidateNotExists();
-        // remove candidate list, balance still locked
-        exitHeightOf[msg.sender] = block.number;
-        if (receivedVotes[msg.sender] > 0) {
-            totalVotes -= receivedVotes[msg.sender];
-        }
-        emit Exit(msg.sender);
     }
 
     function withdrawRegisterFee() external nonReentrant {
@@ -277,8 +270,40 @@ contract Governance is IGovernance, ReentrancyGuard, GovProxyUpgradeable {
         emit Persist(currentConsensus);
     }
 
+    function activateCandidate(address candidate) external {
+        if (msg.sender != POLICY) revert Errors.SideCallNotAllowed();
+        if (exitHeightOf[candidate] > 0 && _activateCandidate(candidate))
+            blacklistedCandidates -= 1;
+    }
+
+    function deactivateCandidate(address candidate) external {
+        if (msg.sender != POLICY) revert Errors.SideCallNotAllowed();
+        if (_deactivateCandidate(candidate)) blacklistedCandidates += 1;
+    }
+
     function getCurrentConsensus() public view returns (address[] memory) {
         return currentConsensus;
+    }
+
+    function _activateCandidate(address candidate) internal returns (bool) {
+        if (!candidateList.add(candidate)) return false;
+        delete exitHeightOf[candidate];
+        if (receivedVotes[candidate] > 0) {
+            totalVotes += receivedVotes[candidate];
+        }
+        emit Activate(candidate);
+        return true;
+    }
+
+    function _deactivateCandidate(address candidate) internal returns (bool) {
+        if (!candidateList.remove(candidate)) return false;
+        // remove candidate list, balance still locked
+        exitHeightOf[candidate] = block.number;
+        if (receivedVotes[candidate] > 0) {
+            totalVotes -= receivedVotes[candidate];
+        }
+        emit Deactivate(candidate);
+        return true;
     }
 
     function _computeReward(
--- a/contracts/solidity/Policy.sol
+++ b/contracts/solidity/Policy.sol
@@ -52,6 +52,7 @@ contract Policy is IPolicy, GovernanceVote, GovProxyUpgradeable {
     {
         if (isBlackListed[_addr]) revert Errors.BlacklistExists();
         isBlackListed[_addr] = true;
+        IGovernance(GOV).deactivateCandidate(_addr);
         emit AddBlackList(_addr);
     }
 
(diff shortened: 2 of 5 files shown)


txhsl:
With this policy check, the total gas cost of `onPersist` in the worst case will increase about 25%, making it slightly out of gas. We may have to lower down the `candidateLimit` at the same time to 1500 or so, or increase the syscall gas limit to 40_000_000 or so.

contracts/solidity/Governance.sol
```diff
@@ -344,13 +344,13 @@ contract Governance is IGovernance, ReentrancyGuard, GovProxyUpgradeable {
...
         uint k
-    ) internal pure {
+    ) internal view {
         uint length = candidates.length;
         for (int j = int(k) / 2 - 1; j >= 0; j--) {
             _heapDown(candidates, votes, uint(j), k);
         }
         for (uint i = k; i < length; i++) {
-            if (votes[i] > votes[0]) {
+            if (votes[i] > votes[0] && !IPolicy(POLICY).isBlackListed(candidates[i])) {
```
AnnaShaleva: This cycle starts from `i=k`, and in best case there will be no iterations in this cycle, and it seems to me that in this case candidates 0-k won't be checked. But what happens if one of candidates 0-k is blocked?

AnnaShaleva: And one more question: unlikely it'll happen, but we have maximum candidates number restriction. If the restriction is close to the number of CN, and sufficient number of candidates are blocked, then there might be not enough "good" candidates to choose CN from. In this case we should fall back to the stand-by CNs.

I understand that this situation likely won't happen because candidates limit is high and blocked accounts is an emergency situation, so it's not a hot path. But we at least may create an issue for that.

AnnaShaleva:
> making it slightly out of gas.

I think we'd better increase the syscall GAS limit.

contracts/solidity/Governance.sol
txhsl: Oh yes, I missed it.

txhsl: I'm trying to solve your last comment, and I think I can make the fallback possible with more codes.

txhsl: BTW, should we remove the votes of blacklisted validators from effective votes for the `targetAmount` check and prevent blacklisted validators to receive more votes?

txhsl:
@AnnaShaleva OK, now I have a good new idea to apply blacklist.

contracts/solidity/Governance.sol
AnnaShaleva: > should we remove the votes of blacklisted validators from effective votes for the targetAmount check and prevent blacklisted validators to receive more votes?

It seems natural and more correct, but I don't think we should, because it complicates voting management. I think that the fact that `_computeConsensus` prevents blacklisted addresses from being elected is enough for clear and safe network maintaining. For example, in NeoN3 this Policy logic works the same way: only `getCandidates` logic of native Neo contract adjusted to check the blocked accounts, so that blocked accounts can't appear in the list of candidates or be elected as consensus members. Other than that, users are allowed to vote for banned account, because [technically] there's an ability for blocked account to become unblocked.

What we definitely should do is to adjust `getCandidates` function in order not to return blocked accounts.

contracts/solidity/Governance.sol
```diff
@@ -277,10 +269,26 @@ contract Governance is IGovernance, ReentrancyGuard, GovProxyUpgradeable {
         emit Persist(currentConsensus);
     }
 
+    function forceExitCandidate(address candidate) external {
+        if (msg.sender != POLICY) revert Errors.SideCallNotAllowed();
+        _exitCandidate(candidate);
+    }
```
AnnaShaleva: To be honest, the previous version was more preferable to me, because the purpose of blocking is to prevent the user from activity on the network, but not to perform any state-changing actions against the user's will. And the current version actually gives the rights to the committee to perform unregistration action without user's behalf which doesn't seem to be ideologically correct.

AnnaShaleva: See also https://github.com/bane-labs/go-ethereum/pull/242#discussion_r1665570818.

AnnaShaleva: The difference between these two approaches is: when account will be unblocked, its voting status remains the same as it was before blocking (with the initial approach); and its voting status will be reset (with the updated approach).

contracts/solidity/Governance.sol
AnnaShaleva: Have you considered creating a separate `_getCandidates` function that will filter out all blocked candidates from the `candidateList` and return only valid candidates? Then we can reuse this function for both public `getCandidates()` and `_computeConsensus`. The only potential drawback is the execution cost, I suppose it's expensive.

txhsl: > Have you considered creating a separate `_getCandidates` function that will filter out all blocked candidates from the `candidateList` and return only valid candidates? Then we can reuse this function for both public `getCandidates()` and `_computeConsensus`. The only potential drawback is the execution cost, I suppose it's expensive.

It will be expensive, with a loop. Loop is not a healthy code in contract.

contracts/solidity/Governance.sol
txhsl: The exited state before claiming back is recoverable, what if I add something like an 'activate' method for blacklist removing?

Updated.

contracts/solidity/Governance.sol
```diff
@@ -132,13 +132,9 @@ contract Governance is IGovernance, ReentrancyGuard, GovProxyUpgradeable {
...
-        if (!candidateList.remove(msg.sender))
+        if (!_deactivateCandidate(msg.sender))
             revert Errors.CandidateNotExists();
-        // remove candidate list, balance still locked
-        exitHeightOf[msg.sender] = block.number;
-        if (receivedVotes[msg.sender] > 0) {
-            totalVotes -= receivedVotes[msg.sender];
-        }
+        deactivatedCandidates += 1;
         emit Exit(msg.sender);
```
AnnaShaleva: `_deactivateCandidate` emits `Deactivate` notification. Maybe then we should remove `Exit` notification? It feels like `Activate`/`Deactivate` duplicate `Register`/`Exit` events.

contracts/solidity/Governance.sol
```diff
@@ -221,7 +218,8 @@ contract Governance is IGovernance, ReentrancyGuard, GovProxyUpgradeable {
         uint amount = votedAmount[msg.sender];
         if (candidateFrom == address(0) || amount <= 0) revert Errors.NoVote();
         if (candidateFrom == candidateTo) revert Errors.SameCandidate();
-        if (!candidateList.contains(candidateTo)) revert Errors.CandidateNotExists();
+        if (!candidateList.contains(candidateTo))
+            revert Errors.CandidateNotExists();
```
AnnaShaleva: No functional changes.

contracts/solidity/Governance.sol
```diff
@@ -132,13 +132,9 @@ contract Governance is IGovernance, ReentrancyGuard, GovProxyUpgradeable {
...
     function exitCandidate() external {
-        if (!candidateList.remove(msg.sender))
+        if (!_deactivateCandidate(msg.sender))
             revert Errors.CandidateNotExists();
-        // remove candidate list, balance still locked
-        exitHeightOf[msg.sender] = block.number;
-        if (receivedVotes[msg.sender] > 0) {
-            totalVotes -= receivedVotes[msg.sender];
-        }
+        deactivatedCandidates += 1;
```
AnnaShaleva: I think we need to change `deactivatedCandidates` only inside policy-callable methods in order not to mix it with "good" candidates.

AnnaShaleva: Otherwise I'd expect `regesterCandidate` and `exitCandidate` be symmetric wrt `deactivatedCandidates` increment/decrement, and currently they are not.

contracts/solidity/Governance.sol
txhsl: `Register`/`Exit` events now get removed.

contracts/solidity/Governance.sol
txhsl: I was thinking about let candidates claim back themselves when I add `deactivatedCandidates`, but now they don't have to do that. Good suggestions, fixed.

contracts/solidity/Governance.sol
```diff
@@ -277,10 +270,42 @@ contract Governance is IGovernance, ReentrancyGuard, GovProxyUpgradeable {
...
         return currentConsensus;
     }
 
+    function _activateCandidate(address candidate) internal returns (bool) {
+        if (!candidateList.add(candidate)) return false;
+        delete exitHeightOf[candidate];
+        if (receivedVotes[candidate] > 0) {
+            totalVotes += receivedVotes[candidate];
+        }
+        emit Activate(candidate);
```
AnnaShaleva, line 294: And just a minor note: if now we use a single notification for both usual registration and activation, then probably "Registered" is more preferable than "Activate"? Because in 99% cases it's a usual candidate registration, so users may be confused by receiving "Activate" notification rather than "Registered".

contracts/solidity/Governance.sol
```diff
@@ -277,10 +270,42 @@ contract Governance is IGovernance, ReentrancyGuard, GovProxyUpgradeable {
...
+    function activateCandidate(address candidate) external {
+        if (msg.sender != POLICY) revert Errors.SideCallNotAllowed();
+        if (exitHeightOf[candidate] > 0 && _activateCandidate(candidate))
+            blacklistedCandidates -= 1;
+    }
+
+    function deactivateCandidate(address candidate) external {
+        if (msg.sender != POLICY) revert Errors.SideCallNotAllowed();
+        if (_deactivateCandidate(candidate)) blacklistedCandidates += 1;
+    }
```
AnnaShaleva, line 282: And a final note regarding blocked accounts: suppose candidate is in the list and is currently a CN. Once committee blocks this account, then this candidate won't be kicked off from the CN list immediately, instead it will remain there until the next epoch. We have a similar bug in NeoGo (https://github.com/nspcc-dev/neo-go/issues/3449), and C# node handles this situation properly.

We may create an issue to handle that in future.

AnnaShaleva (commented):
Otherwise LGTM.

contracts/solidity/Governance.sol
txhsl: It's fine, IMO.

contracts/solidity/Governance.sol
AnnaShaleva: Ref. #248.

PR #243, opened 2024-07-04
protocols/dbft: verify extensible payloads against extensible whitelist

### LINKED ISSUE #240: Verify consensus message sender against extensible whitelist
#### Expected behaviour

Consensus message sender should be verified against the whitelist of extensible payload senders. This list should be constructed based on the validators from Governance contract and should exclude Policy's black listed addresses.

#### Actual behaviour

No verification present:
https://github.com/bane-labs/go-ethereum/blob/cada98a3fa21eda3d8295a3e0f491c1b6674014f/eth/protocols/dbft/ledger.go#L23-L26

AnnaShaleva:
Currently extensible payloads include only consensus payloads, and extensible whitelist includes only validators. This logic may be extended in future to allow additional senders and payload types.

Note that consensus nodes verified consensus payloads before this commit by using OnPayload verification. In other words, this commit makes non-CN nodes verify extensible payloads as far.

Close #240.

@chenquanyu, @txhsl, please review. @songb2 I tested this functionality for both CN and seed nodes under assumption of unchanged CNs. Could you please test the following scenario:
1. Run privnet with stand-by validators.
2. Vote for the new set of validators and wait for the next epoch so that validators list is updated.
3. Check that consensus payloads are handled after new epoch starts. You'll see the following message in logs if something goes wrong: `WARN [07-04|13:24:08.093] can't broadcast consensus message        error="disallowed sender"`.

### DIFF
--- a/consensus/dbft/dbft.go
+++ b/consensus/dbft/dbft.go
@@ -1421,6 +1421,19 @@ func (c *DBFT) validatePayload(p *Payload) error {
 	return nil
 }
 
+// IsExtensibleAllowed determines if address is allowed to send extensible payloads
+// (only consensus payloads for now) at the specified height.
+func (c *DBFT) IsExtensibleAllowed(h uint64, u common.Address) bool {
+	// Only validators are included into extensible whitelist for now.
+	validators, err := c.getValidators(&h, nil, nil)
+	if err != nil {
+		return false
+	}
+	n := sort.Search(len(validators), func(i int) bool { return validators[i].Cmp(u) >= 0 })
+	res := n < len(validators)
+	return res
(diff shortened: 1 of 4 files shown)


txhsl:
Looks good.

txhsl:
It somehow breaks the test here.

https://github.com/bane-labs/go-ethereum/blob/2d4794628ce984805822229c27376417ba56d23f/eth/protocols/dbft/handler_test.go#L31-L33

songb2:
@AnnaShaleva It's normal to switch to other validators. I tested on local private net according to your steps. It can update validators and no warning message appears.

PR #247, opened 2024-07-08
systemcontract: add exit fee

### LINKED ISSUE #246: Add Governance exit fee
# Rationale

There is a DoS risk that attackers register new candidates to occupy the candidate limit capacity, with a very low cost (only transaction fees), since the Governance register fee is now fully returned when a candidate exits.

# Implementation

1. Add an exit fee in Governance contract, so that the contract will not transfer all things back;
2. Increase the register fee field in genesis file allocation.

txhsl:
Close #246.

### DIFF
--- a/contracts/README.md
+++ b/contracts/README.md
@@ -52,13 +52,13 @@ An EOA account is allowed to become a candidate only after successful registrati
 
 1. Registrant invokes `registerCandidate(uint shareRate)` of `0x1212000000000000000000000000000000000001` as message sender;
 2. Registrant is an EOA account and not yet a candidate;
-3. Put at least `1000 GAS` deposit `value` along with the transaction as registration fee;
+3. Put `1000 GAS` deposit `value` along with the transaction as registration fee;
 4. Provide a `shareRate` ranges from `0` to `1000` in parameters, which is a distribution ratio in thousandths. It determines how many rewards of the total that voters can share, and can not be changed until the candidate exits;
 5. (optional) Withdraw past deposits if it has registered and exited before.
 
 If all conditions are met, the new candidate will be added to the candidate list. Only registered candidates can receive votes to be elected as a consensus node.
 
-A candidate can exit without any permission, but it requires 2 epochs to pass until the candidate is allowed to withdraw its registeration deposit. During this period, the candidate can't receive any votes or become a consensus node, but voters can revoke their votes and choose other candidates to share rewards.
+A candidate can exit without any permission, but it requires 2 epochs to pass until the candidate is allowed to withdraw its registeration deposit. During this period, the candidate can't receive any votes or become a consensus node, but voters can revoke their votes and choose other candidates to share rewards. As a prevention of malicious resources occupation, `5%` of the deposited registration fee will be charged by Governance when a candidate tries to exit and claim back.
 
 ### Election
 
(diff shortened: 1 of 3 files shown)


txhsl:
@dusmart

contracts/solidity/Governance.sol
```diff
@@ -115,7 +116,7 @@ contract Governance is IGovernance, ReentrancyGuard, GovProxyUpgradeable {
 
     function registerCandidate(uint shareRate) external payable {
         if (tx.origin != msg.sender) revert Errors.OnlyEOA();
-        if (msg.value < registerFee) revert Errors.InsufficientValue();
+        if (msg.value != registerFee) revert Errors.InsufficientValue();
```
AnnaShaleva, line 119: We need to adjust documentation accordingly.

contracts/solidity/Governance.sol
```diff
@@ -22,6 +22,7 @@ contract Governance is IGovernance, ReentrancyGuard, GovProxyUpgradeable {
     address public constant SYS_CALL =
         0xffffFFFfFFffffffffffffffFfFFFfffFFFfFFfE;
     uint public constant SCALE_FACTOR = 10 ** 18;
+    uint public constant EXIT_FEE_RATE = 5;
```
AnnaShaleva, line 25: And add a couple of words about exit fee to documentation, it's important for the users to know.

AnnaShaleva, line 25: 5% is a substantial amount. Is it ACKed by @steven1227?

txhsl: Yes, he's read the codes. We also want to improve the register fee as well, to 20,000 GAS, so that the exit fee will be 1,000 GAS.

I want to preserve the `balance` storage in case we implement some punishment or reward mechanism in future, so I use a rate here instead of a fixed fee.

contracts/solidity/Governance.sol
txhsl: Sure, will.

AnnaShaleva:
> Increase the register fee field in genesis file allocation.

@txhsl, I suppose it's a matter of #236. Could you please specify the desired storage item and its value?

contracts/solidity/Governance.sol
AnnaShaleva: >so I use a rate here instead of a fixed fee.

Rate is perfectly OK.

> We also want to improve the register fee as well, to 20,000 GAS,

OK, I see, well be fixed in #236.

contracts/solidity/Governance.sol
AnnaShaleva: I mean, not only the web-site documentation, but also `./contracts/README.md`.

contracts/solidity/Governance.sol
AnnaShaleva: Added in b3321187a5c62931389dc97f77ab84be1e45fc7f.

contracts/README.md
```diff
@@ -52,13 +52,13 @@ An EOA account is allowed to become a candidate only after successful registrati
 
 1. Registrant invokes `registerCandidate(uint shareRate)` of `0x1212000000000000000000000000000000000001` as message sender;
 2. Registrant is an EOA account and not yet a candidate;
-3. Put at least `1000 GAS` deposit `value` along with the transaction as registration fee;
+3. Put `20000 GAS` deposit `value` along with the transaction as registration fee;
```
AnnaShaleva: Belongs to #236, please, remove it.

AnnaShaleva (commented):
Otherwise LGTM.

contracts/README.md
txhsl: OK.

AnnaShaleva (approved):
I suppose it still would be nice to get some review from @dusmart. But otherwise LGTM.

txhsl:
> I suppose it still would be nice to get some review from @dusmart. But otherwise LGTM.

I've had a discussion with him, and got acknowledged, so it's OK.

PR #259, opened 2024-07-12
systemcontract: disable implementation contract initializer

### LINKED ISSUE #257: [suggestion] add `_disableInitializers` to every UUPS impl contract's constructor
# Rationale

The bridge contract is already using `initialize` for UUPS contract's initialisation and calling `_disableInitializers()` in the impl contract's constructor now. https://github.com/bane-labs/bridge-evm-contracts/pull/141

If we have to upgrade a contract one day, we <del>'ll have to</del> may change to use this pattern, too. I suggest that we introduce this into our contracts now so that the new contracts will also have these settings in their code.

# Implementation

See openzepplin's documents [here](https://docs.openzeppelin.com/contracts/4.x/api/proxy#UUPSUpgradeable).

> Avoid leaving a contract uninitialized.
>
> An uninitialized contract can be taken over by an attacker. This applies to both a proxy and its implementation contract, which may impact the proxy. To prevent the implementation contract from being used, you should invoke the [_disableInitializers](https://docs.openzeppelin.com/contracts/4.x/api/proxy#Initializable-_disableInitializers--) function in the constructor to automatically lock it when it is deployed:
> ```
> /// @custom:oz-upgrades-unsafe-allow constructor
> constructor() {
>     _disableInitializers();
> }
> ```

txhsl:
Close #257.

We don't have `initializer` or `reinitializer` thing in our system contracts, so there is no risk without this fix. This modification is just following the suggested coding manners from OpenZeppelin.

### DIFF
--- a/contracts/solidity/base/GovProxyUpgradeable.sol
+++ b/contracts/solidity/base/GovProxyUpgradeable.sol
@@ -13,6 +13,11 @@ abstract contract GovProxyUpgradeable is UUPSUpgradeable {
         _;
     }
 
+    /// @custom:oz-upgrades-unsafe-allow constructor
+    constructor() {
+        _disableInitializers();
+    }
+
     function _authorizeUpgrade(
         address newImplementation
     ) internal virtual override onlyAdmin {}
--- a/contracts/solidity/test/MockGovProxyUpgradeable.sol
+++ b/contracts/solidity/test/MockGovProxyUpgradeable.sol
@@ -1,0 +1,10 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.20;
+
+import {GovProxyUpgradeable} from "../base/GovProxyUpgradeable.sol";
+
+contract MockGovProxyUpgradeable is GovProxyUpgradeable {
+    function initialize() external initializer {}
+
+    function reinitialize() external reinitializer(1) {}
+}
--- a/contracts/test/GovProxyUpgradeable.ts
+++ b/contracts/test/GovProxyUpgradeable.ts
@@ -1,0 +1,28 @@
+import { ethers } from "hardhat";
+import { expect } from "chai";
+
+describe("GovProxyUpgradeable", function () {
+    it("Should prevent implementation contract from initialization", async function () {
+        const mockGovProxyUpgradeable = await ethers.deployContract("MockGovProxyUpgradeable");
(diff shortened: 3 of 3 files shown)


dusmart (approved):

config/genesis_testnet.json
```diff
@@ -212,15 +212,24 @@
...
...38503121561252c575f80fd5b612535836124d3565b9150602083013567ffffffffffffffff80821115612551575f80fd5b818501915085601f830112612564575f80fd5b81358181111561257657612576612507565b604051601f8201601f19908116603f0116810190838211818310171561259e5761259e612507565b816040528281528860208487010111156125b6575f80fd5b826020860160208301375f6020848301015280955050505050509250929050565b5f80604083850312156125e8575f80fd5b6125f1836124d3565b946020939093013593505050565b5f5b83811015612619578181015183820152602001612601565b50505f910152565b602081525f825180602084015261263f8160408501602087016125ff565b601f01601f19169190910160400192915050565b5f82516126648184602087016125ff565b9190910192915050565b5f6020828403121561267e575f80fd5b5051919050565b602080825282548282018190525f8481528281209092916040850190845b818110156124b05783546001600160a01b0316835260019384019392850192016126a3565b5f826126d6576126d6612423565b600160ff1b82145f19841416156126ef576126ef6123f8565b500590565b8181035f831280158383131683831282161715611fda57611fda6123f8565b5f600160ff1b8201612727576127276123f8565b505f190190565b634e487b7160e01b5f52603160045260245ffdfea26469706673582212203df8df4231f2b2d21fe02c1e1e7e932825e3edfc3651837059e089eb4fb6a78b64736f6c63430008190033"
```
AnnaShaleva: It's a problem. I don't think we need to update testnet configuration anymore, because we've already launched T4 with it. We'll have a separate configuration for mainnet, and there we can use this new implementation. But T4 have to live with the old one.

AnnaShaleva (commented):
Also, @txhsl there's no need to add labels/milestones to the PR if PR closes an existing issue. We should add labels/milestones to the issue itself (I did it).

config/genesis_testnet.json
txhsl: OK.

AnnaShaleva (commented):
Yep, this way LGTM.

PR #260, opened 2024-07-12
systemcontract: fix Governance transferVote

### LINKED ISSUE #258: invalid totalVotes during transferVote if candidateFrom is deactivated
See different votes updating logic here for `vote` and `revokeVote` as well as `transferVote`.

https://github.com/bane-labs/go-ethereum/blob/6049e3db040262041b099a0d8040937541f516e5/contracts/solidity/Governance.sol#L174-L177

https://github.com/bane-labs/go-ethereum/blob/6049e3db040262041b099a0d8040937541f516e5/contracts/solidity/Governance.sol#L195-L202

https://github.com/bane-labs/go-ethereum/blob/6049e3db040262041b099a0d8040937541f516e5/contracts/solidity/Governance.sol#L222-L227

It's obvious that `transferVote` is a combination of `revokeVote` and `vote`. But the logic above is sligtly different from the combination.

If the candidateFrom was not in candidateList, we should add the votedAmount of msg.sender to totalVotes.

My suggestion is to delete transferVote directly. Another alternative method is to just make `transferVote == revokeVote + vote`. Let's keep safety first priority and then consider the GAS cost.

txhsl:
Close #258.

### DIFF
--- a/contracts/solidity/Governance.sol
+++ b/contracts/solidity/Governance.sol
@@ -222,6 +222,9 @@ contract Governance is IGovernance, ReentrancyGuard, GovProxyUpgradeable {
         // update votes
         receivedVotes[candidateFrom] -= amount;
         receivedVotes[candidateTo] += amount;
+        if (!candidateList.contains(candidateFrom)) {
+            totalVotes += amount;
+        }
         votedTo[msg.sender] = candidateTo;
         voterGasPerVote[msg.sender] = candidateGasPerVote[candidateTo];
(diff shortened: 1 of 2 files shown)


dusmart (approved):

AnnaShaleva (approved):
@dusmart, a really good catch.

@txhsl, we should update T4 governance implementation via voting, but it's possible only after we register our non-stand-by candidates.

PR #263, opened 2024-07-15
internal/ethapi: set necessary fee price before gas estimation

### LINKED ISSUE #262: Avoid gas estimate errors caused by the fee policies of Neo X
# Rationale

There are many tools (including ethers.js, Hardhat, Remix) don't provide a valid `MaxPriorityFeePerGas` when calling the `eth_estimateGas` RPC service.

They work on other EVM chains, but since we applied a policy check in `preCheck`, their users may get errors as following.

```
Returned error: failed with 18750000 gas: transaction underpriced: address 0xbb03c5030cAC72E290Ae185A8b9b375C58f7A9a6, gasTipCap 0, gasFeeCap 26666666666, policy minGasTipCap 20000000000, baseFee 20000000000
```

# Implementation

We can request these tools to improve their codes by taking `eth_maxPriorityFeePerGas` into consideration, like https://github.com/NomicFoundation/hardhat/issues/5514, but I think we should improve our RPC handlers as well.

txhsl:
Close #262.

### DIFF
--- a/internal/ethapi/api.go
+++ b/internal/ethapi/api.go
@@ -1284,6 +1284,9 @@ func DoEstimateGas(ctx context.Context, b Backend, args TransactionArgs, blockNr
 // configuration (if non-zero).
 // Note: Required blob gas is not computed in this method.
 func (s *BlockChainAPI) EstimateGas(ctx context.Context, args TransactionArgs, blockNrOrHash *rpc.BlockNumberOrHash, overrides *StateOverride) (hexutil.Uint64, error) {
+	if err := args.setFeeMinAllowed(ctx, s.b); err != nil {
+		return 0, err
+	}
 	bNrOrHash := rpc.BlockNumberOrHashWithNumber(rpc.LatestBlockNumber)
 	if blockNrOrHash != nil {
 		bNrOrHash = *blockNrOrHash
--- a/internal/ethapi/transaction_args.go
+++ b/internal/ethapi/transaction_args.go
@@ -143,6 +143,32 @@ func (args *TransactionArgs) setDefaults(ctx context.Context, b Backend) error {
 	return nil
 }
 
+// setFeeMinAllowed fills in minimum allowed fee values for underpriced tx.
+func (args *TransactionArgs) setFeeMinAllowed(ctx context.Context, b Backend) error {
+	if head := b.CurrentHeader(); b.ChainConfig().IsNeoXBurn(head.Number, head.Time) {
+		if args.GasPrice != nil || args.MaxFeePerGas != nil || args.MaxPriorityFeePerGas != nil {
+			gastip, err := b.SuggestGasTipCap(ctx)
+			if err != nil {
+				return err
+			}
+			gasprice := new(big.Int).Add(gastip, head.BaseFee)
+			if args.GasPrice != nil {
+				if args.GasPrice.ToInt().Cmp(gasprice) < 0 {
+					args.GasPrice = (*hexutil.Big)(gasprice)
+				}
+			} else {
+				if args.MaxPriorityFeePerGas == nil || args.MaxPriorityFeePerGas.ToInt().Cmp(gastip) < 0 {
+					args.MaxPriorityFeePerGas = (*hexutil.Big)(gastip)
+				}
+				if args.MaxFeePerGas == nil || args.MaxFeePerGas.ToInt().Cmp(gasprice) < 0 {
+					args.MaxFeePerGas = (*hexutil.Big)(gasprice)
+				}
+			}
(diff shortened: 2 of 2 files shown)


txhsl:
By filling up the `MaxPriorityFeePerGas` field, we can let third-party tools to pass `minGasTipCap` check if they don't provide it.

But, let me take Remix as an example, it only provides a `MaxFeePerGas` which equals `4/3` of the latest base fee. Since we are going to burn 50% or less of the transaction fee, the `MaxFeePerGas` should be larger than `2 * baseFee`, otherwise the sanity check will fail. So the issue is not fully solved yet.

Is it correct to overwrite the `MaxFeePerGas` in estimation, or any other solutions?

internal/ethapi/api_test.go
```diff
@@ -671,19 +671,19 @@ func TestEstimateGas(t *testing.T) {
...
-			want:      21000,
+			expectErr: core.ErrInsufficientFundsForTransfer,
 		},
 		// empty create
 		{
 			blockNumber: rpc.LatestBlockNumber,
 			call:        TransactionArgs{},
-			expectErr:   nil,
-			want:        53000,
+			expectErr:   errors.New("gas required exceeds allowance (0)"),
```
AnnaShaleva: Do you know how many tools in the Ethereum ecosystem use this API with empty parameters list? I'm wondering because this is an incompatible change (it was supported before this fix, and now it's not).

AnnaShaleva (commented):
> Is it correct to overwrite the MaxFeePerGas in estimation, or any other solutions?

We may try to modify `args.setFeeDefaults` in order to make special conditions for NeoXBurn hardfork. But I'm not sure if MaxFeePerGas is safe to be modified, even only in case if it's insufficient. Because other tools may rely on their value, whereas backend silently modifies it for its own check, and if it's not a problem with Remix, then it may be a problem for some other tool.

txhsl:
> But I'm not sure if MaxFeePerGas is safe to be modified, even only in case if it's insufficient. Because other tools may rely on their value, whereas backend silently modifies it for its own check, and if it's not a problem with Remix, then it may be a problem for some other tool.

Well, let's have a look at documents. 

The code comment says.

> EstimateGas returns the lowest possible gas limit that allows the transaction to run successfully at block `blockNrOrHash`, or the latest block if `blockNrOrHash` is unspecified. It returns error if the transaction would revert or if there are unexpected failures. The returned value is capped by both `args.Gas` (if non-nil & non-zero) and the backend's RPCGasCap configuration (if non-zero).

The official document says.

> EstimateGas generates and returns an estimate of how much gas is necessary to allow the transaction to complete. The transaction will not be added to the blockchain.

And Alchemy notes.

> - `eth_estimateGas` will check the balance of the sender (to make sure that the sender has enough gas to complete the request). This means that even though the call doesn't consume any gas, the from address must have enough gas to execute the transaction. 
> - If no gas is specified geth uses the block gas limit from the pending block as an upper bound. As a result, the returned estimate might not be enough to execute the call/transaction when the amount of actual gas needed is higher than the pending block gas limit.

So, there are several points.

1. EstimateGas returns a gas limit value if the requested tx can succeed;
2. EstimateGas returns an error if the requested tx failed;
3. EstimateGas checks the balance of the sender and returns an error if the balance is not enough to pay for the requested tx.

I think, set a minimum price for empty request is compatible based on this description. Ideally, we should return a `ErrUnderpriced` error to Remix, but we have to make it successful.

`setFeeDefaults()` works good on filling in default fee values for **unspecified tx fields**, but we are going to do something different, which is filling in minimum allowed fee values for **underpriced tx**. So I'm going to add a new one.

internal/ethapi/api_test.go
txhsl: I think… there is no need to request like that. But I will check.

OK, it is a contract creation request, quite important, but the `from` address is empty.

eth/gasestimator/gasestimator.go
```diff
@@ -75,7 +75,7 @@ func Estimate(ctx context.Context, call *core.Message, opts *Options, gasCap uin
 
 		available := balance
 		if call.Value != nil {
-			if call.Value.Cmp(available) >= 0 {
+			if call.Value.Cmp(available) > 0 {
```
chenquanyu: If value and balance are equal, the fee cannot be paid. What is the purpose of removing the equality condition?

txhsl: For expected behaviour, I'm reverting this.

chenquanyu (approved):

AnnaShaleva (commented):
This way looks better.

PR #271, opened 2024-07-17
Add mainnet genesis configuration

AnnaShaleva:
Add genesis configuration for Neo X mainnet. @txhsl, @chenquanyu please review and ensure that contracts code is OK.

Can't be merged until all TODOs are resolved.

Close #64.

PR #275, opened 2024-07-19
systemcontract: change register fee and exit fee rate

txhsl:
Relate #247, #236 and #271.

Increase the exit fee rate from 5% to 50%, but decrease the register fee from 20000 GAS to 2000 GAS, which is suggested by @steven1227 for mainnet configuration.

NOTE: Genesis allocation is not updated here.

### DIFF
--- a/contracts/README.md
+++ b/contracts/README.md
@@ -68,13 +68,13 @@ An EOA account is allowed to become a candidate only after successful registrati
 
 1. Registrant invokes `registerCandidate(uint shareRate)` of `0x1212000000000000000000000000000000000001` as message sender;
 2. Registrant is an EOA account and not yet a candidate;
-3. Put `20000 GAS` deposit `value` along with the transaction as registration fee;
+3. Put `2000 GAS` (mainnet) or `20000 GAS` (testnet) deposit `value` along with the transaction as registration fee;
 4. Provide a `shareRate` ranges from `0` to `1000` in parameters, which is a distribution ratio in thousandths. It determines how many rewards of the total that voters can share, and can not be changed until the candidate exits;
 5. (optional) Withdraw past deposits if it has registered and exited before.
 
 If all conditions are met, the new candidate will be added to the candidate list. Only registered candidates can receive votes to be elected as a consensus node.
 
-A candidate can exit without any permission, but it requires 2 epochs to pass until the candidate is allowed to withdraw its registeration deposit. During this period, the candidate can't receive any votes or become a consensus node, but voters can revoke their votes and choose other candidates to share rewards. As a prevention of malicious resources occupation, `5%` of the deposited registration fee will be charged by Governance when a candidate tries to exit and claim back.
+A candidate can exit without any permission, but it requires 2 epochs to pass until the candidate is allowed to withdraw its registeration deposit. During this period, the candidate can't receive any votes or become a consensus node, but voters can revoke their votes and choose other candidates to share rewards. As a prevention of malicious resources occupation, `1000 GAS` will be charged by Governance when a candidate tries to exit and claim back.
 
 ### Election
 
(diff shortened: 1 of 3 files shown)


contracts/README.md
```diff
@@ -68,13 +68,13 @@ An EOA account is allowed to become a candidate only after successful registrati
 
 1. Registrant invokes `registerCandidate(uint shareRate)` of `0x1212000000000000000000000000000000000001` as message sender;
 2. Registrant is an EOA account and not yet a candidate;
-3. Put `20000 GAS` deposit `value` along with the transaction as registration fee;
+3. Put `2000 GAS` deposit `value` along with the transaction as registration fee;
```
AnnaShaleva: I think we need to differentiate two networks in the documentation. Keep `20000 GAS` and `5%` for T4 and add a note that for mainnet these values are different and equal to `2000 GAS` and `50%`.

AnnaShaleva (commented):
I think 50% is really a lot, this logic seems to be imbalanced a bit with this value. Do you have a reason of why the charge is so large?

txhsl:
> I think 50% is really a lot, this logic seems to be imbalanced a bit with this value. Do you have a reason of why the charge is so large?

Well, the code expression of 50% seems quite rude, so I changed it a bit. Let me describe the background first.

1. N3 charges 1000 GAS for registration, and it will not be refunded;
2. The first Neo X solution refunds all register fee back;
3. I'm not expected to change the registration logic after audit;
4. We need an exit fee there to prevent DoS attack about candidate limit, and we expect it to be the same amount of N3;
5. The 20000 GAS deposit is too much (comparing with 32 ETH on Ethereum), and market requires us to lower it;

How about now? I declare the exit fee as a fixed value.

contracts/README.md
txhsl: Differentiated.

roman-khimov (commented):
Is it that much different from N3 "pay 1000 GAS and never worry again)" model then?

contracts/solidity/Governance.sol
```diff
@@ -21,7 +21,7 @@ contract Governance is IGovernance, ReentrancyGuard, GovProxyUpgradeable {
     address public constant SYS_CALL =
         0xffffFFFfFFffffffffffffffFfFFFfffFFFfFFfE;
     uint public constant SCALE_FACTOR = 10 ** 18;
-    uint public constant EXIT_FEE_RATE = 5;
+    uint public constant EXIT_FEE = 1000 ether;
```
AnnaShaleva: It would be nice to make this value customizable in future (add setter), but we can update contract later if it's needed.

contracts/solidity/Governance.sol
```diff
@@ -143,8 +143,7 @@ contract Governance is IGovernance, ReentrancyGuard, GovProxyUpgradeable {
         ) revert Errors.CandidateWithdrawNotAllowed();
 
         // send back balance
-        uint amount = (candidateBalanceOf[msg.sender] * (100 - EXIT_FEE_RATE)) /
-            100;
+        uint amount = candidateBalanceOf[msg.sender] - EXIT_FEE;
```
AnnaShaleva: It wasn't possible to get negative amount with the old rate-based code, and now it's possible if we ever reduce candidate deposit to less than 1000.

AnnaShaleva (commented):
To be honest, the previous rate-based way was more preferable to me because fee is always proportional to the deposit which seems to be honest and protective from DoS for *every* deposit value. The question from my side was why 50%. But from your answer I see that it's a necessity, thus may be it's OK to keep 50%.

contracts/solidity/Governance.sol
txhsl: You're right.

txhsl:
> Is it that much different from N3 "pay 1000 GAS and never worry again)" model then?

To have the same thing as N3 “pay 1000 GAS and never worry again” model, I will have to change the register and exit logic here. But I don't expect that after audit, especially closely before mainnet.

PR #279, opened 2024-07-23
systemcontract: add multisig delegator and ut

### LINKED ISSUE #278: Add a vote delegator contract as the owner of bridge
# Rationale

@steven1227 requires a “multi-sig” contract to be the owner of our bridge and its management, instead of an EOA account at the beginning.

# Implementation

Add a vote contract, set it as the owner of bridge, so that it can delegate all management invocations and require a Governance vote.

We may update and remove this contract in future.

txhsl:
Close #278.


contracts/solidity/VoteDelegator.sol
```diff
@@ -0,0 +1,18 @@
...
+contract VoteDelegator is GovernanceVote, GovProxyUpgradeable {
+    function delegateVote(
+        address target,
+        bytes calldata data
+    )
+        external
+        needVote(keccak256(abi.encode(target)), keccak256(abi.encode(data)))
+        returns (bool success, bytes memory res)
+    {
+        (success, res) = target.call(data);
```
chenquanyu: This is a write method, I think we should revert when not success. Please refer to https://github.com/OpenZeppelin/openzeppelin-contracts/blob/9e73c4b58120c17135bf269945d0f148810bf7f1/contracts/utils/Address.sol#L127

txhsl:
@AnnaShaleva We can set its code in genesis file or update it on an empty Stub contract after launch, how do you think?

AnnaShaleva:
> We can set its code in genesis file or update it on an empty Stub contract after launch, how do you think?

For mainnet I think we'd better start right from the appropriate set of system contracts, including this multisignature contract. We can deploy this contract from genesis in a proper way (occupy the `0x1212000000000000000000000000000000000008` address), and I don't think we need to artificially complicate things for ourselves.

The only concern I have is: I'd like to keep the set of stub contract addresses consistent between mainnet and T4. So the question is: are we going to use the same multisig approach for T4? If yes, then implementation of Stub0 contract of T4 should be updated to this multisignature contract and the list of stub contracts for mainnet should be reduced by one. If no, then implementation of Stub0 contract of T4 should never be updated and the list of stub contracts for mainnet should be reduced by one. What do you think about it?

contracts/solidity/CommitteeMultiSig.sol
```diff
@@ -0,0 +1,42 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.25;
+
+import {GovernanceVote} from "./base/GovernanceVote.sol";
+import {ERC1967Utils, GovProxyUpgradeable} from "./base/GovProxyUpgradeable.sol";
+import {Address} from "@openzeppelin/contracts/utils/Address.sol";
+
+contract CommitteeMultiSig is GovernanceVote, GovProxyUpgradeable {
```
AnnaShaleva, line 8: I know we're in a hurry, but we need to make things properly. We need a documentation for this contract to be added to `contracts/README.md`, to adjust "System contract stubs" section of the same document and to add corresponding hash to the `core/systemcontracts/systemcontracts.go`.

AnnaShaleva (commented):
Let's decide on contract stubs management wrt this additional contract and then create a PR with mainnet genesis allocations update to #271.

contracts/solidity/CommitteeMultiSig.sol
txhsl: OK.

txhsl:
> So the question is: are we going to use the same multisig approach for T4?

We are, and @chenquanyu may do that soon.

> If yes, then implementation of Stub0 contract of T4 should be updated to this multisignature contract and the list of stub contracts for mainnet should be reduced by one.

The Stubs don't have real indexes, but I'm worrying about the implementation code at `0x1212100000000000000000000000000000000008`, we may have to move it to `0x1212100000000000000000000000000000000009` for mainnet. So what about using the `0x1212000000000000000000000000000000000007` position of DKG, actually it is now empty.

txhsl:
We have scheduled an upgrade to `0x1212000000000000000000000000000000000007` on T4, but not yet executed.

I've updated the document and codes in this PR, since `DKG` will not be there when we launch mainnet, I prefer to remove it first.

chenquanyu (approved):

PR #283, opened 2024-07-24
params: start v0.2.2 cycle

AnnaShaleva:
Start v0.2.2 that is aimed to be compatible with v0.2.1 released for MainNet launch. If there are some urgent fixes need to be applied at the top of v0.2.1 for MainNet launch, we can add them into this release. If not, then the next v0.3.0 cycle will be started for major anti-MEV protocol changes. I haven't created a designated milestone for v0.2.2 since there are no known bugs right now.

@txhsl, @chenquanyu, please consider this message.

### DIFF
--- a/params/version.go
+++ b/params/version.go
@@ -21,10 +21,10 @@ import (
 )
 
 const (
-	VersionMajor = 0        // Major version component of the current release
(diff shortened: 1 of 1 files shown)


roman-khimov (approved):
A new cycle needs to be started anyway.

PR #285, opened 2024-07-30
antimev: add `EncryptedTx` transaction type

chenquanyu:
`EncryptedTx` is exactly the same as `DynamicFeeTx` except for the type field. Transaction of this type will be encrypted into the data field of the normal GAS transfer transaction to the GovReward contract.

### DIFF
--- a/core/txpool/legacypool/legacypool.go
+++ b/core/txpool/legacypool/legacypool.go
@@ -291,7 +291,8 @@ func New(config Config, chain BlockChain) *LegacyPool {
 // pool, specifically, whether it is a Legacy, AccessList or Dynamic transaction.
 func (pool *LegacyPool) Filter(tx *types.Transaction) bool {
 	switch tx.Type() {
-	case types.LegacyTxType, types.AccessListTxType, types.DynamicFeeTxType:
+	// TODO: Will remove types.EncryptedTxType after tx type testing, EncryptedTx should not be in memory pool
+	case types.LegacyTxType, types.AccessListTxType, types.DynamicFeeTxType, types.EncryptedTxType:
 		return true
 	default:
 		return false
@@ -644,7 +645,8 @@ func (pool *LegacyPool) validateTxBasics(tx *types.Transaction, local bool) erro
 		Accept: 0 |
 			1<<types.LegacyTxType |
 			1<<types.AccessListTxType |
-			1<<types.DynamicFeeTxType,
+			1<<types.DynamicFeeTxType |
+			1<<types.EncryptedTxType, // TODO: Will remove types.EncryptedTxType after tx type testing, EncryptedTx should not be in memory pool
 		MaxSize: txMaxSize,
 		MinTip:  pool.gasTip.Load(),
 	}
--- a/core/txpool/legacypool/legacypool.go
+++ b/core/txpool/legacypool/legacypool.go
@@ -291,8 +291,7 @@ func New(config Config, chain BlockChain) *LegacyPool {
 // pool, specifically, whether it is a Legacy, AccessList or Dynamic transaction.
 func (pool *LegacyPool) Filter(tx *types.Transaction) bool {
 	switch tx.Type() {
-	// TODO: Will remove types.EncryptedTxType after tx type testing, EncryptedTx should not be in memory pool
-	case types.LegacyTxType, types.AccessListTxType, types.DynamicFeeTxType, types.EncryptedTxType:
+	case types.LegacyTxType, types.AccessListTxType, types.DynamicFeeTxType:
 		return true
 	default:
 		return false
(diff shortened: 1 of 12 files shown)


core/types/transaction.go
```diff
@@ -48,6 +48,8 @@ const (
 	AccessListTxType = 0x01
 	DynamicFeeTxType = 0x02
 	BlobTxType       = 0x03
+
+	EncryptedTxType = 0x7 // EncryptedTx is identical to DynamicFeeTx except for the type field.
```
AnnaShaleva: May be I'm paranoid, but let's move this type far away from Eth transaction types to avoid any possible conflicts in future. Some number that's less than `0x7f` will suit, may be `0x70`.

internal/ethapi/api.go
```diff
@@ -1451,7 +1451,7 @@ func newRPCTransaction(tx *types.Transaction, blockHash common.Hash, blockNumber
 		result.ChainID = (*hexutil.Big)(tx.ChainId())
 		result.YParity = &yparity
 
-	case types.DynamicFeeTxType:
+	case types.DynamicFeeTxType, types.EncryptedTxType:
```
AnnaShaleva, line 1454: Can we please adjust `TestTransaction_RoundTripRpcJSON` to test new type?

core/txpool/legacypool/legacypool.go
```diff
@@ -291,7 +291,8 @@ func New(config Config, chain BlockChain) *LegacyPool {
 // pool, specifically, whether it is a Legacy, AccessList or Dynamic transaction.
 func (pool *LegacyPool) Filter(tx *types.Transaction) bool {
 	switch tx.Type() {
-	case types.LegacyTxType, types.AccessListTxType, types.DynamicFeeTxType:
+	// TODO: Will remove types.EncryptedTxType after tx type testing, EncryptedTx should not be in memory pool
+	case types.LegacyTxType, types.AccessListTxType, types.DynamicFeeTxType, types.EncryptedTxType:
```
AnnaShaleva: Let's move forward with these TODOs/testing.

core/txpool/validation.go
```diff
@@ -65,6 +65,8 @@ func ValidateTransaction(tx *types.Transaction, head *types.Header, signer types
 	if !opts.Config.IsLondon(head.Number) && tx.Type() == types.DynamicFeeTxType {
 		return fmt.Errorf("%w: type %d rejected, pool not yet in London", core.ErrTxTypeNotSupported, tx.Type())
 	}
+	// TODO: add EncryptedTxType check
+
```
AnnaShaleva: Do we really need this check if `EncryptedTx` is not supposed to be included in the node's pool? I'd say that we'd better have checks at block verification/processing level.

core/types/tx_encrypted.go
```diff
@@ -0,0 +1,125 @@
...
+	"math/big"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/rlp"
+)
+
+// EncryptedTx represents an Encrypted neox transaction. Identical to DynamicFeeTx except for the txType() method
+type EncryptedTx struct {
+	ChainID    *big.Int
+	Nonce      uint64
```
AnnaShaleva: To be honest, I'm not in favor of code duplication between `DynamicFeeTx` and `EncryptedTx` because if anything is updated in the original Geth's version of `DynamicFeeTx`, we'll likely miss this update (unless we track it carefully every time).

Right now `EncryptedTx` doesn't have any difference between `DynamicFeeTx`, and I suppose it'll be the same in future. Thus, I'd suggest either to embed `DynamicFeeTx` inside `EncryptedTx` or to use a type alias. It allows to simplify `transaction.TxData` methods implementation and marshalling code.

AnnaShaleva (commented):
It would be nice to extend unit-tests a bit.

core/types/transaction.go
chenquanyu: I set this small value because `validateTxBasics` in the mempool limits the type range. We can change this value to a larger value after type testing.

core/txpool/validation.go
chenquanyu: Yes, will remove this todo.

core/types/tx_encrypted.go
chenquanyu: The only difference is `txType()` method, let me have a try to  embed `DynamicFeeTx`.

internal/ethapi/api.go
chenquanyu: Done.

core/txpool/legacypool/legacypool.go
chenquanyu: At present, it is more troublesome to construct new types of transactions because the front-end SDK has not yet been implemented. Encrypted types should not be able to enter the memory pool directly. We can test the actual process later.

core/types/tx_encrypted.go
chenquanyu: Done.

AnnaShaleva:
@chenquanyu I suppose that if we're not introducing new transaction type, then this PR should be closed?

chenquanyu:
After discussing with our front-end colleagues, we found that metamask does not support new types of transactions.

PR #286, opened 2024-08-01
antimev: modify priceHeap to prioritize encrypt wrapper tx

chenquanyu:
Modify `priceHeap` to prioritize encrypt wrapper tx.

### DIFF
--- a/core/systemcontracts/contracts.go
+++ b/core/systemcontracts/contracts.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/ethereum/go-ethereum/accounts/abi"
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/crypto"
 )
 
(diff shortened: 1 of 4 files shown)


core/systemcontracts/contracts.go
```diff
@@ -80,3 +81,9 @@ func GetBaseFeeStateHash() common.Hash {
 func GetBlackListStateHash(addr common.Address) common.Hash {
 	return crypto.Keccak256Hash(common.LeftPadBytes(addr.Bytes(), 32), common.LeftPadBytes([]byte{blackListSlotIndex}, 32))
 }
+
+// IsEncWrapperTx Checks whether a transaction is a wrapper transaction of an encrypted transaction
+func IsEncWrapperTx(tx *types.Transaction) bool {
+	// TODO: now we only check to address is reward contract, consider add gasPrice and data check in future
```
roman-khimov, line 87: For `IsEnc` it's likely sufficient, we shouldn't care much about `data` (unless there is some unencrypted part of it) and `gasPrice` should be checked elsewhere.

roman-khimov: BTW, some fee check can be important to prevent DOSing the pool with minimally priced transactions (that can evict nonencrypted ones). But that's not for `IsEnc` as well, it's a policy in its essense.

AnnaShaleva:
Not relevant anymore.

PR #287, opened 2024-08-02
Integrate amev dbft

### LINKED ISSUE #22: Import an anti-MEV feature based on threshold decryption
# Rationale

Recently, we are doing some researches on MEV (Miner Extractable Value) in DBFT. In Ethereum, for example, MEV is some potential value that miners/validators can earn by changing the transaction sequence in his block. For an Uniswap transaction, users always set a max/min amount limit for their input/output because of slippage. But miners/validators can listen and add another two transactions to force the user to deal at the limit, and take the slippage as earning. The economics about MEV is quite complicated, and the case I mentioned above is just one of existing MEV methods. 

In DBFT, MEV is an untouched area. Our consensus committee has a decisive power in validator MEV, but never used. So we have an plan to import some anti-MEV feature to Bane's DBFT, which is to set up an encrypted area in each block and only expose these transaction scripts after they are sequenced. 

# Implementation

Inspired by Radius (https://github.com/radiusxyz) and BLS consensus (https://github.com/neo-project/neo-modules/issues/637), we have a similar idea called Delayed Threshold Decryption. For example, in each round of DBFT, we can generate a public key and several pieces of a secret key based on a Decentralized Key Generation (DKG). The public key is exposed to everyone so that users can use it to encrypt their transactions before the next round of DBFT. When the next round comes, our consensus committee expose all pieces of the threshold secret key (or just share some partially decrypted information), decrypt and execute these transactions just after the block header is accepted by consensus, when nobody knows the transaction contents. 

The decryption process is clear. In a 3-stage DBFT, validators propose and vote to a new block header. Validators will only share the secret when commit if the header is acknowledged. Then a Verifiable Delay Function (VDF) is realized. However, the DKG related part is still under investigation. 

Recently, we are reading some research papers about DKG and TE in blockchain, like [*Practical Asynchronous Distributed Key Generation*](https://ieeexplore.ieee.org/abstract/document/9833584/) and [*1-Round Distributed Key Generation with Efficient Reconstruction Using Decentralized CP-ABE*](https://ieeexplore.ieee.org/abstract/document/9715067). We will keep on updating our progress in this issue if we have, and if you have any other idea about MEV protocol, let's discuss it as well.

roman-khimov:
1. New version of dBFT library (3.0+) containing anti-MEV extension switcher is integrated.
2. TPKE process is integrated into dBFT engine.
3. Marshalling for anti-MEV keystore is implemented as a temporary solution to be able to test Envelopes decryption.
4. No real DKG process is supported, it's a matter of other PRs. Current code relies on the fact that anti-MEV keystore contains data from completed DKG process.

Close #22.


consensus/dbft/dbft.go
```diff
@@ -356,11 +355,80 @@ func New(config *params.DBFTConfig, _ ethdb.Database) (*DBFT, error) {
...
+				}
+			}
+
+			// TODO: ensure that NewBlockFromContext is called only after _all_ PreBlock's transactions are
+			// fetched on the dBFT library side.
+			var txx []*types.Transaction
+			for _, tx := range pre.transactions {
+				if isEnvelope(tx) {
+					outer, inner := (Envelope)(tx).Decrypt(sharedKey)
+					txx = append(txx, outer, inner) // TODO: ensure the outer/inner order.
```
chenquanyu: We only keep one of the outer and inner with the same nonce, please refer to https://github.com/bane-labs/go-ethereum/pull/235#issuecomment-2230227119

AnnaShaleva:
I will revert crypto-related changes in this PR once #301 is merged. I just need them here for initial integration.

AnnaShaleva:
Not yet ready for review, there's a lot of dirty code, but at least decryption with static ready-to-use keystore works as expected in a good scenario:
```
2024-08-28T14:49:10.356+0300	DEBUG	dbft@v0.3.1-0.20240828111442-3b0a52287b5b/check.go:59	not enough PreCommits to create PreBlock	{"count": 3}
2024-08-28T14:49:10.356+0300	DEBUG	dbft@v0.3.1-0.20240828111442-3b0a52287b5b/dbft.go:229	received message	{"type": "PreCommit", "from": 0, "height": 3, "view": 0, "my_height": 3, "my_view": 0}
2024-08-28T14:49:10.356+0300	DEBUG	dbft@v0.3.1-0.20240828111442-3b0a52287b5b/dbft.go:229	received message	{"type": "PreCommit", "from": 0, "height": 3, "view": 0, "my_height": 3, "my_view": 0}
2024-08-28T14:49:10.356+0300	DEBUG	dbft@v0.3.1-0.20240828111442-3b0a52287b5b/dbft.go:229	received message	{"type": "PreCommit", "from": 1, "height": 3, "view": 0, "my_height": 3, "my_view": 0}
2024-08-28T14:49:10.356+0300	INFO	dbft@v0.3.1-0.20240828111442-3b0a52287b5b/dbft.go:532	received PreCommit	{"validator": 1}
2024-08-28T14:49:10.356+0300	DEBUG	dbft@v0.3.1-0.20240828111442-3b0a52287b5b/check.go:59	not enough PreCommits to create PreBlock	{"count": 4}
2024-08-28T14:49:10.356+0300	DEBUG	dbft@v0.3.1-0.20240828111442-3b0a52287b5b/dbft.go:229	received message	{"type": "PreCommit", "from": 0, "height": 3, "view": 0, "my_height": 3, "my_view": 0}
2024-08-28T14:49:10.356+0300	DEBUG	dbft@v0.3.1-0.20240828111442-3b0a52287b5b/dbft.go:229	received message	{"type": "PreCommit", "from": 6, "height": 3, "view": 0, "my_height": 3, "my_view": 0}
2024-08-28T14:49:10.356+0300	DEBUG	dbft@v0.3.1-0.20240828111442-3b0a52287b5b/dbft.go:229	received message	{"type": "PreCommit", "from": 6, "height": 3, "view": 0, "my_height": 3, "my_view": 0}
2024-08-28T14:49:10.359+0300	DEBUG	dbft@v0.3.1-0.20240828111442-3b0a52287b5b/dbft.go:229	received message	{"type": "PreCommit", "from": 1, "height": 3, "view": 0, "my_height": 3, "my_view": 0}
2024-08-28T14:49:10.361+0300	DEBUG	dbft@v0.3.1-0.20240828111442-3b0a52287b5b/dbft.go:229	received message	{"type": "PreCommit", "from": 4, "height": 3, "view": 0, "my_height": 3, "my_view": 0}
2024-08-28T14:49:10.361+0300	INFO	dbft@v0.3.1-0.20240828111442-3b0a52287b5b/dbft.go:532	received PreCommit	{"validator": 4}
2024-08-28T14:49:10.361+0300	INFO	dbft@v0.3.1-0.20240828111442-3b0a52287b5b/check.go:65	processing PreBlock	{"height": 3, "view": 0, "tx_count": 1}
INFO [08-28|14:49:10.361] decoding shares                          n=1
INFO [08-28|14:49:10.362] decoding shares                          n=1
INFO [08-28|14:49:10.362] decoding shares                          n=1
INFO [08-28|14:49:10.363] decoding shares                          n=1
INFO [08-28|14:49:10.364] decoding shares                          n=1
INFO [08-28|14:49:10.367] ####     Envelope found
INFO [08-28|14:49:10.370] aggregating shares                       len(shares)=5 len(encryptedKeys)=1 len(encryptedMsgs)=1
INFO [08-28|14:49:10.370] shares of                                "validator index +1"=1 num=1 val=[0xc011bee420]
INFO [08-28|14:49:10.370] shares of                                "validator index +1"=2 num=1 val=[0xc011bee430]
INFO [08-28|14:49:10.370] shares of                                "validator index +1"=3 num=1 val=[0xc011bee440]
INFO [08-28|14:49:10.370] shares of                                "validator index +1"=5 num=1 val=[0xc011bee450]
INFO [08-28|14:49:10.370] shares of                                "validator index +1"=7 num=1 val=[0xc011bee460]

// Here's the encrypted data from `TestTPKE` that I reused as an Envelope's data. It was successfully decrypted by CNs:

INFO [08-28|14:49:10.376] #######   Decrypted data                 data="pizza pizza pizza pizza pizza pizza pizza pizza pizza pizza pizza pizza pizza"


2024-08-28T14:49:10.377+0300	INFO	dbft@v0.3.1-0.20240828111442-3b0a52287b5b/send.go:150	sending Commit	{"height": 3, "view": 0}
2024-08-28T14:49:10.377+0300	DEBUG	dbft@v0.3.1-0.20240828111442-3b0a52287b5b/send.go:8	broadcasting message	{"type": "Commit", "height": 3, "view": 0}
2024-08-28T14:49:10.377+0300	DEBUG	dbft@v0.3.1-0.20240828111442-3b0a52287b5b/dbft.go:679	reset timer	{"h": 3, "v": 0, "delay": "10s"}
2024-08-28T14:49:10.377+0300	DEBUG	dbft@v0.3.1-0.20240828111442-3b0a52287b5b/check.go:104	not enough to commit	{"count": 1}
2024-08-28T14:49:10.377+0300	DEBUG	dbft@v0.3.1-0.20240828111442-3b0a52287b5b/dbft.go:229	received message	{"type": "PreCommit", "from": 5, "height": 3, "view": 0, "my_height": 3, "my_view": 0}
2024-08-28T14:49:10.377+0300	INFO	dbft@v0.3.1-0.20240828111442-3b0a52287b5b/dbft.go:532	received PreCommit	{"validator": 5}
2024-08-28T14:49:10.377+0300	INFO	dbft@v0.3.1-0.20240828111442-3b0a52287b5b/check.go:65	processing PreBlock	{"height": 3, "view": 0, "tx_count": 1}
2024-08-28T14:49:10.377+0300	INFO	dbft@v0.3.1-0.20240828111442-3b0a52287b5b/send.go:150	sending Commit	{"height": 3, "view": 0}
2024-08-28T14:49:10.377+0300	DEBUG	dbft@v0.3.1-0.20240828111442-3b0a52287b5b/send.go:8	broadcasting message	{"type": "Commit", "height": 3, "view": 0}
2024-08-28T14:49:10.378+0300	DEBUG	dbft@v0.3.1-0.20240828111442-3b0a52287b5b/dbft.go:679	reset timer	{"h": 3, "v": 0, "delay": "10s"}
2024-08-28T14:49:10.378+0300	DEBUG	dbft@v0.3.1-0.20240828111442-3b0a52287b5b/check.go:104	not enough to commit	{"count": 1}
2024-08-28T14:49:10.378+0300	DEBUG	dbft@v0.3.1-0.20240828111442-3b0a52287b5b/dbft.go:229	received message	{"type": "PreCommit", "from": 5, "height": 3, "view": 0, "my_height": 3, "my_view": 0}
2024-08-28T14:49:10.378+0300	DEBUG	dbft@v0.3.1-0.20240828111442-3b0a52287b5b/dbft.go:229	received message	{"type": "PreCommit", "from": 3, "height": 3, "view": 0, "my_height": 3, "my_view": 0}
2024-08-28T14:49:10.378+0300	INFO	dbft@v0.3.1-0.20240828111442-3b0a52287b5b/dbft.go:532	received PreCommit	{"validator": 3}
2024-08-28T14:49:10.378+0300	INFO	dbft@v0.3.1-0.20240828111442-3b0a52287b5b/check.go:65	processing PreBlock	{"height": 3, "view": 0, "tx_count": 1}
2024-08-28T14:49:10.378+0300	INFO	dbft@v0.3.1-0.20240828111442-3b0a52287b5b/send.go:150	sending Commit	{"height": 3, "view": 0}
2024-08-28T14:49:10.378+0300	DEBUG	dbft@v0.3.1-0.20240828111442-3b0a52287b5b/send.go:8	broadcasting message	{"type": "Commit", "height": 3, "view": 0}
2024-08-28T14:49:10.379+0300	DEBUG	dbft@v0.3.1-0.20240828111442-3b0a52287b5b/dbft.go:679	reset timer	{"h": 3, "v": 0, "delay": "10s"}
2024-08-28T14:49:10.379+0300	DEBUG	dbft@v0.3.1-0.20240828111442-3b0a52287b5b/check.go:104	not enough to commit	{"count": 1}
2024-08-28T14:49:10.379+0300	DEBUG	dbft@v0.3.1-0.20240828111442-3b0a52287b5b/dbft.go:229	received message	{"type": "PreCommit", "from": 5, "height": 3, "view": 0, "my_height": 3, "my_view": 0}
2024-08-28T14:49:10.379+0300	DEBUG	dbft@v0.3.1-0.20240828111442-3b0a52287b5b/dbft.go:229	received message	{"type": "Commit", "from": 4, "height": 3, "view": 0, "my_height": 3, "my_view": 0}
2024-08-28T14:49:10.379+0300	INFO	dbft@v0.3.1-0.20240828111442-3b0a52287b5b/dbft.go:573	received Commit	{"validator": 4}
2024-08-28T14:49:10.379+0300	DEBUG	dbft@v0.3.1-0.20240828111442-3b0a52287b5b/check.go:104	not enough to commit	{"count": 2}
2024-08-28T14:49:10.379+0300	DEBUG	dbft@v0.3.1-0.20240828111442-3b0a52287b5b/dbft.go:229	received message	{"type": "Commit", "from": 1, "height": 3, "view": 0, "my_height": 3, "my_view": 0}
2024-08-28T14:49:10.379+0300	INFO	dbft@v0.3.1-0.20240828111442-3b0a52287b5b/dbft.go:573	received Commit	{"validator": 1}
2024-08-28T14:49:10.379+0300	DEBUG	dbft@v0.3.1-0.20240828111442-3b0a52287b5b/check.go:104	not enough to commit	{"count": 3}
2024-08-28T14:49:10.379+0300	DEBUG	dbft@v0.3.1-0.20240828111442-3b0a52287b5b/dbft.go:229	received message	{"type": "Commit", "from": 3, "height": 3, "view": 0, "my_height": 3, "my_view": 0}
2024-08-28T14:49:10.379+0300	INFO	dbft@v0.3.1-0.20240828111442-3b0a52287b5b/dbft.go:573	received Commit	{"validator": 3}
2024-08-28T14:49:10.379+0300	DEBUG	dbft@v0.3.1-0.20240828111442-3b0a52287b5b/check.go:104	not enough to commit	{"count": 4}
2024-08-28T14:49:10.379+0300	DEBUG	dbft@v0.3.1-0.20240828111442-3b0a52287b5b/dbft.go:229	received message	{"type": "Commit", "from": 3, "height": 3, "view": 0, "my_height": 3, "my_view": 0}
2024-08-28T14:49:10.379+0300	DEBUG	dbft@v0.3.1-0.20240828111442-3b0a52287b5b/dbft.go:229	received message	{"type": "Commit", "from": 0, "height": 3, "view": 0, "my_height": 3, "my_view": 0}
2024-08-28T14:49:10.379+0300	INFO	dbft@v0.3.1-0.20240828111442-3b0a52287b5b/dbft.go:573	received Commit	{"validator": 0}
2024-08-28T14:49:10.380+0300	INFO	dbft@v0.3.1-0.20240828111442-3b0a52287b5b/check.go:113	approving block	{"height": 3, "hash": "0x7ba0b0f6bbc15397408d735f5e78a001f56aaf9c5512e646bdf41489963c89da", "tx_count": 1, "merkle": "0xa5798aa1f76ad0f526d9d61fdad8893ffe40ba4ac6b94a6ab8fa0d7dc4e2f8cc", "prev": "0x884f073532ade37ce7336b303296d73cc896d88d30428d089d6e13757e74101b"}
INFO [08-28|14:49:10.434] Chain reorg detected                     number=1 hash=ab76c6..d9e54d drop=1 dropfrom=c879c1..93e841 add=2 addfrom=a47846..d423e8
INFO [08-28|14:49:10.438] Successfully wrote new block with state  number=3 hash=a47846..d423e8
INFO [08-28|14:49:10.438] New block in the chain                   "dbft index"=3 "chain index"=3 hash=0xa47846d877ad38ba853fa76e6e21b2dfe88435e4b5207421c1a1513499d423e8 "parent hash"=0x884f073532ade37ce7336b303296d73cc896d88d30428d089d6e13757e74101b primary=3 coinbase=0x1212000000000000000000000000000000000003 "mix digest"=0x3e48bc8f2eb502267b1eae6bded2ece31a29829cdf94afb44f37688702643e43
INFO [08-28|14:49:10.438] Fetching latest sealing proposal         "desired number"=4
INFO [08-28|14:49:10.440] Commit new sealing work                  number=4 sealhash=cbeff6..4d22f0 "parent hash"=a47846..d423e8 etherbase=0x1212000000000000000000000000000000000003 txs=0 gas=0     fees=0      elapsed=1.313ms
```

consensus/dbft/dbft.go
```diff
@@ -623,6 +551,7 @@ func New(config *params.DBFTConfig, _ ethdb.Database) (*DBFT, error) {
 		}),
 		dbft.WithNewRecoveryMessage[common.Hash](func() dbft.RecoveryMessage[common.Hash] { return new(recoveryMessage) }),
 		dbft.WithVerifyPrepareResponse[common.Hash](func(_ dbft.ConsensusPayload[common.Hash]) error { return nil }),
+		dbft.WithVerifyPreCommit[common.Hash](func(preCommit dbft.ConsensusPayload[common.Hash]) error { return nil }),
```
vncoelho, line 554: I think that the name `PreCommit` is not correct following the previous specification.
Because we still `Commit` to the block, we just did not processed the decrypted transactions.

consensus/dbft/amev.go
```diff
@@ -0,0 +1,98 @@
...
+
+// decodeEnvelopesData finds Envelope transactions in the provided list and returns
+// their data in deserialized form.
+func decodeEnvelopesData(txs []*types.Transaction) []envelopeData {
+	var res []envelopeData
+	for i, tx := range txs {
+		if isEnvelope(tx) {
+			d, err := decodeEnvelopeData(tx.Data())
+			if err != nil {
+				// TODO: Envelope content unmarshalling may be checked during PrepareRequest level
```
vncoelho: Why is there this PreCommit?
Before, in the discussed designed we had a `Commit`as nowadays, exposing the keys. We just had a PreBlock, that would later be processed into a final ethereum block.

consensus/dbft/dbft.go
AnnaShaleva: Naming can always be improved, and we choose `PreCommit` because it's a phase that preceds the *final* block's Commit with all transactions decrypted. Do you have a suggestion on this message and phase name?

consensus/dbft/amev.go
AnnaShaleva: `PreCommit` message carries a decryption share generated by CN. We can't live without this extra message and extra phase because encrypted transactions decryption is possible only with these extra data.

vncoelho: That is exactly the Commit phase: "message carries a decryption share generated by CN."

This is a STRONG decision of each CN that can not be reverted, this phase is a Commit phase, not PreCommit.
After this the PreBlock is finalized, we just need to process it.

This name needs to be moved back to Commit and the next phase a PreBlock. There is no PreCommit

consensus/dbft/dbft.go
vncoelho: I commented below.

AnnaShaleva:
@roman-khimov @txhsl @chenquanyu @vncoelho @songb2  I did what I could, this PR works and ready for review and testing. It may be tested on 7-nodes privnet (I added ready-to-use keystores to this privnet and adjusted Makefile), an example of how to send an encrypted data is included into README.

There are still a set of TODOs that were moved to separate issues, but this PR contains the most significant part of of TPKE process integration with dBFT. I expect no more dBFT library modifications. Please, review carefully.

@songb2 except the "good" testing cases, the following testing scenario should be considered:
* `F` malicious nodes sending `PreCommit` messages with malformed shares data. In this case the rest `M` nodes should wait for `M` "good" shares trying to decrypt encrypted messages with every incoming PreCommit.

@txhsl from you I'd ask to review carefully the part where `AggregateAndDecryptWithShare` is used (`WithProcessPreBlock` callback) and other usages of TPKE-related crypto. Your comments were very helpful in understanding, hope I managed to integrate TPKE properly.

consensus/dbft/amev.go
AnnaShaleva: > This is a STRONG decision of each CN that can not be reverted

The behaviour of PreCommit matches this definition. If the node sends PreCommit, then there's no way to perform change view for this node.

> After this the PreBlock is finalized, we just need to process it.

We can't, because we need one additional phase during which nodes will collect signatures for the final block.

vncoelho: Yes, just change the name in this case.

Add another name after Commit.

PrepareRequest
PrepareResponse
Commit
ANYNAME
Block

AnnaShaleva: We used `CommitAck` for that `ANYNAME` in our TLA+ model (https://github.com/nspcc-dev/dbft/tree/master/formal-models#basic-mev-resistant-dbft-model), but when I started library integration it was more convenient to introduce `PreCommit` because there's almost no logic adjustment for the `Commit` then. So is it really important how we call this message in the code?

@roman-khimov, what do you thing?

vncoelho: PreCommit is not correct name, just change this please.
The "Commit" is a important name for the phase where is named as "PreCommit".
The name preceding "Commit" I do not care much....aheuahea

I suggest PreBlock Phase

roman-khimov: It's implemented as PreCommit/Commit and I don't see any real reason to change it. The semantics is the same as Commit/CommitAck or Commit/Anyname. But we had enough reasons to use Commit as the last step, it greatly simplifies keeping dbft library compatible with _both_ ways of operation.

consensus/dbft/dbft.go
```diff
@@ -679,6 +734,7 @@ func (c *DBFT) newBlockFromContext(sealingProposal *types.Header) *Block {
 	}
 	return res
 }
+
```
roman-khimov, line 808: :eyes:

consensus/dbft/dbft.go
```diff
@@ -569,50 +631,41 @@ func New(config *params.DBFTConfig, _ ethdb.Database) (*DBFT, error) {
...
+				log.Warn("proposed PreBlock has small timestamp",
 					"ts", dbftBlock.header.Time,
 					"last", c.lastTimestamp)
 				return false
 			}
+
+			// TODO: consider an old-way verification for those blocks that do not
+			// contain enveloped transactions? I.e. verify state-dependent fields in
+			// this callback if there's no encrypted transactions in the proposed
+			// block.
```
roman-khimov: Doesn't seem to be useful if we have an additional phase anyway and need to communicate once more. Better have more consistent flow.

consensus/dbft/dbft.go
```diff
@@ -359,18 +365,64 @@ func New(config *params.DBFTConfig, _ ethdb.Database) (*DBFT, error) {
...
+				encryptedMsgs[i] = encryptedTxs[i].encryptedMsg
+			}
+			c.lock.RLock()
+			ks := c.amevKeystore
+			c.lock.RUnlock()
+
+			var decryptFailed bool
+			decryptedTxsBytes, err := ks.AggregateAndDecryptWithShare(encryptedKeys, encryptedMsgs, shares)
+			if err != nil {
+				// We're toast, accept Envelopes instead of decrypted transactions.
```
roman-khimov: Why isn't it per-msg? Some messages can be OK and some broken, I'd rather have proper ones decrypted.

crypto/tpke/pvss.go
```diff
@@ -13,6 +15,59 @@ type PVSS struct {
...
 }
 
+var (
+	_ rlp.Encoder = &PVSS{}
+	_ rlp.Decoder = &PVSS{}
+)
+
+type pvssAux struct {
+	Commitment *Commitment // The commitment of local secret polynomial
+	R1         [48]byte    // The commitment of of random r
```
roman-khimov: of of

crypto/tpke/private_key.go
```diff
@@ -1,15 +1,42 @@
...
+var (
+	_ rlp.Encoder = &PrivateKey{}
+	_ rlp.Decoder = &PrivateKey{}
+)
+
+type privateKeyAux struct {
+	Fr []byte
+}
+
+func (priv *PrivateKey) EncodeRLP(w io.Writer) error {
```
roman-khimov, line 26: Do we serialize it anywhere?

go.mod
```diff
@@ -33,8 +33,8 @@ require (
 	github.com/golang/protobuf v1.5.3
 	github.com/golang/snappy v0.0.5-0.20220116011046-fa5810519dcb
 	github.com/google/gofuzz v1.2.0
-	github.com/google/uuid v1.3.0
-	github.com/gorilla/websocket v1.4.2
+	github.com/google/uuid v1.6.0
+	github.com/gorilla/websocket v1.5.1
```
roman-khimov: Why in 7d9d9b39c204e02320c53d799656a2038d2f729a?

go.mod
```diff
@@ -30,7 +30,7 @@ require (
 	github.com/gballet/go-verkle v0.1.1-0.20231031103413-a67434b50f46
 	github.com/gofrs/flock v0.8.1
 	github.com/golang-jwt/jwt/v4 v4.5.0
-	github.com/golang/protobuf v1.5.3
+	github.com/golang/protobuf v1.5.4
```
roman-khimov: Why in 704b4b0dc831793fdc8fa4bd40cbab659fd253d1?

core/txpool/legacypool/legacypool.go
```diff
@@ -2011,3 +2016,33 @@ func (t *lookup) RemotesBelowTip(threshold *big.Int) types.Transactions {
 func numSlots(tx *types.Transaction) int {
 	return int((tx.Size() + txSlotSize - 1) / txSlotSize)
 }
+
+// ValidateDecryptedTx checks the validity of the transaction to determine whether the outer envelope transaction should be replaced.
+func (p *LegacyPool) ValidateDecryptedTx(decryptedTx *types.Transaction, envelope *types.Transaction) error {
```
roman-khimov, line 2021: How about GAS limits? Block is limited and we know envelopes are OK with this limit, but if decrypted transactions request more we can get into trouble.

consensus/dbft/preblock.go
```diff
@@ -88,11 +73,25 @@ func (b *PreBlock) Transactions() []dbft.Transaction[common.Hash] {
 // SetTransactions implements [dbft.PreBlock] interface. txx may contain encrypted
 // Envelope transactions.
 func (b *PreBlock) SetTransactions(txx []dbft.Transaction[common.Hash]) {
-	txs := make([]*types.Transaction, len(txx))
+	var (
+		txs       = make([]*types.Transaction, len(txx))
+		envelopes []envelopeData // don't allocate, Envelopes supposed to be rear.
```
roman-khimov: rare?

crypto/tpke/pvss.go
```diff
@@ -13,6 +15,59 @@ type PVSS struct {
...
+	R2         [96]byte    // The verifiable commitment of of r1
+	Bigf       [][48]byte  // The commitment of secret sharing
+}
+
+func (p *PVSS) EncodeRLP(w io.Writer) error {
+	f := make([][48]byte, len(p.bigf))
```
roman-khimov: Don't we have this magic 48 defined somewhere?

consensus/dbft/preblock.go
```diff
@@ -0,0 +1,103 @@
...
+	for i, tx := range txx {
+		txs[i] = tx.(*Transaction).Tx
+		if isEnvelope(txs[i]) {
+			d, err := decodeEnvelopeData(txs[i].Data())
+			if err != nil {
+				// Not an Envelope in fact since it contains malformed data. Include
+				// it as a simple transaction.
+				continue
+			}
+			d.index = i
```
roman-khimov, line 89: Feels like this index can be avoided, although it's not critical at this stage.

consensus/dbft/dbft.go
txhsl: `AggregateAndDecryptWithShare` only returns an error when the decryption share inputs are illegal. If the encrypted messages are bad, their output will be replaced with `nil`.

I prefer to handle these messages together, that makes the computation and verification simple and clear.

consensus/dbft/dbft.go
```diff
@@ -626,6 +670,103 @@ func New(config *params.DBFTConfig, _ ethdb.Database) (*DBFT, error) {
...
+				encryptedKeys[i] = pre.envelopesData[i].encryptedKey
+				encryptedMsgs[i] = pre.envelopesData[i].encryptedMsg
+			}
+			c.lock.RLock()
+			ks := c.amevKeystore
+			c.lock.RUnlock()
+
+			decryptedTxsBytes, err := ks.AggregateAndDecryptWithShare(encryptedKeys, encryptedMsgs, shares)
+			if err != nil {
+				// One of shares is invalid, wait for more shares to be collected.
```
txhsl: I just want to clarify a bit here if I haven't. The case that `AggregateAndDecryptWithShare` returns an error doesn't mean that "one of shares is invalid" but "valid shares are not enough".

If the crypto threshold is `5`, there are `6` shares and `5` of them are valid, then `AggregateAndDecryptWithShare` will give the result.

consensus/dbft/dbft.go
```diff
@@ -626,6 +670,103 @@ func New(config *params.DBFTConfig, _ ethdb.Database) (*DBFT, error) {
...
+				return fmt.Errorf("failed to decrypt Encrypted transactions: %w", err)
+			}
+
+			if len(decryptedTxsBytes) != len(pre.envelopesData) {
```
txhsl, line 721: This case is not going to happen, but a verification is not bad.

consensus/dbft/dbft.go
```diff
@@ -626,6 +670,103 @@ func New(config *params.DBFTConfig, _ ethdb.Database) (*DBFT, error) {
...
+				// One of shares is invalid, wait for more shares to be collected.
+				return fmt.Errorf("invalid number of Decrypted transactions: expected %d, actual %d", len(pre.envelopesData), len(decryptedTxsBytes))
+			}
+			var (
+				txx = make([]*types.Transaction, len(pre.transactions))
+				j   int
+			)
+			for i := range pre.transactions {
+				if pre.envelopesData[j].index != i || // pre.transactions[i] is not an envelope, use it as-is.
```
txhsl, line 730: A bit of confused here, are we having an index for `pre.envelopesData`, because it can be anywhere of the list, even between two normal transactions?

But we may have #286, so that envelopes are always sorted at the beginning of the list. If so, then we can remove the index, since it is a position-to-position replacement.

crypto/bls12381/g1.go
```diff
@@ -18,15 +18,47 @@ package bls12381
...
+
+	"github.com/ethereum/go-ethereum/rlp"
 )
 
 // PointG1 is type for point in G1.
 // PointG1 is both used for Affine and Jacobian point representation.
 // If z is equal to one the point is considered as in affine form.
 type PointG1 [3]fe
 
+var (
```
txhsl, line 33: Codes here and in `crypto/ecies` and `crypto/tpke` seem something has to be removed in future, #310 related.

All of them are just used for persistence, correct?

consensus/dbft/amev.go
vncoelho: There is no sense in using PreCommit name, because the Commit is done in this stage.
There is no way back.

It is a misleading name.

roman-khimov: Take it as a "preliminary commit". It's a commit. But it's a preliminary one, not final, since we're committing to something that is not yet a final block.

vncoelho: No, the block is the same, the next phase is informative just.
The name preliminary commit is also wrong.
It is a Commit, that is it.

The next phase could be removed, but then every node would need to process the final block in the ethereum standard.

The next phase does not make decision, it just processes and publishes a well defined block.

roman-khimov: > the block is the same,

Transactions are different. Side-effects are different.

> The next phase does not make decision

Yep. We've (preliminary) committed to it and the last Commit just finalizes the thing.

consensus/dbft/dbft.go
AnnaShaleva: Fixed!

consensus/dbft/dbft.go
AnnaShaleva: Good, TODO is removed.

consensus/dbft/dbft.go
AnnaShaleva: > If the encrypted messages are bad, their output will be replaced with nil.

Yes, below we have a `nil`-check for decrypted messages, and if the message is nil then it's replaced with the original Envelope.

> Why isn't it per-msg? 

So it actually is per-message, see also 829ed21f8480bc31b140c5eb6d512cb8f5f8d7fa. If `AggregateAndDecryptWithShare` returns an error then it's not enough valid shares and we have to wait for more valid PreCommits to be collected.

crypto/tpke/pvss.go
AnnaShaleva: Fixed.

crypto/tpke/private_key.go
AnnaShaleva: Yes, but just for testing purposes. I know we can't serialize raw private key, but right now we don't have any storage for anti-MEV keystore, it will be implemented as a part of https://github.com/bane-labs/go-ethereum/issues/310#issuecomment-2328518917.

AnnaShaleva: So almost all RLP serialization code should be removed in future.

go.mod
AnnaShaleva: Fixed, it's rebase artifact.

go.mod
AnnaShaleva: Fixed.

core/txpool/legacypool/legacypool.go
AnnaShaleva: This code is a part of https://github.com/bane-labs/go-ethereum/pull/307, I haven't touched it. It's not the only thing that is wrong regarding decrypted transactions verification, I created https://github.com/bane-labs/go-ethereum/issues/314 for that.

consensus/dbft/preblock.go
AnnaShaleva: Yep.

crypto/tpke/pvss.go
AnnaShaleva: Sure we do, fixed. Although this code is a garbage that will be removed.

consensus/dbft/preblock.go
AnnaShaleva: It can, but I created it as an optimisation in order to avoid excessive calls to `isEnvelope`. Also, this code assumes that Envelopes may be mixed with simple transactions.

consensus/dbft/dbft.go
AnnaShaleva: Yes, thanks for clarification, you're clear. It's just bad-formulated comment, the code works as you described. I fixed this comment.

consensus/dbft/dbft.go
AnnaShaleva: Let's keep it just in case.

consensus/dbft/dbft.go
AnnaShaleva: > because it can be anywhere of the list, even between two normal transactions?

Yes, I thought we're not going to prioritize Envelopes. @roman-khimov, what do you think?

Anyway, I'd suggest to merge the existing version firstly and then remove this excessive indexes after #286 is merged.

AnnaShaleva: Let's have https://github.com/bane-labs/go-ethereum/issues/317 for further discussion.

crypto/bls12381/g1.go
AnnaShaleva: Yes, exactly. I've added https://github.com/bane-labs/go-ethereum/issues/310#issuecomment-2328518917 in order not to forget it in future.

They are added because right now we have to test Envelopes decryption on privnet, so currently dBFT engine relies on RLP-serialized keystore with completed DKG process.

consensus/dbft/dbft.go
```diff
@@ -626,6 +669,103 @@ func New(config *params.DBFTConfig, _ ethdb.Database) (*DBFT, error) {
 				log.Warn("can't broadcast consensus message", "error", err)
 			}
 		}),
+		dbft.WithAntiMEVExtensionEnablingHeight[common.Hash](0),
```
chenquanyu, line 672: Maybe we need a genesis config of antimev switch height/time.

consensus/dbft/dbft.go
roman-khimov: I think this code better not make any assumptions about ordering, it should work as 1:1 mapper of envelopes to inner transactions. But the code that fetches envelopes from the mempool can prioritize envelopes.

consensus/dbft/amev.go
vncoelho: Transactions can not be different!
If they are different the implementation has possible attack vectors and problems.

Txs should be exactly the same as well as the order.

consensus/dbft/dbft.go
AnnaShaleva: >  it should work as 1:1 mapper of envelopes to inner transactions

OK, then current code works like this.

roman-khimov (commented):
Good to go and test.

consensus/dbft/amev.go
AnnaShaleva: > Transactions can not be different!

But it's the whole idea of the current Envelopes design.

roman-khimov: I think we're talking the same thing in a different way. To me transactions are different since we extract inner transactions from envelopes and envelopes are thrown away. But we're doing all of this based on the agreed upon (preliminary committed) set of transactions, nothing new (except decrypted transactions) will get in and only envelopes can go out (or can not if decrypted contents is garbage or somehow violates our rules).

consensus/dbft/dbft.go
AnnaShaleva: Yes, let's implement it in https://github.com/bane-labs/go-ethereum/issues/318, in the next PR.

AnnaShaleva:
I checked the final version one more time on privnet with some side encrypted data (just a message, not a fully-qualified transaction), the decryption part works:
```
2024-09-10T15:18:40.026+0300	DEBUG	dbft@v0.3.1-0.20240904101105-212d1768bdcf/dbft.go:452	ignoring PrepareResponse	{"dup": true, "sor": true, "viewChanging": false, "moreThanF": true}
2024-09-10T15:18:40.026+0300	DEBUG	dbft@v0.3.1-0.20240904101105-212d1768bdcf/dbft.go:231	received message	{"type": "PreCommit", "from": 1, "height": 21, "view": 0, "my_height": 21, "my_view": 0}
2024-09-10T15:18:40.032+0300	DEBUG	dbft@v0.3.1-0.20240904101105-212d1768bdcf/dbft.go:231	received message	{"type": "PreCommit", "from": 5, "height": 21, "view": 0, "my_height": 21, "my_view": 0}
2024-09-10T15:18:40.032+0300	INFO	dbft@v0.3.1-0.20240904101105-212d1768bdcf/dbft.go:544	received PreCommit	{"validator": 5}
2024-09-10T15:18:40.036+0300	INFO	dbft@v0.3.1-0.20240904101105-212d1768bdcf/check.go:66	processing PreBlock	{"height": 21, "view": 0, "tx_count": 1, "preCommit_count": 5}
INFO [09-10|15:18:40.041] Envelope data decrypted                  "envelope hash"=2c7837..cd40bb "envelope index"=0 data=736f6d6520646174612074686174206973206d6f7265207468616e2031303520627974657320696e206c656e6774683a2070697a7a612070697a7a612070697a7a612070697a7a612070697a7a612070697a7a612070697a7a612070697a7a612070697a7a612070697a7a612070697a7a612070697a7a612070697a7a61
INFO [09-10|15:18:40.041] Decrypted transaction decoding failed    "envelope hash"=2c7837..cd40bb "envelope index"=0 data=736f6d6520646174612074686174206973206d6f7265207468616e2031303520627974657320696e206c656e6774683a2070697a7a612070697a7a612070697a7a612070697a7a612070697a7a612070697a7a612070697a7a612070697a7a612070697a7a612070697a7a612070697a7a612070697a7a612070697a7a61 error="typed transaction too short"
2024-09-10T15:18:40.050+0300	INFO	dbft@v0.3.1-0.20240904101105-212d1768bdcf/send.go:150	sending Commit	{"height": 21, "view": 0}
2024-09-10T15:18:40.050+0300	DEBUG	dbft@v0.3.1-0.20240904101105-212d1768bdcf/send.go:8	broadcasting message	{"type": "Commit", "height": 21, "view": 0}

...

2024-09-10T15:18:40.059+0300	INFO	dbft@v0.3.1-0.20240904101105-212d1768bdcf/dbft.go:588	received Commit	{"validator": 5}
2024-09-10T15:18:40.059+0300	INFO	dbft@v0.3.1-0.20240904101105-212d1768bdcf/check.go:120	approving block	{"height": 21, "hash": "0xc11f7cf2edf031d616a28571575ed9116767993ac7597cc70f2e864354a61b80", "tx_count": 1, "merkle": "0xad3cc68ad02532fbfad7870fff93b6e6f6a0ab0514e8bf45a5ca92f6212bec73", "prev": "0x7edfd63c7fdc9589b4b0da2b12f9c5f102fe31025adcdf889fc671c78f300e85"}
INFO [09-10|15:18:40.110] Chain reorg detected                     number=19 hash=3a1aa2..9377cf drop=1 dropfrom=89b3ea..d8114c add=2 addfrom=44bcbf..ff8e66
INFO [09-10|15:18:40.116] Successfully wrote new block with state  number=21 hash=44bcbf..ff8e66
INFO [09-10|15:18:40.117] New block in the chain                   "dbft index"=21 "chain index"=21 hash=0x44bcbf567c01b8de0caa52f98035cf387ef916844663aaf6f5bb4ac75eff8e66 "parent hash"=0x7edfd63c7fdc9589b4b0da2b12f9c5f102fe31025adcdf889fc671c78f300e85 primary=0 coinbase=0x1212000000000000000000000000000000000003 "mix digest"=0x3e48bc8f2eb502267b1eae6bded2ece31a29829cdf94afb44f37688702643e43
```

AnnaShaleva (approved):
The main logic is implemented, some missing things are moved to issues. It still needs a good testing, but if some bugs are found, we can fix them in further PRs.

txhsl (approved):

chenquanyu (approved):

PR #288, opened 2024-08-03
evm: fix opcode random nil error by setting random to zero

### LINKED ISSUE #290: Empty `Random` field since Shanghai fork
#### System information

Geth version: `v0.2.1`

#### Expected behaviour

Safe opcode handling in EVM.

#### Actual behaviour

After `The Merge`, there is an extra payload from beacon chain which contains the following attributes.

https://github.com/bane-labs/go-ethereum/blob/a8a87586c143337df53d137e498dd969c7fde549/beacon/engine/types.go#L41-L49

The `Withdrawals` data is now handled by dBFT in #122, but the field `Random` is left empty in EVM context.

Since Solidity `0.8.18`, the `block.difficulty` is replaced by `block.prevrandao`, and the op `opDifficulty` is replaced by `opRandom`. However, the context `Random` is `nil`, so the EVM panics and closes the node.

https://github.com/bane-labs/go-ethereum/blob/a5a4fa7032bb248f5a7c40f4e8df2b131c4186a4/core/vm/instructions.go#L472-L482

There is an enforce order of `Merge` and `Shanghai` instruction set, but we don't have the commit https://github.com/ethereum/go-ethereum/commit/ac0ff044606a663eeb47ef60ed5506f842753084. We don't have `Merge`, but we are now at `Shanghai`.

https://github.com/bane-labs/go-ethereum/blob/a5a4fa7032bb248f5a7c40f4e8df2b131c4186a4/core/vm/jump_table.go#L93-L110

Execute any contract that requests a value from `block.difficulty` or `block.prevrandao`, regardless of compiler versions.

#### Backtrace

Sorry, but I only have a screenshot.

chenquanyu:
Fix opcode random nil error by setting random to zero, refer to https://github.com/ethereum/go-ethereum/issues/29722

### DIFF
--- a/core/evm.go
+++ b/core/evm.go
@@ -43,7 +43,8 @@ func NewEVMBlockContext(header *types.Header, chain ChainContext, author *common
 		beneficiary common.Address
 		baseFee     *big.Int
 		blobBaseFee *big.Int
-		random      *common.Hash
+		// use zero for random
+		random *common.Hash = &common.Hash{}
 	)
 
 	// If we don't have an explicit author (i.e. not mining), extract from the header
--- a/core/vm/runtime/runtime.go
+++ b/core/vm/runtime/runtime.go
@@ -56,22 +56,31 @@ type Config struct {
 
 // sets defaults on the config
 func setDefaults(cfg *Config) {
+	var (
+		shanghaiTime = uint64(0)
+	)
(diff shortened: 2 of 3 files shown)


txhsl:
UTs failed.

PR #301, opened 2024-08-16
core: add antimev related crypto and keystore

### LINKED ISSUE #300: Add amev related crypto library and key management module
# Rationale

To provide our anti-mev feature, we first need to import some codes about the key generation, encryption, and decryption.

# Implementation

We already have an experimental demo [txhsl/tpke](https://github.com/txhsl/tpke) based on the bls library [kilic/bls12-381](https://github.com/kilic/bls12-381), so I will import the codes and implement a management module.

### LINKED ISSUE #308: Replace Geth implementation of BLS12-381 by `consensys/gnark-crypto`
Ref. https://github.com/bane-labs/go-ethereum/pull/301#discussion_r1723322495, can be done only after #233.

txhsl:
Close #300.

Some implemented codes from branch [antimev-crypto-bls12381](https://github.com/bane-labs/go-ethereum/tree/antimev-crypto-bls12381), without contract and zero knowledge, not integrated with anything else.

1. Some necessary methods from [kilic/bls12-381](https://github.com/kilic/bls12-381) are imported;
2. A tpke library that defines some data structures, with their constructing and encoding;
3. A keystore module that provides key management service, with encryption, decryption, and signature.

### DIFF
--- a/core/antimev/keystore.go
+++ b/core/antimev/keystore.go
@@ -1,0 +1,233 @@
+package antimev
+
+import (
+	"errors"
+	"math/big"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/crypto/ecies"
+	"github.com/ethereum/go-ethereum/crypto/tpke"
+)
+
+var (
+	ErrInvalidLength     = errors.New("invalid array length")
+	ErrInvalidThreshold  = errors.New("invalid threshold")
+	ErrNotParticipant    = errors.New("not dkg participant")
+	ErrUnrecoverable     = errors.New("unrecoverable sharing")
+	ErrMessageDecryption = errors.New("message decryption failed")
+)
+
+// AMEVKeyStore is the container of all useful dkg information.
+type AMEVKeyStore struct {
+	size      int // The size of each key group
+	threshold int // The threshold of each key group
+	scaler    int // The scaler to speed up computation, refer to crypto/tpke
+
+	address   common.Address    // Self account address
+	ethPrvKey *ecies.PrivateKey // Self account secret key
+
+	recovering *thresholdKeyGroup // The recovering key group
+	resharing  *thresholdKeyGroup // The resharing key group
+	reshared   *thresholdKeyGroup // The group can decrypt old messages
+	shared     *thresholdKeyGroup // The sharing key group
+	sharing    *thresholdKeyGroup // The group can encrypt and decrypt new messages
+}
+
+// NewKeyStore returns a new instance of antimev key store.
+func NewKeyStore(addr common.Address, prvkey *ecies.PrivateKey, groupSize int, threshold int) (*AMEVKeyStore, error) {
+	// TODO: Recover keystore from persistence
+	if groupSize < threshold {
+		return nil, ErrInvalidThreshold
+	}
+	return &AMEVKeyStore{
+		size:      groupSize,
+		threshold: threshold,
+		scaler:    getScaler(groupSize, threshold),
+		address:   addr,
+		ethPrvKey: prvkey,
+	}, nil
+}
+
+// OnValidatorList initializes sharing and resharing, should be called
+// when the key group members are determined. It returns resharing
+// messages immediately.
+func (ks *AMEVKeyStore) OnValidatorList(validators []common.Address, pubkeys []*ecies.PublicKey) ([][]byte, []byte, error) {
+	if len(validators) != ks.size || len(pubkeys) != ks.size {
+		return nil, nil, ErrInvalidLength
+	}
+	// Set up groups for dkg sharing and resharing
+	if ks.shared != nil {
+		ks.resharing = ks.shared.newTemplateForReshare(validators, pubkeys)
+	}
+	ks.sharing = newThresholdKeyGroup(ks.address, validators, pubkeys)
+	// Generate secret resharing messages and pvss
+	if ks.resharing != nil {
+		rMsgs, pvss, err := ks.shared.dkgReshare(ks.resharing)
+		if err != nil || pvss == nil {
+			return nil, nil, err
+		}
+		return rMsgs, pvss.ToBytes(), nil
+	}
+	// Nothing to reshare
+	return nil, nil, nil
+}
+
+func (ks *AMEVKeyStore) OnRecoverStart(indexes []int, validators []common.Address, pubkeys []*ecies.PublicKey) ([][]byte, error) {
+	if ks.shared == nil {
+		return nil, nil
+	}
+	if len(validators) != len(indexes) || len(pubkeys) != len(indexes) {
+		return nil, ErrInvalidLength
+	}
+	// Return error if amount is more than recoverable
+	if len(indexes) > ks.size-ks.threshold {
+		return nil, ErrUnrecoverable
+	}
+	// Do nothing if no need to recover
+	if len(indexes) < 1 {
+		return nil, nil
+	}
+	// Only place new ones in recovering group
+	ks.recovering = ks.shared.newTemplateForRecover(indexes, validators, pubkeys)
+	// Generate secret recovering messages
+	return ks.shared.dkgRecover(indexes, pubkeys)
+}
+
+// OnRecoverFinish tries to finish ongoing recovering, should be called when
+// resharing fails. It returns resharing messages immediately if recoverd.
+func (ks *AMEVKeyStore) OnRecoverFinish() ([][]byte, []byte, error) {
+	msgs, pvss, err := ks.recovering.dkgReshareRecovered(ks.threshold, ks.resharing)
+	if err != nil || pvss == nil {
+		return nil, nil, err
+	}
+	return msgs, pvss.ToBytes(), nil
+}
+
+// OnReshareFinish tries to finish ongoing resharing, should be called before
+// the new round sharing. It returns sharing messages immediately.
+func (ks *AMEVKeyStore) OnReshareFinish() ([][]byte, []byte, error) {
+	// Check if resharing are finished and aggregate keys
+	if ks.resharing != nil {
+		err := ks.resharing.dkgAggregate(ks.scaler)
+		if err != nil {
+			return nil, nil, err
+		}
+	}
+	// Generate secret sharing messages and pvss
+	sMsgs, pvss, err := ks.sharing.dkgPrepare(ks.threshold)
+	if err != nil || pvss == nil {
+		return nil, nil, err
+	}
+	return sMsgs, pvss.ToBytes(), nil
+}
+
+// OnEpochChange tries to finish ongoing sharing, should be called before
+// the key group is needed for encryption and signing.
+func (ks *AMEVKeyStore) OnEpochChange() error {
+	// Check if sharing are finished and aggregate keys
+	if ks.sharing != nil {
+		err := ks.sharing.dkgAggregate(ks.scaler)
+		if err != nil {
+			return err
+		}
+	}
+	// Set finished dkg as current using
+	ks.reshared = ks.resharing
+	ks.resharing = nil
+	ks.shared = ks.sharing
+	ks.sharing = nil
+
+	return nil
+}
+
+// ReceiveSecretShare tries to verify a sharing message array and store their data.
+func (ks *AMEVKeyStore) ReceiveSecretShare(from common.Address, ess [][]byte, pvss []byte) error {
+	fromIndex := ks.sharing.holderIndex(from)
+	if fromIndex > ks.size || fromIndex < 1 {
+		return ErrNotParticipant
+	}
+	p, err := new(tpke.PVSS).FromBytes(pvss, ks.size, ks.threshold)
+	if err != nil {
+		return ErrDKGPVSS
+	}
+	// If is a member of receiver
+	selfIndex := ks.sharing.holderIndex(ks.address)
+	if selfIndex > 0 {
+		ss, err := ks.decryptSecretShare(ess[selfIndex-1])
+		if err != nil {
+			return ErrDecryptionFailed
+		}
+		return ks.sharing.receiveShareMessage(fromIndex, ss, p)
+	}
+	return ks.sharing.receiveShareMessage(fromIndex, nil, p)
+}
+
+// ReceiveSecretReshare tries to verify a resharing message array and store their data.
+func (ks *AMEVKeyStore) ReceiveSecretReshare(from common.Address, ers [][]byte, pvss []byte) error {
+	fromIndex := ks.shared.holderIndex(from)
+	if fromIndex > ks.size || fromIndex < 1 {
+		return ErrNotParticipant
+	}
+	p, err := new(tpke.PVSS).FromBytes(pvss, ks.size, ks.threshold)
+	if err != nil {
+		return ErrDKGPVSS
+	}
+	// If is a member of receiver
+	selfIndex := ks.resharing.holderIndex(ks.address)
+	if selfIndex > 0 {
+		ss, err := ks.decryptSecretShare(ers[selfIndex-1])
+		if err != nil {
+			return ErrDecryptionFailed
+		}
+		return ks.resharing.receiveReshareMessage(fromIndex, ss, p)
+	}
+	return ks.resharing.receiveReshareMessage(fromIndex, nil, p)
+}
+
+// ReceiveRecoveredReshare tries to verify a resharing message array and store their data.
+func (ks *AMEVKeyStore) ReceiveRecoveredReshare(from common.Address, ers [][]byte, pvss []byte) error {
+	fromIndex := ks.recovering.holderIndex(from)
+	if fromIndex > ks.size || fromIndex < 1 {
+		return ErrNotParticipant
+	}
+	p, err := new(tpke.PVSS).FromBytes(pvss, ks.size, ks.threshold)
+	if err != nil {
+		return ErrDKGPVSS
+	}
+	// If is a member of receiver
+	selfIndex := ks.resharing.holderIndex(ks.address)
+	if selfIndex > 0 {
+		ss, err := ks.decryptSecretShare(ers[selfIndex-1])
+		if err != nil {
+			return ErrDecryptionFailed
+		}
+		return ks.resharing.receiveReshareMessage(fromIndex, ss, p)
+	}
+	return ks.resharing.receiveReshareMessage(fromIndex, nil, p)
+}
+
+// ReceiveRecoverShare tries to verify a recovering message array and store their data.
+func (ks *AMEVKeyStore) ReceiveRecoverShare(from common.Address, ers []byte) error {
+	fromIndex := ks.shared.holderIndex(from)
+	if fromIndex > ks.size || fromIndex < 1 {
+		return ErrNotParticipant
+	}
+	// If is a member of receiver
+	selfIndex := ks.recovering.holderIndex(ks.address)
+	if selfIndex > 0 {
+		ss, err := ks.decryptSecretShare(ers)
+		if err != nil {
+			return ErrDecryptionFailed
+		}
+		return ks.recovering.receiveRecoverMessage(fromIndex, ss)
+	}
+	return nil
+}
+
+func (ks *AMEVKeyStore) decryptSecretShare(ess []byte) (*big.Int, error) {
+	ss, err := ks.ethPrvKey.Decrypt(ess, nil, nil)
+	if err != nil {
+		return nil, err
+	}
+	return new(big.Int).SetBytes(ss), nil
+}
--- a/core/antimev/keystore_test.go
+++ b/core/antimev/keystore_test.go
@@ -1,0 +1,395 @@
+package antimev
+
+import (
+	"math/rand"
+	"testing"
+	"time"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/ethereum/go-ethereum/crypto/ecies"
+)
+
+var size = 7
+var threshold = 5
+
+var addrs = []common.Address{
+	common.HexToAddress("0xcbbeca26e89011e32ba25610520b20741b809007"),
+	common.HexToAddress("0x4ea2a4697d40247c8be1f2b9ffa03a0e92dcbacc"),
+	common.HexToAddress("0xd10f47396dc6c76ad53546158751582d3e2683ef"),
+	common.HexToAddress("0xa51fe05b0183d01607bf48c1718d1168a1c11171"),
+	common.HexToAddress("0x01b517b301bb143476da35bb4a1399500d925514"),
+	common.HexToAddress("0x7976ad987d572377d39fb4bab86c80e08b6f8327"),
+	common.HexToAddress("0xd711da2d8c71a801fc351163337656f1321343a0"),
+}
+
+type MockContractStorage struct {
+	reshareMsgs   [][][]byte
+	resharePVSSes [][]byte
+	shareMsgs     [][][]byte
+	sharePVSSes   [][]byte
+	recoverMsgs   [][][]byte
+}
+
+func TestDKG(t *testing.T) {
+	source := rand.NewSource(time.Now().UnixNano())
+	random := rand.New(source)
+	// Init keystores
+	pubs := make([]*ecies.PublicKey, size)
+	kss := make([]*AMEVKeyStore, size)
+	for i := 0; i < size; i++ {
+		key, _ := ecies.GenerateKey(random, crypto.S256(), nil)
+		pubs[i] = &key.PublicKey
+		ks, err := NewKeyStore(addrs[i], key, size, threshold)
+		if err != nil {
+			t.Fatalf(err.Error())
+		}
+		kss[i] = ks
+	}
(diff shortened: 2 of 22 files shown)


crypto/bls12381/fp.go
```diff
@@ -34,6 +34,37 @@ func fromBytes(in []byte) (*fe, error) {
...
+	}
+	// F = 2 ^ 256 * R
+	F := fe{
+		0x75b3cd7c5ce820f,
+		0x3ec6ba621c3edb0b,
+		0x168a13d82bff6bce,
+		0x87663c4bf8c449d2,
+		0x15f34c83ddc8d830,
+		0xf9628b49caa2e85,
+	}
```
AnnaShaleva: Move it to a package-level variable, please.

crypto/bls12381/fp.go
```diff
@@ -34,6 +34,37 @@ func fromBytes(in []byte) (*fe, error) {
 	return fe, nil
 }
 
+func from64Bytes(in []byte) (*fe, error) {
```
AnnaShaleva: Just out of curiosity: why don't we use 48-bytes form? Also, does this 64-bytes form has its serialization standard name? Is it commonly-used in other projects? It requires a little bit more calculations than 48-bytes form.

crypto/bls12381/g1.go
```diff
@@ -70,6 +70,72 @@ func (g *G1) Q() *big.Int {
 	return new(big.Int).Set(q)
 }
 
+// FromCompressed expects byte slice at least 48 bytes and given bytes returns a new point in G1.
+// Serialization rules are in line with zcash library. See below for details.
+// https://github.com/zcash/librustzcash/blob/master/pairing/src/bls12_381/README.md#serialization
```
AnnaShaleva: Please, use links pointing to a specific commit instead of fresh master. ZCash has already moved this package to another location.

crypto/bls12381/g1.go
```diff
@@ -70,6 +70,72 @@ func (g *G1) Q() *big.Int {
...
+		return nil, errors.New("point is not on correct subgroup")
+	}
+	return p, nil
+}
+
+// ToCompressed given a G1 point returns bytes in compressed form of the point.
+// Serialization rules are in line with zcash library. See below for details.
+// https://github.com/zcash/librustzcash/blob/master/pairing/src/bls12_381/README.md#serialization
+// https://docs.rs/bls12_381/0.1.1/bls12_381/notes/serialization/index.html
+func (g *G1) ToCompressed(p *PointG1) []byte {
```
AnnaShaleva: I'd prefer `ToBytesCompressed` to follow the package naming style.

crypto/bls12381/g1.go
```diff
@@ -70,6 +70,72 @@ func (g *G1) Q() *big.Int {
 	return new(big.Int).Set(q)
 }
 
+// FromCompressed expects byte slice at least 48 bytes and given bytes returns a new point in G1.
+// Serialization rules are in line with zcash library. See below for details.
+// https://github.com/zcash/librustzcash/blob/master/pairing/src/bls12_381/README.md#serialization
+// https://docs.rs/bls12_381/0.1.1/bls12_381/notes/serialization/index.html
+func (g *G1) FromCompressed(compressed []byte) (*PointG1, error) {
```
AnnaShaleva: And `FromBytesCompressed` respectively.

crypto/bls12381/g1.go
```diff
@@ -70,6 +70,72 @@ func (g *G1) Q() *big.Int {
 	return new(big.Int).Set(q)
 }
 
+// FromCompressed expects byte slice at least 48 bytes and given bytes returns a new point in G1.
```
AnnaShaleva: > at least 48 bytes

It's exactly 48 bytes.

crypto/bls12381/g1.go
```diff
@@ -70,6 +70,72 @@ func (g *G1) Q() *big.Int {
...
+// https://docs.rs/bls12_381/0.1.1/bls12_381/notes/serialization/index.html
+func (g *G1) FromCompressed(compressed []byte) (*PointG1, error) {
+	if len(compressed) != 48 {
+		return nil, errors.New("input string length must be equal to 48 bytes")
+	}
+	var in [48]byte
+	copy(in[:], compressed[:])
+	if in[0]&(1<<7) == 0 {
+		return nil, errors.New("compression flag must be set")
+	}
```
AnnaShaleva: Isn't it ZCash-specific flags?

crypto/bls12381/g2.go
```diff
@@ -79,6 +79,72 @@ func (g *G2) Q() *big.Int {
 	return new(big.Int).Set(q)
 }
 
+// FromCompressed expects byte slice at least 96 bytes and given bytes returns a new point in G2.
+// Serialization rules are in line with zcash library. See below for details.
+// https://github.com/zcash/librustzcash/blob/master/pairing/src/bls12_381/README.md#serialization
+// https://docs.rs/bls12_381/0.1.1/bls12_381/notes/serialization/index.html
+func (g *G2) FromCompressed(compressed []byte) (*PointG2, error) {
+	if len(compressed) != 2*48 {
```
AnnaShaleva: This code is kind of a hot path, I'd suggest move it to some constant.

crypto/bls12381/g2.go
```diff
@@ -79,6 +79,72 @@ func (g *G2) Q() *big.Int {
 	return new(big.Int).Set(q)
 }
 
+// FromCompressed expects byte slice at least 96 bytes and given bytes returns a new point in G2.
+// Serialization rules are in line with zcash library. See below for details.
+// https://github.com/zcash/librustzcash/blob/master/pairing/src/bls12_381/README.md#serialization
+// https://docs.rs/bls12_381/0.1.1/bls12_381/notes/serialization/index.html
+func (g *G2) FromCompressed(compressed []byte) (*PointG2, error) {
```
AnnaShaleva: Regarding thees G1/G2 serialization/deserialization: have you considered reusing existing Go libraries for this? For example, in NeoGo we use `Consensys/gnark-crypto` https://github.com/nspcc-dev/neo-go/blob/dfd4566a04a946c17e16631a7a1033a3e69dd990/pkg/core/native/crypto_blspoints.go#L112. It's easy to convert points from gnark-crypto format to Geth format.

I'm a little bit afraid of manual EC-related crypto implementations because it may miss some edge-cases and it needs to be well-tested.

core/antimev/keystore.go
```diff
@@ -0,0 +1,233 @@
...
+)
+
+// AMEVKeyStore is the container of all useful dkg information.
+type AMEVKeyStore struct {
+	size      int // The size of each key group
+	threshold int // The threshold of each key group
+	scaler    int // The scaler to speed up computation, refer to crypto/tpke
+
+	address   common.Address    // Self account address
+	ethPrvKey *ecies.PrivateKey // Self account secret key
```
AnnaShaleva, line 27: What is the relation between `address` and `ethPrvKey`? May I use Eth CN's address with a separate newly-generated `*ecies.PrivateKey`?

core/antimev/keystore_test.go
```diff
@@ -0,0 +1,395 @@
...
+	}
+	// Send secret sharing messages
+	for i := 0; i < size; i++ {
+		for j := 0; j < size; j++ {
+			err := kss[i].ReceiveSecretShare(kss[j].address, contract.shareMsgs[j], contract.sharePVSSes[j])
+			if err != nil {
+				t.Fatalf(err.Error())
+			}
+		}
+	}
```
AnnaShaleva, line 128: For this we need a separate network protocol to broadcast DKG-related messages to other candidates. @roman-khimov, @txhsl do you think it should be a part of `eth/protocols/dbft`?

core/antimev/tpke_test.go
```diff
@@ -0,0 +1,94 @@
...
+	// Decrypt
+	results, err := kss[0].AggregateAndDecryptWithShare([]*tpke.CipherText{encryptedKey}, [][]byte{encryptedMsg}, shares)
+	if err != nil {
+		t.Fatalf(err.Error())
+	}
+	for i := 0; i < len(msg); i++ {
+		if msg[i] != results[0][i] {
+			t.Fatalf("decryption failed.")
+		}
+	}
```
AnnaShaleva, line 94: @txhsl, I'm using this test as an example to implement transaction decryption on the dBFT side. Could you please confirm or adjust the following statement:
* `encryptedMsg` corresponds to the encrypted transaction, all CNs retrieve these transactions from Envelopes content.
* `shares` corresponds to the messages that CNs exchange during `Commit` phase for every encrypted transaction in block.
* `encryptedKey` - I'm not quite sure where CNs should get these data from, supposing that some user encrypts transaction via `kss[0].Encrypt(msg)` and shares only the resulting `encryptedMsg`. Could you, please, clarify?


Also, I have a question: I thought that `M` CNs are enough to decrypt the message (i.e. `M` commit messages is enough to collect `M` shares and call `kss[0].AggregateAndDecryptWithShare`. However, in your example, I see `N` calls to `kss[i].DecryptWithShare` (where `N` is a `size` and `M` is a `threshold`). Where I'm wrong?

AnnaShaleva (commented):
I can't comment a lot about crypto-related operations without very careful review. However, the resulting DKG API seems to be sufficient to implement key generation protocol, and TPKE API also looks good, tests are very helpful to understand it. I'm trying to integrate TPKE into dBFT, and have a couple of questions (see the review).

crypto/bls12381/fp.go
txhsl: The 48-bytes form is, yes, used for normal serialization. But this 64-bytes form is required by `hashToFpXMDSHA256()` and `EncodeToCurve()`, that is, to transform message data to a valid curve point.

It is some standard, so that we can use bls signatures following `BLS12381G1_XMD:SHA-256_SSWU_RO_` or `BLS12381G2_XMD:SHA-256_SSWU_RO_`.

Ref https://www.ietf.org/archive/id/draft-irtf-cfrg-hash-to-curve-06.html.

crypto/bls12381/fp.go
txhsl: Just following the original coding, ref https://github.com/kilic/bls12-381/blob/master/fp.go#L38, this is where the Ethereum `crypto/bls12381` comes from.

crypto/bls12381/g1.go
txhsl: Actually this is a copy-paste, and `ToCompressed` is a widely recognized naming.

core/antimev/keystore.go
txhsl: Well, let me clarify. The `address` and `ethPrvKey` here has no relationship, `address` here is used for self-identification, and `ethPrvKey` is used for secret share message decryption.

They are combined like what [`thresholdKeyHolder`](https://github.com/bane-labs/go-ethereum/pull/301/files#diff-c25664f9a99b8a950756dee1fc80387a235a572a46274c7733b59df226e51e4aR38) does. It depends on the context, we can use newly-generated keys, if we can not get the public key from address and have to upload a public key anyway. How do you want this to be, we can make a change.

crypto/bls12381/g1.go
txhsl: Yes, correct.

crypto/bls12381/g1.go
txhsl: Let me check.

crypto/bls12381/g1.go
txhsl: It is, but their codes become somehow a standard library, ref https://github.com/zkcrypto/bls12_381/blob/main/src/g1.rs#L227.

crypto/bls12381/g2.go
txhsl: Yes, but no. If you look through the files, thing are all like that. If you mean change `2*48` to `96`, then we can make it.

crypto/bls12381/g1.go
txhsl: Updated.

crypto/tpke/encryption.go
```diff
@@ -0,0 +1,155 @@
...
+				g1.Neg(minor, minor)
+			}
+			g1.Add(rpk, rpk, minor)
+		}
+		// Divide d1 by d
+		g1.MulScalar(rpk, rpk, denominator)
+		// Decrypt
+		results[i] = g1.Sub(g1.Zero(), cts[i].cMsg, rpk)
+		// Verify the decryption
+		go parallelVerify(i, cts[i], pub.pg1, rpk, ch)
```
txhsl: @AnnaShaleva I need suggestions about this.

Here is a performance bottleneck (e.g. 1000 transactions), If I remove the `go` word here, then the time cost of decryption will increase from milliseconds (hundreds, sort of fast) to seconds (usually 1 or 2 seconds).

But goruntine like this may take over a lot of memories immediately.

Which one is better?

core/antimev/keystore_test.go
txhsl: The DKG related messages are planned to be handled through contract.

A protocol message also works, but we need this to be connected with Governance epoch.

core/antimev/tpke_test.go
txhsl: About `encryptedMsg` and `shares`, it's YES. About `encryptedKey`, it should be a part of encrypted transaction, like some extra data.

Some user encrypts his transaction and gets `encryptedMsg` and `encryptedKey`, and above two things together is called an `envelope`. For CNs, they need to decrypt `encryptedKey` through commit messages, then decrypt `encryptedMsg`, which is the transaction locally.

> However, in your example, I see `N` calls to `kss[i].DecryptWithShare` (where `N` is a `size` and `M` is a `threshold`).

Yes, `M` shares is enough. But just suppose, 1 of this `M` is not correct, then the decryption fails, and we are not able to figure out which one is not correct.

The `AggregateAndDecryptWithShare` is accepting more than `M` shares. It will succeed if there is **any** `M` of these shares **can** generate a verifiable decryption result, but will fail if **every** `M` of the shares **can't** generate a verifiable decryption result.

crypto/bls12381/g2.go
txhsl: Yes... I just found the `ethereum/crypto/bls12381` was replaced by `consensys/gnark-crypto/ecc/bls12-381` since v1.14.

I can rework TPKE when we upgrade and drop `ethereum/crypto/bls12381`, but for now, they are different things. I don't expect two BLS12381 library here, especially this old one may be outdated.

crypto/bls12381/fp.go
AnnaShaleva: Good, then let's preserve the origin library style.

crypto/bls12381/g2.go
AnnaShaleva: You're right, the whole package uses it like this.

crypto/bls12381/g2.go
AnnaShaleva: Let's have https://github.com/bane-labs/go-ethereum/issues/308 for further development.

txhsl: Good.

core/antimev/tpke_test.go
AnnaShaleva: Thank you for the explanation, now it's clear to me and it requires one more dBFT library modification, because with this scheme we can verify `Commit` messages only when at least `M` commits are collected.

crypto/tpke/encryption.go
AnnaShaleva: I'd suggest to use an intermediate way: define some constant number of working threads and run parallel verification in batches. Workers will work in parallel, and every worker need to perform sequential verification of his own batch of tasks in cycle. You're right that we can't allow unlimited goroutines growth due to resource constraint and also we can't allow sequential execution.

As an example, you may take a look on how we check multisignature in NeoGo, but this example is a bit simpler than TPKE decryption because standard multisignature in N3 assumes that both public keys and signatures are sorted. https://github.com/nspcc-dev/neo-go/blob/dc6c195637e98a537f9128c2e05bef752fee1c91/pkg/vm/vm.go#L1895

AnnaShaleva: The exact number of working threads needs to be defined in experimental way, we need to benchmark different numbers and check how much memory they require.

Also, I have a question (or suggestion) regarding this method: it allows to decrypt multiple transactions (e.g. we have several encrypted transactions in a single block), but if at least one decryption fails then the whole method will return an error. It's not quite correct to me, because the rest of transactions may still be successfully decrypted (is it true?). If so, then for those transactions that are decrypted, we need to include the decrypted tx to the final block, and for those that can't be decrypted we need to include their Envelopes to the final block.

core/antimev/tpke.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+	return ks.reshared.decryptShare(cts)
+}
+
+// AggregateAndDecryptWithShare tries to aggregate decryption shares and finally
+// decrypt the aes keys and the raw messages. The encrypted keys, encrypted
+// messages and decryption shares should have the same input order. The key of
+// inputs is dkg index which starts from 1, when the array index of a member in
+// the key group starts from 0.
+func (ks *AMEVKeyStore) AggregateAndDecryptWithShare(cts []*tpke.CipherText, msg [][]byte, inputs map[int]([]*tpke.DecryptionShare)) ([][]byte, error) {
```
AnnaShaleva, line 64: May we add an additional check in the start of this method: length of `cts` should be equal to the length of `msg`.

crypto/tpke/encryption.go
txhsl: > It's not quite correct to me, because the rest of transactions may still be successfully decrypted (is it true?).

I think I have to explain it a bit more, the failed “decryption” here doesn't equal whether the transaction can be decrypted or not. 

When a user is encrypting his AES key, he is declaring a secret random number $r$, and combine this number with global public key $PK$ to encrypt a BLS curve G1 point $M$.

So the TPKE encryption is computing $C=M+rPK$, $C$ is the encrypted AES key, for CNs' decryption, they are actually trying to recover $M$ by computing the $rPK$.

To realize that, TPKE requires $R_1=r{G_1}$ and $R_2=r{G_2}$ as commitments, $R_1$ is the thing that CNs work on to get $rPK$, and $R_2$ is used for the verification.

When a node receives an envelope, it can verify $R_2$ against $R_1$, in order to know **whether the AES key is well-encrypted**. Let's call it the `Verification I`.
When CNs receive shares, they can verify $R_2$ against the result, in order to know **whether the AES key is well-decrypted**. Let's call it the `Verification II`.

For an encrypted key $C$, any value $R_1$ can point to a decrypted $M'$.

- If the user declares a valid $R_1$ and a valid $R_2$, then `Verification I` will pass, and the transaction is decryptable.
- If the user declares a valid $R_1$ but an invalid $R_2$, then `Verification I` will fail, then the transaction is rejected by mempool.
- If the user declares an invalid $R_1$ but a valid $R_2$ (contains the same $r$ as $R_1$), then `Verification I` will pass, then the transaction is not decryptable but accepted by mempool.

So when the envelope arrives to CNs, it can be whether decryptable or not. But anyway, its $R_1$ and $R_2$ point to a same value $M'$.

The point is, CNs don't need to care about whether $M'$ is the exact $M$, but whether the $M'$ is the same thing as $R_1$ and $R_2$ pointed to, which is the `Verification II`.

The `Verification II` here fails only when the combination of CNs' shares are not valid. Any failure here, even just only one, means that at least one guy in CNs shared something wrong about the transactions.

So, it is a choice of handling, when some information of CNs‘ shares are wrong, the codes throw an error and give up this attempt. I think it is clearer than only accepting the correct $M'$ s, because we have to wait for other shares to compute the left $M'$ s anyway. The error is caused by malicious CNs, not the user.

If a transaction is not decryptable, then the AES decryption should fail.

If a transaction is decryptable but not legal, then the verification after AES decryption should fail.

core/antimev/tpke.go
txhsl: Yes, I will add it, and also check if other methods should check like this way.

crypto/tpke/encryption.go
txhsl: > The exact number of working threads needs to be defined in experimental way, we need to benchmark different numbers and check how much memory they require.

I test some benchmarks on my laptop. The memory issue is mainly caused by AES decryption with `crypto/aes`, so I lower the data size from `1 MB` to `128 KB` as mentioned in https://github.com/bane-labs/go-ethereum/pull/235#issuecomment-2301394531, then everything works well with smooth memory management.

I also set the `verifierCount` to `24`, it works well for me, with almost no performance choke.

The simulated result shows we can handle `5000 envelopes` (`128 KB` each) with TPKE in about `1.8 seconds`, about `1.5 GB` memory will be allocated and released in this process.

txhsl:
Will clean up commits before merge.

crypto/tpke/encryption.go
AnnaShaleva: > we can handle 5000 envelopes (128 KB each) with TPKE in about 1.8 seconds

We should consider restricting the number of envelopes per block to some lower value. I'm thinking about it because this part of code (`Verification II`, `(ks *AMEVKeyStore) AggregateAndDecryptWithShare` method in your PR) should be called every time when CN receives subsequent `PreCommit` message with share (starting from `M` `PreComits`) to check that Envelopes may be decrypted before the final Block construction. And in worst scenario this method will be called `F+1` times in terms of dBFT. For 7 CNs it's 3 times per block which easily gives us 5.4 seconds which is almost half of block time, we just can't allow such verification delays.

What do you think?

crypto/tpke/encryption.go
```diff
@@ -117,7 +125,11 @@ func AggregateAndDecrypt(cts []*CipherText, matrix [][]int, shares [][]*Decrypti
...
 	}
-	ch := make(chan error, len(cts))
+	in := make(chan verifyTask, len(cts))
+	out := make(chan error, len(cts))
+	for i := 0; i < verifierCount; i++ {
+		go startVerifier(in, out)
+	}
 	g1 := bls12381.NewG1()
 	for i := 0; i < len(cts); i++ {
 		rpk := g1.Zero()
```
AnnaShaleva: A question: why not to start the goroutine right after `for i := 0; i < len(cts); i++ {`? So that this goroutine contains not only verification, but also decryption part.

core/antimev/keystore.go
```diff
@@ -0,0 +1,233 @@
...
+	recovering *thresholdKeyGroup // The recovering key group
+	resharing  *thresholdKeyGroup // The resharing key group
+	reshared   *thresholdKeyGroup // The group can decrypt old messages
+	shared     *thresholdKeyGroup // The sharing key group
+	sharing    *thresholdKeyGroup // The group can encrypt and decrypt new messages
+}
+
+// NewKeyStore returns a new instance of antimev key store.
+func NewKeyStore(addr common.Address, prvkey *ecies.PrivateKey, groupSize int, threshold int) (*AMEVKeyStore, error) {
+	// TODO: Recover keystore from persistence
```
AnnaShaleva, line 38: Regarding this TODO: I needed keystore recovering in #287 to be able to test TPKE-dBFT integration, and hence I created RLP encoding for the whole set of components in keystore (see the https://github.com/bane-labs/go-ethereum/pull/287/commits/59409aa46442935ad09a39cc86d641f6d2402b38). I doubt that it's useful for production because it's a bad idea to store unencrypted private keys and DKG-related data in a single blob on disk, but may be some of this code may be reused for this task. For now I'd suggest to move this TODO to a separate issue and continue with PR finalization.

AnnaShaleva (commented):
I tried to integrate TPKE part into dBFT, see the last two commits in #287. Integration is not finished yet, but from an intermediate point I can tell that TPKE part is sufficient for dBFT operations. Can't say anything from practical POW about DKG part, because I didn't use it in code, but from review it also looks good to me. I'd suggest to solve remaining conversations, create issues for TODOs in the code and merge this PR as an initial implementation. It may always be improved in further PRs (if needed).

crypto/tpke/encryption.go
txhsl: Yes, we need.

crypto/tpke/encryption.go
txhsl: Yes, we can, good suggestion.

crypto/tpke/encryption.go
chenquanyu: In fact, a block can hold at most 1,400+ transactions, because the block gas limit is around 30 million. I think maybe 500 envelope transactions is enough.

txhsl:
Sad, I found a potential bug in the current codes, the benchmark test sometimes fails and the whole batch of data is reported as not-decryptable.

This case never gives an error before https://github.com/txhsl/tpke/commit/7619c5c97d86e9268fcaeb8c8d48fa88cf02e60c. I need some time to locate it.

txhsl:
> Sad, I found a potential bug in the current codes, the benchmark test sometimes fails and the whole batch of data is reported as not-decryptable.

Fixed this by importing `gnark-crypto/ecc/bls12-381` and remove `ethereum/bls12381`. Will close #308 as well.

core/antimev/keystore.go
txhsl: Ref #310.

txhsl:
Ready for merge.

PR #307, opened 2024-08-21
antimev: add envelope type define and validateDecryptedTx

chenquanyu:
Adding envelope type definition and validateDecryptedTx, validateTransactionWithState and implementing decryption using dkg is still in progress.

### DIFF
--- a/consensus/dbft/amev.go
+++ b/consensus/dbft/amev.go
@@ -1,15 +1,31 @@
 package dbft
 
-import "github.com/ethereum/go-ethereum/core/types"
+import (
+	"bytes"
 
+	"github.com/ethereum/go-ethereum/core/systemcontracts"
+	"github.com/ethereum/go-ethereum/core/types"
+)
+
+const (
+	// txMinSize is the minimum size a single transaction can have.
+	// The size of a simple gas transfer with 1 gwei is 105 bytes, we use this as the minimum tx.
+	txMinSize = 105
+)
+
+// encryptDataPrefix is the prefix of Envelope tx data
+var encryptDataPrefix = []byte{0xff, 0xff, 0xff, 0xff}
+
+// isEnvelope checks whether a transaction is an envelope transaction,
+// including to address, data prefix and data length check.
 func isEnvelope(tx *types.Transaction) bool {
-	return false
-}
+	if tx.To() == nil || *(tx.To()) != systemcontracts.GovernanceRewardProxyHash {
+		return false
+	}
 
-// Envelope is an interface representing encrypted enveloped transaction.
-type Envelope interface {
-	// Decrypt decrypts the encripted content of the Envelope transaction using the
-	// provided shared key and returns two transactions: the original outer one and
-	// an unencrypted inner one.
-	Decrypt(sharedKey any) (*types.Transaction, *types.Transaction)
+	if len(tx.Data()) < txMinSize || !bytes.HasPrefix(tx.Data(), encryptDataPrefix) {
+		return false
+	}
+
+	return true
 }
--- a/consensus/dbft/dbft.go
+++ b/consensus/dbft/dbft.go
@@ -227,8 +227,10 @@ type DBFT struct {
 	sealingTransactions types.Transactions
 
 	// chain and mempool instances needed for proper dBFT callbacks functioning.
-	chain    ChainHeaderReader
-	txpool   txPool
+	chain      ChainHeaderReader
+	txpool     txPool
+	legacypool legacyPool
+
 	quit     chan struct{}
 	finished chan struct{}
 
@@ -430,7 +432,7 @@ func New(config *params.DBFTConfig, _ ethdb.Database) (*DBFT, error) {
 						txx[i] = pre.transactions[i]
 						continue
 					}
-					err := validateDecryptedTx(decryptedTx, pre.transactions[i])
+					err = c.legacypool.ValidateDecryptedTx(decryptedTx, pre.transactions[i])
 					if err != nil {
 						txx[i] = pre.transactions[i]
 						continue
@@ -897,6 +899,11 @@ events:
 // (fetching unknown transactions).
 func (c *DBFT) WithTxPool(pool txPool) {
 	c.txpool = pool
+}
+
+// WithLegacyPool initializes transaction legacy pool API for ValidateDecryptedTx
+func (c *DBFT) WithLegacyPool(pool legacyPool) {
+	c.legacypool = pool
 }
 
 // postBlock is a callback that updates latest accepted block data and resets
--- a/consensus/dbft/txpool.go
+++ b/consensus/dbft/txpool.go
@@ -12,3 +12,9 @@ type txPool interface {
 	// tx hash.
 	Get(hash common.Hash) *types.Transaction
 }
+
+// legacyPool defines the methods needed from a legacy pool
+type legacyPool interface {
+	// ValidateDecryptedTx checks the validity of the transaction to determine whether the outer envelope transaction should be replaced.
+	ValidateDecryptedTx(decryptedTx *types.Transaction, envelope *types.Transaction) error
+}
--- a/core/txpool/legacypool/legacypool.go
+++ b/core/txpool/legacypool/legacypool.go
@@ -19,6 +19,7 @@ package legacypool
 
 import (
 	"errors"
+	"fmt"
 	"math"
 	"math/big"
 	"sort"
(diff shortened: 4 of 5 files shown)


consensus/dbft/amev.go
```diff
@@ -1,15 +1,126 @@
...
+	"github.com/ethereum/go-ethereum/core"
+	"github.com/ethereum/go-ethereum/core/systemcontracts"
+	"github.com/ethereum/go-ethereum/core/txpool"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/params"
+)
+
+const (
+	// txMinSize is used to check if the envelope tx data has the minimum tx size.
+	txMinSize = 100
```
AnnaShaleva: Does this value has an explanation?

consensus/dbft/amev.go
```diff
@@ -1,15 +1,126 @@
...
+	// takes up based on its size. The slots are used as DoS protection, ensuring
+	// that validating a new transaction remains a constant operation (in reality
+	// O(maxslots), where max slots are 4 currently).
+	txSlotSize = 32 * 1024
+
+	// txMaxSize is the maximum size a single transaction can have. This field has
+	// non-trivial consequences: larger transactions are significantly harder and
+	// more expensive to propagate; larger transactions also take more resources
+	// to validate whether they fit into the pool or not.
+	txMaxSize = 4 * txSlotSize // 128KB
```
AnnaShaleva: Please, don't copy-past things from Geth if possible, because there might be unexpected consequences of synchronizing with fresh Geth source code.

Let's make legacypool's constants exported. It brings an unwanted dependency to dbft package, but to me it's still better than an alternative option.

consensus/dbft/amev.go
```diff
@@ -1,15 +1,126 @@
...
+
+// isEnvelope checks whether a transaction is an envelope transaction,
+// including to address, data prefix and data length check.
 func isEnvelope(tx *types.Transaction) bool {
-	return false
+	if tx.To() == nil || *(tx.To()) != systemcontracts.GovernanceRewardProxyHash {
+		return false
+	}
+
+	if len(tx.Data()) < txMinSize || bytes.HasPrefix(tx.Data(), encryptDataPrefix) {
```
AnnaShaleva: `|| bytes.HasPrefix` -> `|| !bytes.HasPrefix`

consensus/dbft/amev.go
```diff
@@ -1,15 +1,126 @@
...
+		return false
+	}
+
+	return true
+}
+
+// Decrypt decrypts the encripted content of the Envelope transaction using the
+// provided shared key and returns two transactions: the original outer one and
+// an unencrypted inner one.
+func Decrypt(sharedKey any, envelope *types.Transaction) (*types.Transaction, error) {
```
AnnaShaleva: Fix the documentation wrt this change.

consensus/dbft/amev.go
```diff
@@ -1,15 +1,126 @@
...
-	// provided shared key and returns two transactions: the original outer one and
-	// an unencrypted inner one.
-	Decrypt(sharedKey any) (*types.Transaction, *types.Transaction)
+// validateDecryptedTx checks the validity of the transaction to determine whether the outer envelope transaction should be replaced.
+func validateDecryptedTx(decryptedTx *types.Transaction, envelope *types.Transaction, head *types.Header) error {
+	if decryptedTx == nil {
+		return fmt.Errorf("decryptedTx is nil")
+	}
+	// Check transaction type
+	if decryptedTx.Type() != types.EncryptedTxType {
```
AnnaShaleva: We should merge #285 firstly.

consensus/dbft/amev.go
```diff
@@ -1,15 +1,126 @@
...
 }
 
-// Envelope is an interface representing encrypted enveloped transaction.
-type Envelope interface {
-	// Decrypt decrypts the encripted content of the Envelope transaction using the
-	// provided shared key and returns two transactions: the original outer one and
-	// an unencrypted inner one.
-	Decrypt(sharedKey any) (*types.Transaction, *types.Transaction)
+// validateDecryptedTx checks the validity of the transaction to determine whether the outer envelope transaction should be replaced.
+func validateDecryptedTx(decryptedTx *types.Transaction, envelope *types.Transaction, head *types.Header) error {
```
AnnaShaleva: This method is almost the full copy of txpool's `ValidateTransaction`. I suggest to extend txpool's `ValidateTransaction` with additional necessary checks and reuse this method for dBFT-related validation. Additional `isDecryptedTx bool` argument may be added to this method to distinguish calls from dBFT.

core/types/transaction.go
```diff
@@ -48,6 +48,8 @@ const (
 	AccessListTxType = 0x01
 	DynamicFeeTxType = 0x02
 	BlobTxType       = 0x03
+
+	EncryptedTxType = 0x7 // EncryptedTx is identical to DynamicFeeTx except for the type field.
```
AnnaShaleva: Remove it, it belongs to another PR.

AnnaShaleva:
@chenquanyu, we need to split the tasks, because I'm currently working on TPKE integration into dBFT, and there might be conflicts between this PR and the base `integrate-amev-dbft` branch. I suggest the following:
1. Finish #285 firstly and merge it, because both this PR and #287 depend on it.
2. Close #235 because it's not relevant anymore.
4. Refactor this PR wrt the review comments, make it small and not to implement DKG decryption in this PR, leave it for #287. Merge this PR once done.
5. In a new PR implement the necessary static decryption service fee logic for both Encrypted and Envelop txs (discussed earlier under https://github.com/bane-labs/go-ethereum/pull/235#issuecomment-2298346686).
6. Continue work onto https://github.com/bane-labs/go-ethereum/pull/286.

What do you think?

consensus/dbft/amev.go
AnnaShaleva: This method should be removed, it's not relevant anymore.

consensus/dbft/amev.go
```diff
@@ -0,0 +1,126 @@
...
+	if err != nil {
+		return txpool.ErrInvalidSender
+	}
+	decryptedFrom, err := types.Sender(signer, decryptedTx)
+	if err != nil {
+		return txpool.ErrInvalidSender
+	}
+	if envelopeFrom != decryptedFrom {
+		return fmt.Errorf("decryptedTx from mismatch: decryptedFrom %v, envelopeFrom %v", decryptedFrom, envelopeFrom)
+	}
```
AnnaShaleva: Please, remove this verification condition. The requirement of the same sender for both Envelope and encrypted transaction exposes sender of encrypted transaction to the outside world, so our encrypted transactions become not so encrypted.

Do you have strong objections against removing this requirement?

AnnaShaleva (commented):
@chenquanyu, let's fix all comments, rebase onto updated `antimev-tx-replace` and mege this PR.

consensus/dbft/amev.go
AnnaShaleva: Not relevant anymore, so let's just remove this check.

consensus/dbft/amev.go
chenquanyu: If different senders are allowed, then the same nonce is useless.

AnnaShaleva: Oh, you're right.

roman-khimov: I thought that hiding the sender could be beneficial. But then if we're to allow that we can have a collision between unencrypted and open transaction for the same sender/nonce. It can be solved (whatever pays more wins, for example), but having the same sender/nonce outside as we have inside prevents this collision. In any event, this can be adjusted in some future iterations, dropping requirements is easier than adding them.

consensus/dbft/amev.go
chenquanyu: This is from https://ethereum.stackexchange.com/questions/30175/what-is-the-size-bytes-of-a-simple-ethereum-transaction-versus-a-bitcoin-trans. But I tried with 1 gwei on bsc and got 105 bytes.

consensus/dbft/amev.go
chenquanyu: Move `ValidateDecryptedTx` to legacypool now.

consensus/dbft/amev.go
chenquanyu: Fixed.

consensus/dbft/amev.go
chenquanyu: Removed.

consensus/dbft/amev.go
chenquanyu: Move `ValidateDecryptedTx` to legacypool now.

core/types/transaction.go
chenquanyu: Removed.

AnnaShaleva (approved):
LGTM, let's merge and afterwards I'll extend/polish this code a bit.

consensus/dbft/amev.go
AnnaShaleva: Ref. https://github.com/bane-labs/go-ethereum/issues/309.

Issue #295, opened 2024-08-14
Add ZKP to DKG contract


doubiliu:
**Why should this feature exist?**
 Currently, our DKG contract uses the NIZK protocol. This is an interactive zkp protocol. The principle is to construct a Pedersen commitment to prove that DKG participants correctly sent encrypted key fragments to other participants. But this method requires the recipient to send a reply confirmation message. At present, this part of the code for confirmation information has not been written yet.
 
 We are going to replace the NIZK protocol with the non-interactive ZKP protocol. In this way, the receiver does not need to send additional confirmation transactions. 
   
The principle is that DKG participants attach a zk proof when sending encryption key fragments. The DKG contract can confirm that the key fragments are generated according to a fixed and correct logic by verifying the proof.We use [gnark](https://github.com/Consensys/gnark) to develop this function    

Regarding the zkp circuit, we have prepared 2 implementation solutions:

-   Option 1: The asymmetric encryption method used in the existing DKG code is hybrid encryption. This requires us to implement this method in the circuit. The implementation is relatively complicated and there is no progress for the time being.
-   Option 2: Simplify the asymmetric encryption method used in the existing DKG code and implement it according the formula `CipherText=KeyPoint+r*PublicKey`. We wrote a simple circuit [demo](https://github.com/doubiliu/tpke/tree/add-zkp).Because there may be a bug in a type conversion API used in the circuit, the demo cannot yet fully run.We are waiting for a reply from the developer of gnark : https://github.com/Consensys/gnark/discussions/1242

doubiliu:
By the way, let me explain the specific things we prove in this circuit.
1) Prove that the input `Publickey`, `CipherText`, `KeyPoint`, `Fi` are points on the elliptic curve
2) Prove that `CipherText ` is generated by `KeyPoint` according to `CipherText=KeyPoint+r*PublicKey`
3) Prove that the shared public key `Fi` is generated by the shared private key, that is, `Fi=fiG`
4) Prove that KeyPoint is encoded by the shared private key, that is, `KeyPoint.X==fi`

txhsl:
@AnnaShaleva According to our current evaluation, I'm afraid the adoption of ZKP may destroy the usage of `./privnet`, since we have to fix the amount of CNs as `7` and it is almost impossible to run 7 CNs on only 1 physical machine.

AnnaShaleva:
> it is almost impossible to run 7 CNs on only 1 physical machine.

[...]

Issue #293, opened 2024-08-13
Remove unused overrode proxy functions in system contracts


txhsl:
# Rationale

Ref https://github.com/bane-labs/bridge-evm-contracts/pull/181.

# Implementation

Remove the same part of the codes.

AnnaShaleva:
Ref. https://github.com/bane-labs/go-ethereum/pull/312#discussion_r1745056947.

Issue #233, opened 2024-06-21
Upgrade Geth to v1.14+


AnnaShaleva:
# Rationale

Fetch Geth [v1.14+](https://github.com/ethereum/go-ethereum/releases/tag/v1.14.0), we need to be up-to-date.

# Implementation

We have to deal with at least two major problems (see the release notes for more details):
1. > Geth v1.14.0 stops automatically constructing the pending block (https://github.com/ethereum/go-ethereum/pull/28623).
2. > Geth v1.14.0 drops support for running pre-merge networks (https://github.com/ethereum/go-ethereum/pull/29169).

AnnaShaleva:
One more problem introduced by [ac0ff044](https://github.com/ethereum/go-ethereum/pull/29023). Shanghai (and other post-merge hardforks) are not acceptable before the merge anymore even if configured from the genesis block:
```go
   	IsMerge:          isMerge,
   	IsShanghai:       isMerge && c.IsShanghai(num, timestamp),
   	IsCancun:         isMerge && c.IsCancun(num, timestamp),
   	IsPrague:         isMerge && c.IsPrague(num, timestamp),
   	IsVerkle:         isMerge && c.IsVerkle(num, timestamp),
```

The problem is connected with the way how the merge is defined:
```go
isPostMerge := header.Difficulty.Cmp(common.Big0) == 0
```

dBFT always sets non-zero difficulty (ref. #58 and #117) and zero difficulty is not possible without significant changes in the beacon engine. The problem is: dBFT is integrated as a pre-merge engine, and pre-merge engines are not supposed to have zero difficulty (we've already discussed this problem in ).

txhsl:
Can we ignore or revert the Merge things? Because we are not going to have a beacon chain or a beacon node.

AnnaShaleva:
> Can we ignore or revert the Merge things?

We can't ignore them because we need them, new features are introduced and users will expect it to work proterly. We can revert *some* of them, but it's not that simple and it's a way that will end in larger and larger efforts from our side to keep our codebase at least *working*. We need Geth 1.14+, and the last described problem is relatively easy to solve (just revert these `isMerge` changes). But first two problems go deep in the protocol, we'll stuck in reverting them and making the rest of the code work properly, because Geth is actively developing and there is new code written over the new post-merge base, and we need this new code.

> Because we are not going to have a beacon chain or a beacon node.

[...]

Issue #189, opened 2024-04-27
Enable Dencun


AnnaShaleva:
Deneb and Cancun are already active in Eth, we need to carefully investigate the changes introduced by these hardofkrs and enable them (or at least partially enable them) for dBFT.

Depends on #188.

AnnaShaleva:
I've reviewed all changes introduced in the original repo up to v1.13.15 release, and Dencun is already presented there. However, further Get version (v1.14.0) contains some code that is included into Dencun as far (https://github.com/ethereum/go-ethereum/pull/29180), thus I doubt we can include and fully support this hardfork in v0.1.2. Although I need some more time to check.

AnnaShaleva:
Moving to the next release since we'd like v0.1.2 to be stable.

AnnaShaleva:
Within the scope of this issue we need to update Hardhat settings for native contracts compilation:
https://github.com/bane-labs/go-ethereum/blob/971cac59f17c2c4673f8170daf2a9fc94acee74b/contracts/hardhat.config.ts#L8

AnnaShaleva:
Fetched in #232, enabled for Testnet in #522.

PR #312, opened 2024-09-02
contracts: add and update system contracts for dkg

### LINKED ISSUE #159: Adjust governance contracts for DKG
# Rationale

Anti-MEV requires update to contracts to be implemented, candidates should exchange key material to be able to work with encrypted transactions, see #22.

txhsl:
Ready for review. Close #159, but #295 not included.

Glad to see BLS12381-related precompile contracts are finally supported in https://github.com/ethereum/go-ethereum/commit/823719b9e1b72174cd8245ae9e6f6f7d7072a8d6.

Contracts in this PR are compatible with #301, so there can be an integration for automatic secret sharing with amev keystore.

Prague is not supported yet by neither Hardhat or dBFT, so a manual test is necessary before merge.

Checklist:

- [x] Test BLS and DKG contracts manually (PVSS verification and global key aggregation);
- [x] Lock Governance vote for sharing period;
- [x] Enable Governance `onPersist` check.
- [x] Fix tests


contracts/hardhat.config.ts
```diff
@@ -5,7 +5,8 @@ const config: HardhatUserConfig = {
   solidity: {
     version: "0.8.25",
     settings: {
-      evmVersion: "shanghai",
+      evmVersion: "cancun",
```
AnnaShaleva, line 8: We can't do this right now, Cancun is not supported by dBFT yet and disabled on Mainnet/Testnet. Cancun opcodes just won't be handled. At the same time we can't make BLS precompiles to be a Shanghai extension because there will be a problem on Mainnet/Testnet update since Shanghai is already enabled starting from 0 in our networks.

@txhsl, could you confirm? And if so, then we need #189, and for that we firstly need #233.

contracts/solidity/Governance.sol
```diff
@@ -252,25 +260,43 @@ contract Governance is IGovernance, ReentrancyGuard, GovProxyUpgradeable {
...
+        ) {
+            // compute and update consensus
+            address[] memory candidates = candidateList.values();
+            uint length = candidates.length;
+            if (length < consensusSize || totalVotes < voteTargetAmount) {
+                pendingConsensus = standByValidators;
+            } else {
+                pendingConsensus = _computeConsensus(candidates);
+            }
+            emit Persist(pendingConsensus);
```
AnnaShaleva: Could you please clarify, why do we use `3 * sharePeriodDuration`? And what is the relation between `sharePeriodDuration` and `epochDuration`? I thought we're going to use the same `epochDuration` as a period for keys setup.

AnnaShaleva: We calculate `pendingConsensus` once in the middle of the epoch then this value is cached till the end of the epoch. But what happens if during that period there will be more votes? These votes won't be taken into account during the subsequent consensus update in the end of the epoch. However, `epochStartGasPerVote` will be updated wrt these votes which leads to incorrect reward computations.

contracts/solidity/KeyManagement.sol
```diff
@@ -0,0 +1,290 @@
...
+    // Only for precompiled uups implementation in genesis file, need to be removed when upgrading the contract.
+    // This override is added because "immutable __self" in UUPSUpgradeable is not avaliable in precompiled contract.
+    function _checkProxy() internal view virtual override {
+        if (
+            address(this) == SELF || // Must be called through delegatecall
+            ERC1967Utils.getImplementation() != SELF // Must be called through an active proxy
+        ) {
+            revert UUPSUnauthorizedCallContext();
+        }
+    }
```
AnnaShaleva, line 49: Well, I think it may be removed right now because we use a stub for this contract in both Testnet/Mainnet, and for privnet testing this check is just not needed.

contracts/solidity/KeyManagement.sol
```diff
@@ -0,0 +1,290 @@
...
+
+    // Only for precompiled uups implementation in genesis file, need to be removed when upgrading the contract.
+    // This override is added because "immutable __self" in UUPSUpgradeable is not avaliable in precompiled contract.
+    function _checkNotDelegated() internal view virtual override {
+        if (address(this) != SELF) {
+            // Must not be called through delegatecall
+            revert UUPSUnauthorizedCallContext();
+        }
+    }
```
AnnaShaleva, line 58: As far as this one.

contracts/solidity/KeyManagement.sol
```diff
@@ -0,0 +1,290 @@
...
+        // record messages
+        if (reshareMsgs[targetHeight][index].length > 0)
+            revert Errors.MessageExists();
+        reshareMsgs[targetHeight][index] = messages;
+
+        // emit a event
+        emit Reshare(targetHeight, index, msg.sender);
+    }
+
+    function onPersist() external returns (bool) {
```
AnnaShaleva: We'll need a hardfork in the node code for this to start calls to KeyManagement's OnPersist from the specific height.

contracts/solidity/KeyManagement.sol
```diff
@@ -0,0 +1,290 @@
...
+                keccak256(output) !=
+                keccak256(pvss[bigfiOffset:bigfiOffset + BLS12381.G1_SIZE])
+            ) revert Errors.InvalidPVSS();
+            bigfiOffset += BLS12381.G1_SIZE;
+        }
+    }
+
+    function numberAndThreshold() public view returns (uint, uint) {
+        uint number = IGovernance(GOV).consensusSize();
+        return (number, (number * 2) / 3 + 1);
```
AnnaShaleva: I'd say that `n - (n-1)/3` is more usual for our code.

contracts/solidity/KeyManagement.sol
```diff
@@ -0,0 +1,290 @@
...
+    function reshare(bytes calldata pvss, bytes[] calldata messages) external {
+        // check period
+        uint currentEpochHeight = IGovernance(GOV).currentEpochStartHeight();
+        uint epochDuration = IGovernance(GOV).epochDuration();
+        uint periodDuration = IGovernance(GOV).sharePeriodDuration();
+        uint targetHeight = currentEpochHeight + epochDuration;
+        if (block.number < targetHeight - 4 * periodDuration)
+            revert Errors.PeriodNotStarted();
+        if (block.number >= targetHeight - 3 * periodDuration)
+            revert Errors.PeriodEnded();
```
AnnaShaleva: I see a problem in the sharing periods design: pending consensus is being recalculated and cached at block `targetHeight - 3 * periodDuration`. `reshare` happens during `[targetHeight - 4 * periodDuration; targetHeight - 3 * periodDuration)`, which means that starting from the block `targetHeight - 4 * periodDuration` candidates should *assume* given the *current state* of voting that they will be elected as CNs in the end of this voting period. However, votes distribution may easily be changed during `[targetHeight - 4 * periodDuration; targetHeight - 3 * periodDuration)` period. As a result, some of candidates that already shared data may be kicked of the top N candidates list.

contracts/solidity/KeyManagement.sol
```diff
@@ -0,0 +1,290 @@
...
+
+    function share(bytes calldata pvss, bytes[] calldata messages) external {
+        // check period
+        uint currentEpochHeight = IGovernance(GOV).currentEpochStartHeight();
+        uint epochDuration = IGovernance(GOV).epochDuration();
+        uint periodDuration = IGovernance(GOV).sharePeriodDuration();
+        uint targetHeight = currentEpochHeight + epochDuration;
+        if (block.number < targetHeight - periodDuration)
+            revert Errors.PeriodNotStarted();
+        if (block.number >= targetHeight) revert Errors.PeriodEnded();
```
AnnaShaleva: This code is common between all methods, only multipliers of `periodDuration` are different. It may be moved to a helper.

contracts/solidity/KeyManagement.sol
```diff
@@ -0,0 +1,290 @@
...
+        address[] memory currentConsensus = IGovernance(GOV)
+            .getCurrentConsensus();
+        for (uint i = 0; i < currentConsensus.length; i++) {
+            if (currentConsensus[i] == addr) {
+                return i + 1;
+            }
+        }
+        return 0;
+    }
+}
```
AnnaShaleva: We'll have troubles with indexes, because dBFT uses validators list sorted by public keys, whereas this code uses validators list sorted by votes.

contracts/solidity/Governance.sol
```diff
@@ -252,25 +260,43 @@ contract Governance is IGovernance, ReentrancyGuard, GovProxyUpgradeable {
...
         }
 
-        // compute and update consensus
-        if (length < consensusSize || totalVotes < voteTargetAmount) {
-            currentConsensus = standByValidators;
-        } else {
-            currentConsensus = _computeConsensus(candidates);
+        // set pending consensus as running, and update tag values
+        if (block.number >= currentEpochStartHeight + epochDuration) {
+            currentEpochStartHeight = block.number;
```
AnnaShaleva: This condition looks strange a bit in this context because it may seem that `block.number` may be higher than `currentEpochStartHeight + epochDuration`.  Revert this condition and return if true:

```go
        if (block.number < currentEpochStartHeight + epochDuration) {
            return
        }
        // set pending consensus as running, and update tag values
        currentEpochStartHeight = block.number;
        ...
```

contracts/solidity/KeyManagement.sol
```diff
@@ -0,0 +1,290 @@
...
+
+        // emit a event
+        emit Reshare(targetHeight, index, msg.sender);
+    }
+
+    function onPersist() external returns (bool) {
+        uint currentEpochHeight = IGovernance(GOV).currentEpochStartHeight();
+        uint epochDuration = IGovernance(GOV).epochDuration();
+        uint targetHeight = currentEpochHeight + epochDuration;
+        if (block.number < targetHeight) return false;
```
AnnaShaleva: The order of OnPersist calls for Governance and KeyManagement is important. Governance is called firstly. But then the condition `block.number < targetHeight` is always true because `currentEpochHeight` is updated by Governance contract in OnPersist call in the start of every epoch.

contracts/solidity/KeyManagement.sol
```diff
@@ -0,0 +1,290 @@
...
+                128
+            )
+            mcopy(add(input, 160), add(g2One, 32), 256)
+            mcopy(add(input, 416), add(g1One, 32), 128)
+            calldatacopy(
+                add(input, 544),
+                add(add(pvss.offset, 128), mul(t, 128)),
+                256
+            )
+        }
```
AnnaShaleva: 🙀

contracts/solidity/libraries/BLS12381.sol
```diff
@@ -0,0 +1,117 @@
...
+
+    function g1One() internal pure returns(bytes memory) {
+        return _g1One;
+    }
+
+    function g2One() internal pure returns(bytes memory) {
+        return _g2One;
+    }
+
+    function g1Add(
```
AnnaShaleva, line 25: May be we can try to enable BLS12-381 precompiles starting from some new hardfok and use them instead of this code? We'll need a hardfork for OnPersist anyway.

contracts/hardhat.config.ts
txhsl: I know, `MCOPY` is only available in `Cancun`, if I can use this opcode then the contract can be quite simple in assembly blocks. If we are not planned to, then I need to update the codes.

And also, the BLS12381 I enabled is in `Prague`, which is a future fork for Ethereum, so we cannot use it as well.

So I mean, I'm testing a future code anyway, we can decide the environment.

contracts/solidity/Governance.sol
txhsl: `sharePeriodDuration` and `epochDuration` are independent, but `sharePeriodDuration` is much shorter.

The reason we use `3 * sharePeriodDuration` is we need 3 period of secret sharing.

- A reshare of the last epoch, from old committee to the new one;
- A recover for reshare, from old committee to the new one;
- A share of the coming epoch, among new committee.

Maybe we can merge them, how do you think?

txhsl: `pendingConsensus` is cached for next epoch, but the current validators are still working, we can lock the vote for a while. This PR is WIP, I forgot to add this to TODO list.

contracts/solidity/KeyManagement.sol
txhsl: I'm using genesis file for booting, so let me keep it until the BLS test gets finished.

contracts/solidity/KeyManagement.sol
txhsl: Will.

contracts/hardhat.config.ts
AnnaShaleva: OK, I'll try to implement #233 in the nearest future, let's see how it goes and decide about `MCOPY` afterwards.

contracts/solidity/KeyManagement.sol
txhsl: Yep, we need indexes in decryption, they determine the computation result. If the node code has an idea of its index, then we may delete this, the `index` variable in this contract seems not that necessary.

contracts/solidity/KeyManagement.sol
txhsl: Just some assembly codes, don't worry. :)

contracts/solidity/libraries/BLS12381.sol
txhsl: This is just a packaged interface for BLS12-381 precompiles, we need a hardfork to enable them anyway.

contracts/solidity/Governance.sol
AnnaShaleva: To me, it's better to merge independent processes into a single period and execute them in parallel. Because it seems to me that we have to lock voting process during sharing period (although let's think if we can avoid that), and voting lock isn't a good thing, so the less sharing period lasts, the better for us.

Can these processes be performed in parallel?

AnnaShaleva: > we can lock the vote for a while

It's the thing I'm afraid of. Even if that lock period is relatively small, it doesn't seem good for the users to suspend voting. I think we may still allow users to vote, but we need to correctly handle this "gap" in votes on the contract side. Have you considered the following:
1. Epoch 1 ends, voters reward distributed according to the current list of CNs, we have new pending consensus nodes.
2. Epoch 2 starts, old CNs still drive the network, whereas pending CNs perform resharing. Reward from Epoch 2 still distributed according to the old list of CNs and according to the old GasPerVote values.
3. Epoch 3 starts: if pending CNs are successful in resharing then they become the current consensus, and GasPerVote is updated to the values that were relevant by the end of Epoch 1. If pending CNs are not successful in resharing then old CNs still drive the network with old GasPerVote value.

So we have a "gap" of one epoch between old/new consensus replacement, and during this gap we use old GasPerVote values for reward distribution.

@roman-khimov, what do you think about voting process lock period?

contracts/solidity/KeyManagement.sol
AnnaShaleva: Yes, the node should be aware of indexes: it can take the current CNs from Governance contract, then sort them and find own address in this list, and here's the index. But Governance contract probably still have to validate this index/address relation somehow, do you think it's needed?

contracts/solidity/KeyManagement.sol
AnnaShaleva: Well, it looks like VM code and it's readable/understandable, but I have to review/test it somehow :D

contracts/solidity/KeyManagement.sol
txhsl: Good.

contracts/solidity/KeyManagement.sol
txhsl: Yes. :(

contracts/solidity/KeyManagement.sol
txhsl: No, we lock the election when the sharing starts, so the CN list gets fixed.

contracts/solidity/Governance.sol
txhsl: > So we have a "gap" of one epoch between old/new consensus replacement, and during this gap we use old GasPerVote values for reward distribution.

It is one of the options we have considered when coding the Governance, but it means a voter have to wait for 1 week
to start to benefit from the election.

txhsl: I would like to say yes, Governance only cares about the final sharing result. Only the period “recover of reshare” may need “reshare” to confirm a failure, about which index is absent and needs to recover.

Reduced from 4 to 2 now, I need some comments from @doubiliu.

contracts/solidity/KeyManagement.sol
txhsl: Added.

contracts/solidity/KeyManagement.sol
txhsl: Well, it requires a lot of changes in tests, I may need a separated PR to that. Ref #293.

contracts/solidity/KeyManagement.sol
txhsl: I see, updated.

contracts/solidity/Governance.sol
txhsl: 🤔Well, let me change the style.

contracts/solidity/Governance.sol
txhsl: If we want to remove this locking period, then we need a solution to allow pending CNs to send their DKG messages to VM directly instead of through transactions.

AnnaShaleva: > but it means a voter have to wait for 1 week to start to benefit from the election.

That's true, but at the same time we may consider reducing epoch duration.

> to allow pending CNs to send their DKG messages to VM directly instead of through transactions.

Could you please clarify why do we need that? I think we can proceed with standard transactions. Pending CNs may use the next epoch to submit their DKG-related data to the contract, and by the end of next epoch a decision is taken whether everything is correct and pending CNs should become current CNs.

txhsl: > That's true, but at the same time we may consider reducing epoch duration.

We may reduce the epoch duration, e.g. to even 21 blocks, but I have to say the DKG transaction is expensive, cost at least `1,000,000` gas in EVM.

> Could you please clarify why do we need that?

I'm just talking about how to prevent this locking period, which is ensuring the DKG messages can be uploaded without any pending and incident.

contracts/hardhat.config.ts
AnnaShaleva: >the BLS12381 I enabled is in Prague

Regarding this, what do you think about enabling it in `NeoXAMEV` fork? It's introduced in #321.

contracts/solidity/Governance.sol
AnnaShaleva: Based on our Discord poll seems that we agree on voting lock period, so let's stick to this path, it's much simpler.

contracts/hardhat.config.ts
txhsl: Glad to see, but we may need related codes in `core/vm` to enable them.

I can update in this PR. never mind.

AnnaShaleva: Let's try it, then we won't be bound to Cancun.

contracts/solidity/KeyManagement.sol
```diff
@@ -0,0 +1,280 @@
...
+import {IGovernance} from "./interfaces/IGovernance.sol";
+import {IKeyManagement} from "./interfaces/IKeyManagement.sol";
+import {ERC1967Utils, GovProxyUpgradeable} from "./base/GovProxyUpgradeable.sol";
+
+contract KeyManagement is GovProxyUpgradeable, IKeyManagement {
+    address public constant SELF = 0x1212100000000000000000000000000000000008;
+    // governance contact
+    address public constant GOV = 0x1212000000000000000000000000000000000001;
+    // dkg enabled block number
+    uint public constant enabledNumber = type(uint).max;
```
AnnaShaleva: I think we can safely go without `enabledNumber`, because system won't call `OnPersist` before NeoXAMEV fork anyway, the contract can rely on that.

contracts/solidity/KeyManagement.sol
```diff
@@ -0,0 +1,280 @@
...
+        // emit a event
+        emit Reshare(targetHeight, index, msg.sender);
+    }
+
+    function onPersist() external returns (bool) {
+        // return true before dkg enabled
+        if (block.number < enabledNumber) return true;
+
+        uint currentEpochHeight = IGovernance(GOV).currentEpochStartHeight();
+        uint n = IGovernance(GOV).consensusSize();
```
AnnaShaleva: We need to add a check that only system caller is allowed to call `onPersist`, like for Governance.

contracts/solidity/KeyManagement.sol
```diff
@@ -0,0 +1,280 @@
...
+        // check index
+        uint index = indexOfResharing(msg.sender);
+        if (index < 1) revert Errors.NotShareMember();
+
+        // check absent secret index & record messages
+        uint n = IGovernance(GOV).consensusSize();
+        for (uint i = 0; i < idxs.length; i++) {
+            if (idxs[i] > n || idxs[i] == 0) revert Errors.IndexOutOfRange();
+            if (reshareMsgs[targetHeight][idxs[i]].length > 0)
+                revert Errors.NoNeedForRecover();
```
AnnaShaleva, line 142: Is it really an error? Recover here is not needed because everything is collected during Share period, so probably we should just return here.

Otherwise, can consensus nodes check from the outside whether recover is needed, i.e. do they have access to `reshareMsgs`? May be create some `isRecoverRequired` helper in the contract.

contracts/solidity/KeyManagement.sol
```diff
@@ -0,0 +1,280 @@
...
+    // height=>index=>index
+    mapping(uint => mapping(uint => bytes)) public rpvsses;
+    mapping(uint => mapping(uint => bytes)) public spvsses;
+    // bigA0 for verification and key generation
+    mapping(uint => mapping(uint => bytes)) public sharedPubs;
+    // global public keys
+    // NOTE: this not the direct key for keystore encryption, should use pk = globalPub * scaler,
+    // the scaler is used for speed up decryption, and not cool to be computed in contract.
+    // ref https://github.com/bane-labs/go-ethereum/blob/a07310bd9a3a117ae0876ad69bbe8b6ed624aaa5/core/antimev/util.go#L27
+    mapping(uint => bytes) public globalPubs;
```
AnnaShaleva, line 38: Could you please clarify where should users get the `scaler`?

contracts/test/KeyManagement.ts
```diff
@@ -0,0 +1,6 @@
+import { ethers } from "hardhat";
+import { expect } from "chai";
+
+describe("KeyManagement", function () {
+    // No runnable test in Hardhat on BLS12381 yet
```
AnnaShaleva, line 5: Can we move it to a separate issue? Eventually it'll be available.

contracts/solidity/KeyManagement.sol
```diff
@@ -0,0 +1,280 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.25;
+
+import {Errors} from "./libraries/Errors.sol";
+import {BLS12381} from "./libraries/BLS12381.sol";
+import {IGovernance} from "./interfaces/IGovernance.sol";
+import {IKeyManagement} from "./interfaces/IKeyManagement.sol";
+import {ERC1967Utils, GovProxyUpgradeable} from "./base/GovProxyUpgradeable.sol";
+
+contract KeyManagement is GovProxyUpgradeable, IKeyManagement {
```
AnnaShaleva, line 10: This contract seems to be almost ready. Can we add some basic documentation about it?

contracts/solidity/Governance.sol
```diff
@@ -113,7 +128,9 @@ contract Governance is IGovernance, ReentrancyGuard, GovProxyUpgradeable {
         return candidateList.values();
     }
 
-    function registerCandidate(uint shareRate) external payable {
+    function registerCandidate(
+        uint shareRate
+    ) external payable onlyElectionNotLocked {
```
AnnaShaleva, line 133: Can we adjust documentation wrt this change?

contracts/solidity/Governance.sol
```diff
@@ -252,24 +274,40 @@ contract Governance is IGovernance, ReentrancyGuard, GovProxyUpgradeable {
...
         uint length = candidates.length;
+        if (pendingConsensus.length == 0) {
+            // compute and update pending consensus
+            if (length < consensusSize || totalVotes < voteTargetAmount) {
+                pendingConsensus = standByValidators;
+            } else {
+                pendingConsensus = _computeConsensus(candidates);
+            }
+            electionLock = true;
+            emit Persist(pendingConsensus);
```
AnnaShaleva: I remember we discussed it, but could you please remind what happens if elected consensus is not changed by the end of this epoch? Should the same CNs perform yet another DKG iteration?

contracts/solidity/Governance.sol
```diff
@@ -252,24 +274,40 @@ contract Governance is IGovernance, ReentrancyGuard, GovProxyUpgradeable {
...
             ] = candidateGasPerVote[candidates[i]];
         }
-
-        // compute and update consensus
-        if (length < consensusSize || totalVotes < voteTargetAmount) {
-            currentConsensus = standByValidators;
-        } else {
-            currentConsensus = _computeConsensus(candidates);
+        // check if key generation succeeds, keep the same members if not
+        if (IKeyManagement(KEY_MANAGEMENT).onPersist()) {
```
AnnaShaleva: Can we rework this scheme a bit and use a system call for KeyManagement contract instead of a call from Governance? Then we may enable onPersist calls to KeyManagement starting from fork, and call KeyManagement's onPersist before Governance to process keys. Then Governance can rely on KeyManagement computations if successful.

What do you think?

AnnaShaleva (commented):
Voting lock period is handled properly, I didn't find bugs in it. Could you also please add these contracts to privnet after PR finalization?

contracts/solidity/KeyManagement.sol
AnnaShaleva: It's probably easier to handle this indexes problem on the node's side. Of course it's easier for dBFT to provide indexes that match alphabetic order (because all dBFT payloads contain index o the node), but node code may also provide indexes ordered by votes. So if it's easier for you not to implement this conversion on the contract side then we can move it to dBFT. You may create an issue for that.

contracts/solidity/KeyManagement.sol
txhsl: They can check from outside, this is just preventing an overwrite.

contracts/solidity/KeyManagement.sol
txhsl: Ref https://github.com/bane-labs/go-ethereum/blob/a07310bd9a3a117ae0876ad69bbe8b6ed624aaa5/core/antimev/util.go#L27, they can compute it with size and threshold. For a fixed pair of `n` and `t`, the `scaler` output is the same.

contracts/test/KeyManagement.ts
txhsl: OK.

contracts/solidity/Governance.sol
txhsl: No, with the current code, they will have to use the old one, should sounds like the current epoch gets doubled.

contracts/solidity/KeyManagement.sol
txhsl: I see.

contracts/solidity/Governance.sol
txhsl: Yes, it seems OK, but I think there may have to be invocation from Governance to KeyManagement, so that it can decide whether to change the validators to the pending ones.

contracts/solidity/Governance.sol
AnnaShaleva: Good, then it might be a good idea not to lock voting if consensus is not changed, because in general the less locks we have the better for users. Can we adjust this behaviour in the contract code?

contracts/solidity/Governance.sol
AnnaShaleva: >  may have to be invocation from Governance to KeyManagement, 

Yes, it's perfectly OK, it will be done via a separate designated method. But I don't insist, if the current scheme seems better for you then we may keep it, because it also works. The proposed change just allows us to get rid of the fork height tracking in the KeyManagement contract.

contracts/hardhat.config.ts
txhsl: > Let's try it, then we won't be bound to Cancun.

So how about the `MCOPY`? I think it will not be enabled in `NeoXAMEV`, right?

contracts/solidity/Governance.sol
txhsl: > Can we adjust this behaviour in the contract code?

We may need an array comparison here, but I think it is possible.

Added with tests.

contracts/solidity/KeyManagement.sol
txhsl: It's not bad to have this in contract, but I see, changing this number may make the compiled deploy code different and cause some matters in privnet settings.

contracts/solidity/Governance.sol
txhsl: I added a `dkgEnabled` parameter in Governance `onPersist`, which will make a difference in function behaviour.

contracts/solidity/KeyManagement.sol
txhsl: Ref #323.

contracts/test/KeyManagement.ts
txhsl: Ref #324.

contracts/solidity/Governance.sol
```diff
@@ -247,29 +269,55 @@ contract Governance is IGovernance, ReentrancyGuard, GovProxyUpgradeable {
...
+            } else {
+                pendingConsensus = _computeConsensus(candidates);
+            }
+            // lock election if consensus members are going to change
+            if (dkgEnabled) {
+                for (uint i = 0; i < consensusSize; i++) {
+                    bool notChanged;
+                    for (uint j = 0; j < consensusSize; j++) {
+                        if (pendingConsensus[i] == currentConsensus[j])
+                            notChanged = true;
```
chenquanyu: Here and below, `break` should be used to jump out of the loop.

contracts/hardhat.config.ts
AnnaShaleva: Yes, right, let's enable it in this fork.

contracts/README.md
```diff
@@ -174,12 +170,40 @@ Different from [GovProxyAdmin](https://github.com/bane-labs/go-ethereum/blob/ban
...
 
-## System contract stubs
+## KeyManagement
 
-[Stub](https://github.com/bane-labs/go-ethereum/blob/bane-main/contracts/solidity/Stub.sol) is reserved system contract implementation that has pre-assigned addresses in the genesis allocations (Stup0-Stub9 Proxies). Once designated role for the stub contract is created, its code will be updated correspondingly to serve the needs of the Neo X chain.
+[KeyManagement](https://github.com/bane-labs/go-ethereum/blob/bane-main/contracts/solidity/KeyManagement.sol) is a system contract assigned as the Neo X Distributed Key Generation (DKG) contract. This contract manages anti-MEV related cryptography operations needed for consensus nodes to participate in the Envelope transactions processing and block signing.
+
+There are 3 different generation periods to participant, reshare, share and recover. Before receiving any sharing messages from this contract, it is necessary to register your message encryption key through `registerMessageKey(...)`, this key should be a Secp256k1 public key which is generated by the antimev module of Neo X node.
```
AnnaShaleva, line 177: >  which is generated by the antimev module of Neo X node.

Why do we require antimev module to generate this key? I thought it should be generated by CN operator, just some random Secp256k1 key.

contracts/solidity/Governance.sol
```diff
@@ -247,29 +269,55 @@ contract Governance is IGovernance, ReentrancyGuard, GovProxyUpgradeable {
         else return _computeReward(voter, votedCandidate);
     }
 
-    function onPersist() external {
+    function onPersist(bool dkgEnabled) external {
```
AnnaShaleva: Oh, now I see there will be a problem with the contract update. The problem is: Governance contract update is performed manually, it's not bind to any hardfork. Consensus engine doesn't not know when exactly contract will be updated, and hence it doesn't know when to start to pass the parameter to `onPersist`. We need to think how to avoid `onPersist` signature modification and at the same time enable DKG-related logic in Governance contract starting from `NeoXAMEV`. But at the same time I'd still like to avoid storing `NeoXAMEV` anabling height in the contract itself.

contracts/solidity/Governance.sol
```diff
@@ -247,29 +269,55 @@ contract Governance is IGovernance, ReentrancyGuard, GovProxyUpgradeable {
...
             ] = candidateGasPerVote[candidates[i]];
         }
-
-        // compute and update consensus
-        if (length < consensusSize || totalVotes < voteTargetAmount) {
-            currentConsensus = standByValidators;
-        } else {
-            currentConsensus = _computeConsensus(candidates);
+        // check if key generation succeeds, keep the same members if not
+        if (!dkgEnabled || IKeyManagement(KEY_MANAGEMENT).onPersist()) {
```
AnnaShaleva: So you think that it's better to call KeyManagement's `onPersist` from Governance (not from the system)?

contracts/solidity/Governance.sol
AnnaShaleva: To me, the possible solution is to have two `onPersist` methods: one is old (without arguments), and another one is something like `onPersistV1` (also without arguments, but is called after `NeoXAMEV` fork and contains DKG-related code). What do you think?

txhsl: You're right, I see the problem.

contracts/README.md
txhsl: 🤔I think it's like an initialization, the CN operator creates an antimev keystore through command line, and a new random Secp256k1 key will be generated and appear in the storage.

So I say it is the antimev module that generates this key.

contracts/solidity/Governance.sol
txhsl: A good suggestion to save gas.

contracts/solidity/Governance.sol
txhsl: Yes, but now I probably know about your ideas. Since we have to change Governance `onPersist` and its handling in `consensus/dbft`, including checking the height and providing some parameters (maybe not) to a new `onPersist`, I can make things a bit different.

txhsl: How about now, I added a `isCurrentRoundReady` for dkg check, and now we have two separated `onPersist` functions. The only thing is KeyManagement’s `onPersist` should be called before Governanace `onPersistV2`.

contracts/solidity/Governance.sol
doubiliu: I don't support the reduction from 4 to 2. The purpose of reshare itself is to avoid the inability to decrypt old encrypted transactions due to consensus node switching. If reduced to 2 stages, the fault tolerance of this function cannot be guaranteed. We must admit that a distributed system cannot guarantee that all nodes will work 100% normally for a long time. Once a few of these nodes fail, do we still allow decryption functionality (for old transactions) to be hit or miss? This is chaos. In addition, I think the current design is only a little complicated in terms of process and will not affect efficiency.

txhsl: > I don't support the reduction from 4 to 2. The purpose of reshare itself is to avoid the inability to decrypt old encrypted transactions due to consensus node switching. If reduced to 2 stages, the fault tolerance of this function cannot be guaranteed.

It is the process of share and reshare becomes parallel, and the successful condition of verification is still both share and reshare receive `n` proper submissions. So the decryption functionality for old transactions still gets guaranteed.

contracts/hardhat.config.ts
AnnaShaleva: Sorry for misleading response, I just noticed that I read the question wrongly.

> I think it will not be enabled in NeoXAMEV, right?

So my original idea was to enable MCOPY in NeoXAMEV. We already discussed it, but let this comment be here for the record. It is implemented in #328.

contracts/solidity/Governance.sol
AnnaShaleva: Yes, the current version is good, it's exactly what I'd like to see. I implemented VM-level OnPersist switch in #328. However, I just discovered that I can't perform two contract invocations in a single script, and hence, we have to do it on the Governance contract side. So Shili, could you please adjust Governance's OnPersistV2 method so that it calls KeyManagement's OnPersist before any other code?

txhsl: > So Shili, could you please adjust Governance's OnPersistV2 method so that it calls KeyManagement's OnPersist before any other code?

If we go back to do that in Governance contract, a parameter about the fork height becomes necessary again. Why not call VM twice, the first on KeyManagement and the second on Governance?

AnnaShaleva: > Why not call VM twice, the first on KeyManagement and the second on Governance?

Let me try.

AnnaShaleva: Yes, you're right, it's actually possible, somehow I didn't even think about two sequential runs of VM, will create a PR for that.

PR #316, opened 2024-09-05
antimev: add signature cache config in txpool.

chenquanyu:
This is a special node configuration. If `--txpool.signaturecache` is set, all local transactions (except envelope tx) sent to this node will be stored in an encrypted list in `LegacyPool`. Only transactions with transaction `Nonce` equal to the `PendingNonce` and sufficient balance will be accepted by the encrypted list, and new transactions will always replace old transactions with the same `Nonce` and `From`. If the transaction passes the verification of txpool, the node will return a special error `transaction signature cached`. The cached tx can be obtained through the `eth_getEncryptedTransaction` rpc interface, the parameters are nonce and metamask signature.
For example, get nonce 0 transaction from 0x74f4EFFb0B538BAec703346b03B6d9292f53A4CD
`{
    "jsonrpc": "2.0",
    "method": "eth_getEncryptedTransaction",
    "params": ["0x0", 
   "0x20409b5aae34cc6bf83cb537c1e314f4847a21b5f0d353ee9baa581a287f63242a0dad11e7af43038f5fb970f30820f76d5b9742a87feeeb23255e050ff510f51b"
    ],
    "id": 1
}`


core/txpool/legacypool/legacypool.go
```diff
@@ -55,12 +57,18 @@ const (
...
+	sigMaxNum = 1024 * 1024
 )
 
 var (
 	// ErrTxPoolOverflow is returned if the transaction pool is full and can't accept
 	// another remote transaction.
 	ErrTxPoolOverflow = errors.New("txpool is full")
+	// ErrTxPoolSigCached is returned if the transaction signature is cached successfully.
+	// We use an error return here because we don't want the nonce to change after the signature is cached.
+	ErrTxPoolSigCached = errors.New("transaction signature cached")
```
AnnaShaleva: Is Metamask supposed to handle this error?

core/txpool/legacypool/legacypool.go
```diff
@@ -740,6 +764,32 @@ func (pool *LegacyPool) add(tx *types.Transaction, local bool) (replaced bool, e
...
+		if tx.Type() == types.LegacyTxType {
+			V = V.Sub(V, big.NewInt(27)).Mod(V, common.Big2)
+		}
+
+		sig[64] = byte(V.Uint64())
+		pool.signatureCache.Add(hash, sig)
+		err := ErrTxPoolSigCached
+		log.Trace(err.Error(), "hash", hash)
+		return false, err
+	}
```
AnnaShaleva: This code here leads to the fact that legacypool is always empty, so this node literally caches only transactions signatures without transactions themselves. Could you please clarify the purpose of such node in our anti-MEV scheme?

I thought that we need a special node that will encrypt all received transactions.

core/txpool/legacypool/legacypool.go
```diff
@@ -740,6 +764,32 @@ func (pool *LegacyPool) add(tx *types.Transaction, local bool) (replaced bool, e
...
 	from, _ := types.Sender(pool.signer, tx)
 
+	// If signaturescache config is true and from local, then just cache the tx and return an special error for signature cache
+	if pool.config.SignaturesCache && local {
+		if len(pool.senderMap[from]) >= int(pool.config.AccountQueue) {
+			oldHash := pool.senderMap[from][0]
+			pool.signatureCache.Remove(oldHash)
+			pool.senderMap[from] = pool.senderMap[from][1:]
+		}
+		pool.senderMap[from] = append(pool.senderMap[from], hash)
```
AnnaShaleva: Have you considered DOS possibility for such "caching" nodes? Any malicious sender is able to flood the node with dozens of transactions for cheap.

AnnaShaleva (commented):
I need to know the whole idea before the review.

core/txpool/legacypool/legacypool.go
chenquanyu: The transaction pool here does not accept `local` transactions, but the nodes will still synchronize transactions to keep the pending nonce correct. Our previous plan was to encrypt transactions on the front end, so only transaction signatures are cached here.

core/txpool/legacypool/legacypool.go
chenquanyu: Yes, we limit the same account to cache only 64 transaction signatures, and the node's RPC interface also needs to limit the access frequency according to the IP address.

core/txpool/legacypool/legacypool.go
txhsl: Metamask will handle this as some "RPC error" only.

core/txpool/legacypool/legacypool.go
AnnaShaleva: Could you please clarify, do I understand correctly:
1. User switches Metamask to a "special" node
2. Front-end sends transaction to Metamask for signing
3. Metamask signs transaction and sends it to the selected "special" node
4. "Special" node caches transaction signature and returns an error to Metamask
5. Metamask returns an error to front-end
6. Front-end receives error, recognizes it as ErrTxPoolSigCached, fetches signature from the special node, encrypts the signed transaction and inserts it into some new Envelope.
7. Front-end asks Metamask to sign Envelope and send it to some "standard" node

core/txpool/legacypool/legacypool.go
AnnaShaleva: Got it, it's restricted by `pool.config.AccountQueue`. But due to the eviction policy (`pool.senderMap` is FIFO in fact), an account is allowed to continiousely send transactions to the node, keeping the node busy with verification. Transactions will replace each other, and neither of them will be accepted to the real chain.

> node's RPC interface also needs to limit the access frequency according to the IP address.

Should it be done by some external rate limiter tool? To be honest, I'd prefer to avoid that as mach as possible, because Ethereum node itself has a good existing prioritization rules in the mempool. Why not to collect transactions (not just signatures) and apply standard mempool ordering rules (by fees)? This way DoS attack is less possible because account should have enough funds available to cover these transactions. And transactions signatures still may be retrieved by tx hash.

core/txpool/legacypool/legacypool.go
```diff
@@ -740,6 +765,32 @@ func (pool *LegacyPool) add(tx *types.Transaction, local bool) (replaced bool, e
...
```
AnnaShaleva: Also, current implementation doesn't care about how to clear `pool.senderMap`. It will grow through the node's lifetime and eventually node will suffer from OOM, this scheme is vulnerable to DoS.

It seems to me that using standard mempool prioritization rules and keeping full Encrypted transactions in pool allows to avoid that.

core/txpool/legacypool/legacypool.go
chenquanyu: Yes, this is the process we planned, but whether the front end can get the original error message we return still needs to be tested.

core/txpool/legacypool/legacypool.go
chenquanyu: I think using price-based sorting rules cannot avoid the problem of frequent verification requests, because the attacker can still continue to send transactions with the same nonce and different prices, and these transactions will not be on the chain without gas costs. And this will lead to the situation that the attacker can use the gas held to monopolize the cache pool.

chenquanyu: We may not need an external rate limiter tool because the public node rpc interface has a rate limit at the nginx level.

core/txpool/legacypool/legacypool.go
chenquanyu: Yes, this is a problem. We may need to store the transaction's from and signature and clean up pool.senderMap when the cache is full.

core/txpool/legacypool/legacypool.go
txhsl: Nginx is not always there, we may need an inside mechanism to ensure security.

chenquanyu: In fact, all rpc interfaces may be subject to DoS attacks, so frequency limiting is a necessary protection measure for public rpc nodes.

AnnaShaleva: > using price-based sorting rules cannot avoid the problem of frequent verification requests

But price-based sorting at least will be as effective against DoS as standard Eth nodes. And if we keep all submitted transactions in the node’s pool then we don’t have to build any additional logic for signatures management. This node just has to turn off transactions resharing functionality. Also, standard transactions eviction mechanism can be used to remove from pool those transactions that already decrypted and accepted in block.

But from the other hand, if we keep whole transactions in the node’s pool, then anyone is able to request “private” transaction from this node (if this node is public).

I think we need @roman-khimov’s opinion on this topic, because I’d avoid pool modifications as much as possible.

core/txpool/legacypool/legacypool.go
chenquanyu: Fixed.

core/txpool/legacypool/legacypool.go
roman-khimov: I'd expect this cache to behave similar to mempool which will not just help avoid DoS, but also help ensure the _correctness_ of transaction set we have in memory. Because nonces should still be checked. And GAS balance should also be checked. Otherwise you can accept a number of incorrect transactions or transactions that can't be paid for. They can even be packed into envelopes, but then this will just lead to failure to include unencrypted transaction into a block instead of rejecting the transaction right at the submission.

We can create another _standard_ pool for these transactions (not shared to other nodes) and push transactions into it according to standard node rules.

chenquanyu: @roman-khimov @AnnaShaleva Regarding the new `standard pool`, do you think it should inherit `LegacyPool` and copy all the transactions in it? Or add some verification methods in `LegacyPool`, and the new memory pool contains the current `LegacyPool` instance, which may cause some duplicate code.

chenquanyu: @AnnaShaleva @roman-khimov  Now we have an encrypted list like pending/queue list in `LegacyPool`, and the transactions in the encrypted list can only be obtained by the hash value in the following APIs `Get(hash common.Hash)` and `GetSignature(txHash common.Hash)`.. I plan to add a method to obtain encrypted transaction data to replace these two external interfaces, but only the miner node has antimev keystore. Can you help review the current entry and cleanup rules for encrypted transactions?

antimev/envelope.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/crypto/tpke"
+)
+
+var (
+	// encryptedDataPrefix is the prefix of Envelope transaction's data. It's used to
+	// distinguish simple transactions that have GovernanceRewardProxy contract as a
+	// receiver from Envelope transactions carrying encrypted information inside. In
+	// future this prefix may be used for encrypted content versioning.
+	EncryptedDataPrefix = []byte{0xff, 0xff, 0xff, 0xff}
```
AnnaShaleva, line 17: It's good that you made it public and moved to a separate package. Please, reuse this variable in TestGenerateEncryptedTx.

core/txpool/legacypool/legacypool.go
```diff
@@ -234,12 +249,13 @@ type LegacyPool struct {
...
 
-	reserve txpool.AddressReserver       // Address reserver to ensure exclusivity across subpools
-	pending map[common.Address]*list     // All currently processable transactions
-	queue   map[common.Address]*list     // Queued but non-processable transactions
-	beats   map[common.Address]time.Time // Last heartbeat from each known account
-	all     *lookup                      // All transactions to allow lookups
-	priced  *pricedList                  // All transactions sorted by price
+	reserve   txpool.AddressReserver       // Address reserver to ensure exclusivity across subpools
+	pending   map[common.Address]*list     // All currently processable transactions
+	encrypted map[common.Address]*list     // All currently processable encrypted transactions
```
AnnaShaleva, line 254: The case is we don't even need this extra `encrypted` map. We already have `pending`. So if `SignaturesCache` is set to `true`, then it means that every transaction stored in `pending` is an encrypted transaction. We need to reuse as much functionality from existing legacy pool as possible, and reusing `pending` map will help us. The usual sorting/prioritization rules will be applied to encrypted transactions in this case.

core/txpool/legacypool/legacypool.go
```diff
@@ -828,6 +868,55 @@ func (pool *LegacyPool) add(tx *types.Transaction, local bool) (replaced bool, e
 		}
 	}
 
+	// If signaturescache config is true and from local, then just cache the tx and return an special error for signature cache
+	if pool.config.SignaturesCache && local && !antimev.IsEnvelope(tx) {
```
AnnaShaleva, line 872: OK, the previous comment is relevant only if the node treats *all* incoming transactions as potentially encrypted. And here I see that it's not the case, and Envelopes are still maintained as a usual transactions in the `pending` map. If it's done intentionally, then this `if pool.config.SignaturesCache && local && !antimev.IsEnvelope(tx)` switch should be performed at the upper level (upper than mempool, I mean). As I see it:
1. Node with `txpool.signaturecache` flag set to `true` should maintain literally *two* mempools (two instances of `LegacyPool`):
    * Mempool1 is standard (with `SignaturesCache: false`), for transactions that should be relayed to the network. This mempool in fact will maintain only ready-to-use Envelopes transactions and relay them. These Envelopes transactions are maintained in the `pending` map.
    * Mempool2 is for encrypted transactions (with `SignaturesCache: true`). This mempool collects all trasactions except Envelopes and maintains them *in the `pending` map as far*. Then all copy-pasted code that currently maintains `encrypted` map will be removed (along with the `encrypted` map itself).
2. `pool.config.SignaturesCache` may enable additional logic on transactions sorting/validation/eviction. Also, `pool.config.SignaturesCache` should be used in order to prohibit transaction's relaying in case if `pool.config.SignaturesCache` is true, and to prohibit exposing transactions from `pending` map to the outside world via mempool APIs. Also, `pool.config.SignaturesCache` should be used to properly handle `GetSignature` API call.
3. On new transaction arriving, the node (at the upper level) should check if transaction is `antimev.IsEnvelope`. If it's an envelope, then the node should push the transaction to Mempool1. If it's not an envelope, then the node should push the transaction to Mempool2.

That's it. Mempool1 and Mempool2 are just included into `TxPool.subpools`.

AnnaShaleva (commented):
Much better than the previous implementation, but there's one thing that I'd like to improve at the architecture level (see the last reviewcomment). Let's firstly resolve it and finalize the design, and after that I'll review more precisely.

core/txpool/legacypool/legacypool.go
chenquanyu: No, pending transactions will come from other nodes, and it will also change the pending nonce.

core/txpool/legacypool/legacypool.go
chenquanyu: We added `!antimev.IsEnvelope(tx)` to reduce the switching of user nodes. When users switch to the encryption node, they can directly use this node to construct and send encrypted transactions.

txhsl: Yes, since we have had an internal discussion several days ago, it is possible to allow users to use amev service based on only one node.

- When a normal transaction arrives, this node caches it instead of adding it to the legacy pool;
- When an envelope transaction arrives, this node adds it to the legacy pool instead of caching it.

Then, users will not have to switch from a normal node to this special node and switch back. As far as we investigated, Metamask also doesn't support this kind of switching request. We can add a new network, but can't change it if there is already a network with the same `networkId`.

AnnaShaleva: > allow users to use amev service based on only one node.

It's OK, and to me the proposed subpool split scheme can also provide the same guarantees. But OK, let's proceed with the implemented solution and have https://github.com/bane-labs/go-ethereum/issues/372 for further investigation/development.

chenquanyu:
After testing, we found that sending transactions through metamask cannot get the original error message returned by rpc. Therefore, we discussed and decided to add a new `eth_getEncryptedTransaction` rpc method to use metamask's signature on the transaction nonce as a parameter to get the signed transaction.

internal/ethapi/api.go
```diff
@@ -1746,6 +1746,39 @@ func (s *TransactionAPI) GetRawTransactionByHash(ctx context.Context, hash commo
...
+func (s *TransactionAPI) GetEncryptedTransaction(ctx context.Context, Nonce hexutil.Uint64, signature hexutil.Bytes) (hexutil.Bytes, error) {
+	sender, err := recoverSignerAddress(Nonce, signature)
+	if err != nil {
+		return nil, err
+	}
+	if tx := s.b.GetEncryptedTransaction(uint64(Nonce), *sender); tx != nil {
+		return tx.MarshalBinary()
+	}
+	return nil, nil
+}
```
AnnaShaleva, line 1780: This API differs from the original intention. It seems that right now anyone can get an encrypted transaction from the node via RPC, is it true?

txhsl: Yes, but it requires the correct transaction `nonce` and a signature for that `nonce`.

AnnaShaleva: Good, then not a problem.

Issue #315, opened 2024-09-05
Single-node privnet is broken


AnnaShaleva:
#### System information

Commit hash : 2de874a672b6cdb44c190172188f592f78b77b46

#### Expected behaviour

 Single-node privnet is running normally.

#### Actual behaviour

Single-node privnet processes a single block and stales in awaiting. It's likely yet another consequence of sealing proposal awaiting refactoring.
```
anna@kiwi:~/Documents/GitProjects/bane-labs/go-ethereum$ cat privnet/single/node1/geth_node.log 
INFO [09-05|09:35:16.757] Enabling metrics collection
INFO [09-05|09:35:16.758] Maximum peer count                       ETH=50 total=50
INFO [09-05|09:35:16.759] Smartcard socket not found, disabling    err="stat /run/pcscd/pcscd.comm: no such file or directory"
INFO [09-05|09:35:16.761] Set global gas cap                       cap=50,000,000
INFO [09-05|09:35:16.761] Initializing the KZG library             backend=gokzg
INFO [09-05|09:35:16.814] Allocated trie memory caches             clean=154.00MiB dirty=256.00MiB
INFO [09-05|09:35:16.814] Using pebble as the backing database
INFO [09-05|09:35:16.814] Allocated cache and file handles         database=/home/anna/Documents/GitProjects/bane-labs/go-ethereum/privnet/single/node1/geth/chaindata cache=512.00MiB handles=524,288
INFO [09-05|09:35:16.832] Opened ancient database                  database=/home/anna/Documents/GitProjects/bane-labs/go-ethereum/privnet/single/node1/geth/chaindata/ancient/chain readonly=false
INFO [09-05|09:35:16.832] State scheme set to already existing     scheme=hash
INFO [09-05|09:35:16.832] Initialising Ethereum protocol           network=2,312,051,126 dbversion=<nil>
INFO [09-05|09:35:16.834] 
INFO [09-05|09:35:16.834] ---------------------------------------------------------------------------------------------------------------------------------------------------------
INFO [09-05|09:35:16.834] Chain ID:  2312051126 (unknown)
INFO [09-05|09:35:16.834] Consensus: dBFT (proof-of-authority)
INFO [09-05|09:35:16.834] 
INFO [09-05|09:35:16.834] Pre-Merge hard forks (block based):
[...]

PR #326, opened 2024-09-26
antimev: move to root and add proper keystore persistence

### LINKED ISSUE #310: Add amev key local persistence
Ref https://github.com/bane-labs/go-ethereum/pull/301#discussion_r1729314256.

txhsl:
Close #310. Ready for review, four-node and seven-node privnets are working with new files and codes, except single-node privnet, ref #315.

- [x] Remove RLP encoding and decoding added by #287;
- [x] Move `core/antimev` to root, make things similar with `accounts`;
- [x] Apply eth2 wallet persistence to antimev keystore;
- [x] Update cli and startup related codes;
- [x] Fix unit tests;
- [x] Update privnet files;
- [x] Add keystore operation commands;
- [x] Update node document about setup.

Currently, the PVSS and other public data is also persisted, should remove when contract is properly connected.


antimev/util.go
```diff
@@ -95,3 +102,178 @@ func abs(a int) int {
...
+type ObjType int
+
+const (
+	Regular ObjType = iota
+	Directory
+)
+
+func readWritePermissions() os.FileMode {
+	if runtime.GOOS == "windows" {
+		return 0666
```
roman-khimov: Why?

antimev/util.go
```diff
@@ -95,3 +102,178 @@ func abs(a int) int {
...
+	if info == nil {
+		return false, errors.New("file info is nil")
+	}
+
+	isDir := info.IsDir()
+
+	return objType == Directory && isDir || objType == Regular && !isDir, nil
+}
+
+func expandPath(p string) (string, error) {
```
roman-khimov: Do you really need it?

roman-khimov (commented):
So the idea is to store keys in JSON-based keystore, right?

antimev/util.go
txhsl: Currently, for test usage, only because I'm coding on Windows.

I will remove this after UT implemented. In this PR, soon.

txhsl:
> So the idea is to store keys in JSON-based keystore, right?

Yes.

antimev/util.go
txhsl: Removed.

antimev/keygroup.go
```diff
@@ -42,92 +40,6 @@ type thresholdKeyHolder struct {
 	ethPubKey *ecies.PublicKey // The account public key
 }
 
-var (
-	_ rlp.Encoder = &thresholdKeyGroup{}
-	_ rlp.Decoder = &thresholdKeyGroup{}
-)
-
-// thresholdKeyGroupAux is an auxiliary structure for thresholdKeyGroup RLP marshalling.
-type thresholdKeyGroupAux struct {
```
AnnaShaleva, line 51: Belongs to the previous commit.

node/config.go
```diff
@@ -478,3 +487,23 @@ func (c *Config) GetKeyStoreDir() (string, bool, error) {
 
 	return keydir, isEphemeral, nil
 }
+
+func (c *Config) GetAntiMEVKeyStorePath() (string, error) {
```
AnnaShaleva, line 492: It would be nice to have a comment for exported method.

cmd/geth/main.go
```diff
@@ -431,6 +434,22 @@ func startNode(ctx *cli.Context, stack *node.Node, backend ethapi.Backend, isCon
...
+		case *dbft.DBFT:
+			bft = t
+		case *beacon.Beacon:
+			switch inner := t.InnerEngine().(type) {
+			case *dbft.DBFT:
+				bft = inner
+			}
+		}
+		if bft != nil {
+			unlockAntiMEVKeyStore(ctx, stack)
```
AnnaShaleva, line 453: An error may happen here. Return error from this method and fail the node startup if it's not nil.

antimev/keystore.go
```diff
@@ -0,0 +1,500 @@
...
+}
+
+// keyStoreAux is an auxiliary structure for KeyStore JSON marshalling.
+type keyStoreAux struct {
+	Size      int `json:"size"`      // The size of each key group
+	Threshold int `json:"threshold"` // The threshold of each key group
+	Scaler    int `json:"scaler"`    // The scaler to speed up computation, refer to crypto/tpke
+
+	Address   common.Address `json:"address"`     // Self account address
+	EthPrvKey string         `json:"eth_prv_key"` // Self account secret key
```
AnnaShaleva, line 393: Private key is stored in ecrypted form, right?

antimev/keystore_test.go
```diff
@@ -14,13 +16,13 @@ var size = 7
...
-	common.HexToAddress("0x01b517b301bb143476da35bb4a1399500d925514"),
-	common.HexToAddress("0x7976ad987d572377d39fb4bab86c80e08b6f8327"),
-	common.HexToAddress("0xd711da2d8c71a801fc351163337656f1321343a0"),
+	common.HexToAddress("0x74f4effb0b538baec703346b03b6d9292f53a4cd"),
+	common.HexToAddress("0x910ad1641b7125eff746accdca1f11148b22f472"),
+	common.HexToAddress("0xfef5f250af14df73f983caab7b1f5002189c42e0"),
+	common.HexToAddress("0xc51964013acbc6b271feecb0febd9e7a01202930"),
+	common.HexToAddress("0xc5bbd9652546bc96be3dec97a38ee335f7873dfa"),
+	common.HexToAddress("0x26f1794b81df2b832545b8b6bbca196b82e4feb1"),
+	common.HexToAddress("0x0b51369d02e47ee3f143391b837aa08c31aaa19b"),
```
AnnaShaleva: Please, make a comment that these are privnet addresses.

antimev/tpke_test.go
```diff
@@ -2,62 +2,29 @@ package antimev
...
-}
-
 func TestTPKE(t *testing.T) {
-	const saveKeystore = false
-
-	privnetCNAddresses := make([]common.Address, len(privnetCNs))
-	for i := range privnetCNAddresses {
-		privnetCNAddresses[i] = privnetCNs[i].addr
-	}
-
```
AnnaShaleva, line 51: Please, don't remove it or create another generator test instead. We need to be able to generate ready-to-use keystore for privnet and encrypt some message with it. It is required for testing, we should be able to reproduce keystore generation if something is changed.

Also, this test should encrypt some message and print it in the proper form to the user. We need to have at least some basic data to README to be able to immediately verify AMEV privnet setup. Also please, update the tail of https://github.com/bane-labs/go-ethereum/tree/bane-main/privnet#commands-in-javascript-console.

cmd/utils/flags.go
```diff
@@ -503,11 +503,15 @@ var (
...
+		Name:     "antimev.keystore",
 		Usage:    "Path to anti-MEV keystore",
-		Value:    "",
-		Category: flags.AMEVCategory,
+		Category: flags.AntiMEVCategory,
+	}
+	AntiMEVPasswordFlag = &cli.PathFlag{
+		Name:     "antimev.password",
+		Usage:    "Password file to unlock anti-MEV keystore",
+		Category: flags.AntiMEVCategory,
```
AnnaShaleva, line 514: Please, add a section to the node documentation on how to setup CN with anti-MEV keystore.

cmd/utils/flags.go
```diff
@@ -1390,6 +1409,9 @@ func SetNodeConfig(ctx *cli.Context, cfg *node.Config) {
 	if ctx.IsSet(KeyStoreDirFlag.Name) {
 		cfg.KeyStoreDir = ctx.String(KeyStoreDirFlag.Name)
 	}
+	if ctx.IsSet(AntiMEVKeyStoreFlag.Name) {
+		cfg.AntiMEVKeyStorePath = ctx.String(AntiMEVKeyStoreFlag.Name)
+	}
```
AnnaShaleva, line 1410: Don't we need to store password in config?

crypto/tpke/private_key.go
```diff
@@ -58,6 +30,21 @@ func NewPrivateKey(secretShares []*big.Int) *PrivateKey {
...
+}
+
+func (sk *PrivateKey) Decode(b []byte) (*PrivateKey, error) {
+	if len(b) != 32 {
+		return nil, ErrTPKEScalarDecoding
+	}
+	fe := new(fr.Element).SetBytes(b)
+	sk.fr = fe.BigInt(new(big.Int))
+	return sk, nil
+}
```
AnnaShaleva, line 46: Do we use unprotected private key encoding anywhere?

eth/backend.go
```diff
@@ -178,12 +178,7 @@ func New(stack *node.Node, config *ethconfig.Config) (*Ethereum, error) {
 		p2pServer:         stack.Server(),
 		shutdownTracker:   shutdowncheck.NewShutdownTracker(chainDb),
 	}
-	if len(config.AMEVKeystorePath) != 0 {
-		eth.antimevKeystore, err = antimev.LoadAMEVKeyStore(config.AMEVKeystorePath)
-		if err != nil {
-			return nil, fmt.Errorf("failed to load AMEV keystore: %w", err)
-		}
-	}
+	eth.antimevKeystore = stack.AntiMEVKeyStore()
```
AnnaShaleva, line 181: Handle error here, don't hide/process it in the handler.

node/config.go
```diff
@@ -76,6 +77,14 @@ type Config struct {
 	// is created by New and destroyed when the node is stopped.
 	KeyStoreDir string `toml:",omitempty"`
 
+	// AntiMEVKeyStorePath is the file path that contains antimev data. The directory can
+	// be specified as a relative path, in which case it is resolved relative to the
+	// current directory.
+	//
+	// If AntiMEVKeyStorePath is empty, the default location is DataDir with default file
+	// name.
+	AntiMEVKeyStorePath string `toml:",omitempty"`
```
AnnaShaleva, line 86: How about password file config?

AnnaShaleva: Path is useless without password anyway.

cmd/utils/flags.go
AnnaShaleva, line 514: I didn't see it in the PR, but we need to add a requirement: if `--mine` flag is specified, i.e. the node is CN and dBFT engine is used, then it should be explicitly required from the user to specify AMEV keystore with password.

antimev/keystore.go
txhsl: Yes.

antimev/tpke_test.go
txhsl: No, it is not good to produce files within tests without deleting them after execution, but the generator is not destroyed here.

While the keystore files are now realtime persisted, I write them to `t.TempDir()` instead. So change this dir setting, then these files will be written to another location and not be deleted.

> Also, this test should encrypt some message and print it in the proper form to the user.

I'm OK with this idea.

> Also please, update the tail of https://github.com/bane-labs/go-ethereum/tree/bane-main/privnet#commands-in-javascript-console.

I think this is something afterward, like exporting some APIs to JS console.

cmd/utils/flags.go
txhsl: It is not here but the same as `--password` flag solution.

https://github.com/bane-labs/go-ethereum/blob/7484c42b6d4686bdcf81bdb5f9671f52567bb78f/cmd/utils/flags.go#L1330

node/config.go
txhsl: I think we have, just refer to the startup process of miner account, the file path is first imported, but get unlocked with password latter.

cmd/geth/main.go
txhsl: I handle it the same as `unlockAccounts`, which `os.Exit(1)` inside directly and stops the startup.

cmd/utils/flags.go
txhsl: We have, ref https://github.com/bane-labs/go-ethereum/pull/326#discussion_r1791433169.

> if --mine flag is specified, i.e. the node is CN and dBFT engine is used, then it should be explicitly required from the user to specify AMEV keystore with password.

I'm not sure if we need another check, because `unlockAntiMEVKeyStore` will `utils.Fatalf` if the file is not found or the password is not correct, when `btf != nil`.

eth/backend.go
txhsl: No, no error exists here, the `antimevKeystore` won't be `nil`, it is something empty with only directory path, nothing is loaded.

crypto/tpke/private_key.go
txhsl: I want to export it in a proper format when writing to antimev keystore files.

We generate it and use as `big.Int`, but I expect a `bls12381.fr` in persistence, how do you think?

antimev/tpke_test.go
AnnaShaleva: > No, it is not good to produce files within tests without deleting them after execution

Unless it's a test for data generation. We commonly use this practice in NeoGo, take a look at https://github.com/nspcc-dev/neo-go/blob/0968c3a81fb034ffe95ede7667d4deb4954b4021/pkg/core/basic_chain_test.go#L37 and https://github.com/nspcc-dev/neo-go/blob/0968c3a81fb034ffe95ede7667d4deb4954b4021/pkg/wallet/regenerate_test.go#L91 as an example.

> So change this dir setting,

I have to type the new dir name - it's not a lot, but still it's more than changing `const saveKeystore` from `false` to `true`.

> I'm OK with this idea.

Then don't remove this code. Currently it's the only way for me (as for developer) to check that decryption works on privnet.

> I think this is something afterward, 

Again, currently I consider it as an instruction for developers, not for the end-users. I myself use this instruction to check that everything is OK. Otherwise we don't even have any documentation on how to submit encrypted transactions. And currently with new changes, this part of documentation is out-of-date.

crypto/tpke/private_key.go
AnnaShaleva: Both ways are OK, the main thing is that we shouldn't store it in unencrypted form.

eth/backend.go
AnnaShaleva: I'd like to ensure two things:
1. The user should be able to start seed node without anti-MEV keystore (some simple RPC that does not require keystore settings).
2. If the node is running as a CN, then we must check that anti-MEV keystore is provided and properly opened.

Could you add these checks to the node's code?

cmd/utils/flags.go
AnnaShaleva: > because unlockAntiMEVKeyStore will utils.Fatalf if the file is not found

From what I see, there's a case when it will silently skip keystore initialisation if both file/password are not provided, which should be a no-op in case of dBFT-enabled node:
https://github.com/bane-labs/go-ethereum/blob/7484c42b6d4686bdcf81bdb5f9671f52567bb78f/antimev/keystore.go#L435-L438

crypto/tpke/private_key.go
txhsl: All encrypted, don't worry.

antimev/tpke_test.go
txhsl: > Unless it's a test for data generation.

> I have to type the new dir name - it's not a lot, but still it's more than changing const saveKeystore from false to true.

I don't find any similar solutions in tests of Geth, so I still insist, let's do it in the same way.

> Then don't remove this code. Currently it's the only way for me (as for developer) to check that decryption works on privnet.

I will check.

OK, I get the idea.

> And currently with new changes, this part of documentation is out-of-date.

Updated in this PR as well, but seems not working. Let me check.

Well, the reason is the index order. Fixed, quite working now.

cmd/utils/flags.go
txhsl: > From what I see, there's a case when it will silently skip keystore initialisation if both file/password are not provided

There is a default file path, so it cannot be empty, but let me have some tests and find if we need a better check.

So these are the cases and results:
1. File doesn't exist (whether password provided or not) — `Fatal: Failed to find or load antimev keystore: no such file or directory`;
2. File exists but empty/invalid format (whether password provided or not) — `Fatal: Failed to find or load antimev keystore: unexpected end of JSON input`;
3. File is correctly provided but password not provided or wrong — `Fatal: Failed to find or load antimev keystore: could not decrypt keystore`;

However, the referred codes should get changed. Fixed.

eth/backend.go
txhsl: > I'd like to ensure two things:
> 
> 1. The user should be able to start seed node without anti-MEV keystore (some simple RPC that does not require keystore settings).
> 2. If the node is running as a CN, then we must check that anti-MEV keystore is provided and properly opened.
> 
> Could you add these checks to the node's code?

Well, it's exactly working like this. As a seed node, the `node.antimevKeystore` is only a structure with `path` and `password` configured, so as the `eth.antimevKeystore`. Only if before the miner starts, or we can say when the `--miner` is setted, the `unlockAntiMEVKeyStore` method will try to access and decrypt the specific file.

Please check https://github.com/bane-labs/go-ethereum/pull/326/files#diff-30437b401e56caf63af6a19560e47ff5c65bfe20970c044d5c497f2158729b3cR432, or have a try on four-node privnet node5 or seven-node privnet node8. They don't have an `antimev-keystore` file, but correctly working as seed nodes.

cmd/utils/flags.go
txhsl: Well, this needs some geth commands like `./geth --datadir ./nodes/node1 account new`, but I haven't implemented them in this PR.

Give me some time to do that.

txhsl: Done, I also updated the documentation.

txhsl:
@songb2. Need some tests.

antimev/keystore_test.go
```diff
@@ -13,14 +15,26 @@ import (
 var size = 7
 var threshold = 5
 
+// Here use the same address list as the seven-node privnet, sorted by address not index.
```
AnnaShaleva: Please, don't sort them manually; implement sorting in `init()` function, because you'll have to do it anyway if we ever update privnet addresses.

antimev/tpke_test.go
```diff
@@ -97,40 +65,24 @@ func TestTPKE(t *testing.T) {
...
-			ks := new(AMEVKeyStore)
-			fmt.Printf("%s (CN %d):\n\t%s\n", privnetCNs[i].addr, privnetCNs[i].i, hex.EncodeToString(ksBytes))
-			os.WriteFile(fmt.Sprintf("../../privnet/seven/node%d/amev_keystore.txt", privnetCNs[i].i), []byte(hex.EncodeToString(ksBytes)), os.ModePerm)
-
-			// Double-check to ensure keystore marshalling works correctly, and it's
-			// possible to decrypt message with deserialized keystore.
-			require.NoError(t, ks.FromBytes(ksBytes))
-			kss[i] = ks
-		}
-	}
```
AnnaShaleva, line 113: Just for the record, I'm still against of this change.

cmd/geth/antimevcmd.go
```diff
@@ -0,0 +1,207 @@
...
+			break
+		}
+	}
+	// All trials expended to unlock account, bail out
+	utils.Fatalf("Failed to unlock antimev keystore (%v)", err)
+	return nil
+}
+
+// antimevStatus prints the status of specified antimev keystore
+func antimevStatus(ctx *cli.Context) error {
```
AnnaShaleva, line 146: BTW, this command is extremely useful.

cmd/utils/flags.go
```diff
@@ -1321,6 +1325,25 @@ func MakePasswordList(ctx *cli.Context) []string {
 	return lines
 }
 
+// MakeAntiMEVPasswordList reads password lines from the file specified by the global
+// --antimev.password flag.
+func MakeAntiMEVPasswordList(ctx *cli.Context) []string {
```
AnnaShaleva, line 1317: I don't think we need a copy-past of MakePasswordList method. It's better to reuse existing code as much as possible, and the comment of this method may be adjusted to include `--antimev.password` flag.

eth/backend.go
```diff
@@ -510,6 +505,11 @@ func (s *Ethereum) StartMining() error {
 			}
 		}
 		if cli != nil || bft != nil {
+			dkgIdentity := s.antimevKeystore.Address()
+			if eb != dkgIdentity {
+				log.Error("Antimev keystore mismatch", "etherbase", eb, "keystore", dkgIdentity)
```
AnnaShaleva: > "etherbase"

It's actually not an etherbase in case of dBFT, it's just the CN address. And etherbase is our Governance contract address, the destination of mining reward.

But the check itself is correct, `eb` holds CN address.

AnnaShaleva (commented):
Some minor comments left, they are not critical. Otherwise LGTM, I agree to merge once testing is done by @songb2. From my side I tried new keystore init/inspect commands and checked the privnet.

eth/backend.go
txhsl: Yes, this message seems misleading.

songb2:
I've finished testing the PR. No issues were found.

Issue #323, opened 2024-09-14
Use Governance contract provided and ordered address list in dBFT


txhsl:
# Rationale

The crypto scheme of antimev requires an indexed validator list for decryption and signature aggregation. This list and its order can only be provided by Governance contract, otherwise we need to add a same sort algorithm in Government or feedback the sort result back to the contract, so that the DKG contract can perform the correct verification.

# Implementation

~~Directly use the validator list provided by Governance in dBFT, instead of sorting it again.~~

Implement an order conversation to the address list in dBFT when handling DKG.

Ref https://github.com/bane-labs/go-ethereum/pull/312#discussion_r1758546687.

PR #328, opened 2024-10-03
vm: move OnPersist, BLS precompiles and MCOPY under NeoXAMEV fork

### LINKED ISSUE #322: Remove `NeoXBurn` fork
# Rationale

`NeoXBurn` fork is enabled from 0 in T4/Mainnet, and T3 is dropped. This fork is kind of "transiition" and it's not needed anymore.

# Implementation

Remove `NeoXBurn` from the node configuration, move its logic out of the fork. It should be a default node behavior.

### LINKED ISSUE #327: Move `MCOPY` and BLS precompiles under `NeoXAMEV` fork
# Rationale

The reason is described in https://github.com/bane-labs/go-ethereum/pull/312#discussion_r1744909684.

# Implementation

Enable `MCOPY` and BLS precompiles starting from `NeoXAMEV`, exclude them from Cancun/Prague enabling code.

AnnaShaleva:
Required for #312. Close #322, close #327.

~I'm thinking about moving NeoXAMEV from block-based to timestamp-based, but there are some pitfalls, i.e. switching dBFT to anti-MEV version by using time-based trigger.~ NeoXAMEV is now post-merge hardfork, but at the same time it is block-based, nothing prevents us from it.

Requires @txhsl approval. @chenquanyu, @songb2, welcome to review&test.


roman-khimov:
> from block-based to timestamp-based

But why?

AnnaShaleva:
> But why?

Because it should be a post-merge hardfork, because we already have Shanghai enabled. And all post-merge hardforks are timestamp-based. But right now I’m checking is it possible to make post-merge hardfork to be a block-based.

roman-khimov:
> all post-merge hardforks are timestamp-based

Wow. Unexpected.

core/vm/jump_table.go
```diff
@@ -81,15 +82,20 @@ func validate(jt JumpTable) JumpTable {
 }
 
 func newCancunInstructionSet() JumpTable {
-	instructionSet := newShanghaiInstructionSet()
+	instructionSet := newNeoXAMEVInstructionSet()
 	enable4844(&instructionSet) // EIP-4844 (BLOBHASH opcode)
 	enable7516(&instructionSet) // EIP-7516 (BLOBBASEFEE opcode)
 	enable1153(&instructionSet) // EIP-1153 "Transient Storage"
-	enable5656(&instructionSet) // EIP-5656 (MCOPY opcode)
```
AnnaShaleva, line 88: A question to @txhsl: this commit enables only MCOPY instruction. However, to compile DKG contract we have to use Cancun hardhat compiler target, which means that some other post-Shanghai instructions may be emitted by compiler. Do you think we need the rest of Cancun instructions to be enabled in NeoXAMEV?

My answer is: we definitely don't need BLOBHASH and BLOBBASEFEE, because we anyway won't be able to process these opcodes in NeoX chains right now.  But I'm not sure about Transient Storage opcodes, because this feature may be used by some of our dependent contracts. But I'm not an expert, I'm just worried about TLOAD and TSTORE.

cmd/evm/internal/t8ntool/transition.go
```diff
@@ -197,12 +197,12 @@ func Transition(ctx *cli.Context) error {
 }
 
 func applyNeoXBurnChecks(env *stEnv, chainConfig *params.ChainConfig) error {
-	if !chainConfig.IsNeoXBurn(big.NewInt(int64(env.Number)), env.Timestamp) {
+	if chainConfig.DBFT == nil {
```
AnnaShaleva, line 200: @roman-khimov, I tried but a lot of tests need to be fixed, refactored or removed to make Policy-based behaviour commonly-used. I created #329 for that, are we OK with it? If no, then I can manage with tests, but it requires some time.

AnnaShaleva:
Ready for review.

core/vm/jump_table.go
roman-khimov: This also worries me from the compatibility perspective, can people compile contracts with Cancun-enabled compilers or not really?

cmd/evm/internal/t8ntool/transition.go
roman-khimov: Merging with red tests in never a good idea.

consensus/misc/eip1559/eip1559.go
```diff
@@ -57,6 +57,7 @@ func VerifyEIP1559Header(config *params.ChainConfig, parent, header *types.Heade
 		return fmt.Errorf("invalid baseFee: have %s, want %s, parentBaseFee %s, parentGasUsed %d",
 			header.BaseFee, expectedBaseFee, parent.BaseFee, parent.GasUsed)
 	}
+
```
roman-khimov: :eyes:

roman-khimov (commented):
OK code-wise, but needs to be tested against real contracts.

core/vm/jump_table.go
txhsl: > But I'm not sure about Transient Storage opcodes, because this feature may be used by some of our dependent contracts.

Transient storage is designed for some cases where a value in storage changes, but also changes back in the same function call, ref https://soliditylang.org/blog/2024/01/26/transient-storage. For the DKG contract, transient storage is not used, but in Governance, it might be used by `ReentrancyGuard` in the future. We can compile Governance under Shanghai to avoid it, and if we keep using `openzeppelin-contracts/release-v5.0`, they are not even imported in Solidity 0.8.20, just keep `KeyManagement` clean. But for other user contracts compiled under Cancun, I can't guarantee.

I also find some notes in Solidity 0.8.27 release announcement, which says it can be only used in low level codings.
> This release introduces support for transient storage variables into the parser. To clarify, the feature as a whole is not yet complete.

> The compiler will now accept the syntax for marking some variables as `transient` and let users generate transient storage layout. However, it is still not possible to generate bytecode for contracts using such variables.

> We plan to introduce high-level language support for transient storage gradually, over the next few releases. In particular, the next release will provide full support for transient state variables of value types. The support will be extended from there to cover more complex use cases.

AnnaShaleva: > can people compile contracts with Cancun-enabled compilers or not really?

No. And it's not our goal, we don't support Cancun yet, we only support Shanghai. Cancun compiler target allows to use BLOBHASH, BLOBBASEFEE and transient storage whereas we don't support it yet, it's a matter of https://github.com/bane-labs/go-ethereum/issues/189. But of course if contract don't use Cancun-specific instructions, then it can be compiled with any target.

> We can compile Governance under Shanghai to avoid it

Could you clarify, where exactly do we use MCOPY? Only in BLS precompiles, right? If so, then we're OK with the current version of PR.

cmd/evm/internal/t8ntool/transition.go
AnnaShaleva: Tests are green on the current version of PR, because I keep `if chainConfig.DBFT == nil` branches. So existing Geth tests work as expected.

consensus/misc/eip1559/eip1559.go
AnnaShaleva: Fixed!

core/vm/jump_table.go
txhsl: > Only in BLS precompiles, right?

Yes.

AnnaShaleva:
Good, @txhsl, seems that we agree on the PR content an implementation is finished. Please, review.

cmd/evm/internal/t8ntool/transition.go
roman-khimov: OK then.

core/vm/jump_table.go
roman-khimov: Just as an idea --- can we enable transient storage in XAMEV? What's required for it? Maybe it'll make life a bit easier for contract developers if we have it, even though we don't support all of Cancun (at least we could compile of our contracts with the same toolchain). Or maybe it'll create even more confusion since _most_ of contracts will likely work, but some (BLOBHASH, BLOBBASEFEE, I expect these to be less popular) won't.

AnnaShaleva: > can we enable transient storage in XAMEV?

We can, that was the original conversation point.

> Or maybe it'll create even more confusion

But I agree with this, it may be misleading for the users. I stick to opinion that we should introduce as less changes as possible with this NeoXAMEV fork. Also, since our codebase is out-of-date wrt the Cancun Geth version, we may miss some updates from the original transient storage codebase, its implementation is a little bit more complex than MCOPY.

Thus, I suggest not to enable it.

txhsl (approved):

core/vm/jump_table.go
roman-khimov: OK, let's keep the minimum amount required for DKG contracts now.

roman-khimov (approved):

AnnaShaleva:
@songb2, testing is still required, and we're moving to a point where it will be possible to test it with DKG contracts. But for that we firstly need to update DKG a bit and implement https://github.com/bane-labs/go-ethereum/issues/323. I'll merge this PR, but you may still test it in the current version. If any issues are found, then please, submit issues and we may fix them in a separate PR.

PR #330, opened 2024-10-09
core: extend systemcontracts OnPersist with call to KeyManagement

AnnaShaleva:
Add a stub of KeyManagement contract. Extend system OnPersist call with a call to KeyManagement OnPersistV2 function. Enable it starting from NeoXAMEV fork. Use the same notation (V2) as for Governance OnPersistV2 to keep the naming fork-aligned.

Implement
https://github.com/bane-labs/go-ethereum/pull/312#discussion_r1791494184.

@txhsl, please review. I added a stub for KeyManagement contract, we can merge this PR as is and then you'll replace this stub by your code in #312.


privnet/seven/genesis_privnet.json
```diff
@@ -241,6 +241,10 @@
...
+    },
+    "0x1212100000000000000000000000000000000009": {
```
txhsl: I'm afraid it is not an easy copy-paste here. Let me update in #312 and plz rebase this on that.

core/systemcontracts/contracts.go
```diff
@@ -9,8 +9,14 @@ import (
...
+	// keyManagementABI is a partial ABI of system anti-MEV KeyManagement contract,
+	// it contains a minimum required list of methods needed for system interactions.
+	keyManagementABI = `[{"inputs":[],"name":"onPersistV2","outputs":[],"stateMutability":"nonpayable","type":"function"}]`
```
txhsl, line 19: I also expect a simpler ABI for Governance like this, only a `onPersistV2` and some other necessary methods, without `withdrawRegisterFee`, `votedTo` and so on.

If you agree, I can later open a new PR for this, or after dkg integration.

AnnaShaleva: > If you agree, I can later open a new PR for this, or after dkg integration.

I thought about it, but I have some tests that actually use some other governance methods. I'll fix it in the current PR, but a set of methods will be extended a bit compared to the set that is used by dBFT.

privnet/seven/genesis_privnet.json
AnnaShaleva: Well, it's not a copy-past, I actually compiled a KeyManagement stub and checked that it works. But it's even better now when the full KeyManagement contract is merged, I'll update PR soon.

AnnaShaleva: Oh, now I got your idea, I forgot to recompile the Stub contract with updated self address.

core/systemcontracts/contracts.go
AnnaShaleva: Fixed.

privnet/seven/genesis_privnet.json
AnnaShaleva: Fixed.

txhsl (approved):

PR #332, opened 2024-10-18
antimev: integrate key management contract

### LINKED ISSUE #160: Integrate dynamic keys into AMEV solution
# Implementation

Depends on #22, #159.

### LINKED ISSUE #222: Rework voter reward distribution scheme for `transferVote`
The problem is described in https://github.com/bane-labs/go-ethereum/pull/182#discussion_r1576317119.

We need to decide on the desired behaviour (both cases are technically valid).

### LINKED ISSUE #323: Use Governance contract provided and ordered address list in dBFT

### LINKED ISSUE #333: Integrate DKG process in dBFT with KeyManagement system contract
# Rationale

Ref #301, #312. We already have a full process of local DKG for antimev, and a DKG system contract. Now we need them to be integrated with dBFT, and become an automatic task that will be executed by CNs every Governance epoch.

# Implementation

Update dBFT to send and listen DKG messages regularly.

### LINKED ISSUE #335: Remove public attributions in local antimev keystore and its persistence, and always fetch them from the contract
# Rationale

The current implementation of antimev keystore persists public DKG data, especially the PVSS, to support a local verification of DKG messages. But now we are going to integrate with DKG contract, then these data can be removed from local memory and file persistence. It will be helpful in reducing the keystore file size.

Ref https://github.com/bane-labs/go-ethereum/pull/326#issue-2549455739.

# Implementation

1. Remove local cross verifications and persistence of DKG PVSS;
2. Make sure that local keystore data can be compatible with the DKG contract state in any case.

### LINKED ISSUE #340: Use epoch number instead of epoch start height as the index in key management mapping
# Rationale

The current implementation of DKG contract use epoch height as mapping index, which may cause an issue in data fetching if our epoch duration gets changed in the future.

# Implementation

Use epoch number instead will be better.

### LINKED ISSUE #345: Add a `NeoXDKGBlock` setting to enable DKG before Anti-MEV dBFT
# Rationale

The current `NeoXAMEVBlock` setting enables the Anti-MEV feature in dBFT. For the Privnet, we have preset keystore files to support this feature, but for Testnet and Mainnet, we need a successful round of DKG before this height, otherwise no secret is shared and usable.

# Implementation

1. Add a `NeoXDKGBlock` height before `NeoXAMEVBlock`;
2. Enable new precompiles and the `KeyManagement` contract from `NeoXDKGBlock`;
3. Still Enable Anti-MEV dBFT from `NeoXAMEVBlock`.

chenquanyu:
Close #160, close #222, close #323, close #333, close #335, close #340, close #345.

Add `handleDKG` in dbft to integrate `KeyManagement` contract. This method will interact with the key management contract to update the `amevKeystore` of the new consensus node at a specific block height.


core/systemcontracts/contracts.go
```diff
@@ -13,10 +13,11 @@ import (
...
...pe": "function"},{"inputs": [],"name": "onPersistV2","outputs": [],"stateMutability": "nonpayable","type": "function"},{"inputs": [{"internalType": "uint256","name": "shareRate","type": "uint256"}],"name": "registerCandidate","outputs": [],"stateMutability": "payable","type": "function"},{"inputs": [],"name": "revokeVote","outputs": [],"stateMutability": "nonpayable","type": "function"},{"inputs": [],"name": "sharePeriodDuration","outputs": [{"internalType": "uint256","name": "","type": "uint256"}],"stateMutability": "view","type": "function"},{"inputs": [{"internalType": "address","name": "candidateTo","type": "address"}],"name": "transferVote","outputs": [],"stateMutability": "nonpayable","type": "function"},{"inputs": [{"internalType": "address","name": "voter","type": "address"}],"name": "unclaimedRewardOf","outputs": [{"internalType": "uint256","name": "","type": "uint256"}],"stateMutability": "view","type": "function"},{"inputs": [{"internalType": "address","name": "to","type": "address"}],"name": "vote","outputs": [],"stateMutability": "payable","type": "function"},{"inputs": [],"name": "withdrawRegisterFee","outputs": [],"stateMutability": "nonpayable","type": "function"}]`
```
txhsl: We don't need all of them, please only add new ones that get used in this PR, as well as the `keyManagementABI`.

contracts/solidity/KeyManagement.sol
```diff
@@ -312,6 +312,19 @@ contract KeyManagement is GovProxyUpgradeable, IKeyManagement {
         return idxs;
     }
 
+    function isShareReady() external view returns (bool) {
+        // check period
+        (, uint targetHeight) = _checkPeriodAllowed(Period.Recover);
+
+        uint n = IGovernance(GOV).consensusSize();
+        for (uint i = 0; i < n; i++) {
+            if (shareMsgs[targetHeight][i + 1].length == 0) {
```
txhsl: I prefer to check like `if (shareMsgs[targetHeight][i + 1].length < n)`, and start count `i` from `1` if possible.

consensus/dbft/dbft.go
```diff
@@ -2020,3 +2041,533 @@ func (c *DBFT) getValidators(blockNum *uint64, state *state.StateDB, header *typ
...
+		}
+		c.targetHeight = currentEpochStartHeight + epochDuration
+		c.shareDuration = sharePeriodDuration
+		c.shareReady = false
+		c.consensusList = consensusList
+		c.needRecover = false
+	}
+
+	shareStartHeight := c.targetHeight - 2*c.shareDuration
+	recoverStartHeight := shareStartHeight + c.shareDuration
```
txhsl: Maybe we should move these two computations into Governance contract? If `shareDuration` is not necessary anywhere else.🤔

consensus/dbft/dbft.go
```diff
@@ -2020,3 +2041,533 @@ func (c *DBFT) getValidators(blockNum *uint64, state *state.StateDB, header *typ
...
+	consensusSize := uint64(len(c.consensusList))
+
+	// send share and reshare tx when currentHeight == shareStartHeight+1
+	if currentHeight == shareStartHeight+1 {
+		state, err := c.chain.StateAt(h.Root)
+		if err != nil {
+			return fmt.Errorf("failed to call StateAt: %w", err)
+		}
+		indexOfResharing, err := c.indexOfResharing(&c.signer, state, h)
```
txhsl: The following code needs to be updated as well, otherwise the index of DKG and the index of AMEV decryption will mismatch.

https://github.com/bane-labs/go-ethereum/blob/fc9b436aa2c8d9bea0b03258de6f916100fa66c4/consensus/dbft/dbft.go#L2026-L2031

Also the same logic in signature verification.

https://github.com/bane-labs/go-ethereum/blob/fc9b436aa2c8d9bea0b03258de6f916100fa66c4/consensus/dbft/dbft.go#L908-L914

consensus/dbft/dbft.go
```diff
@@ -2020,3 +2041,533 @@ func (c *DBFT) getValidators(blockNum *uint64, state *state.StateDB, header *typ
...
+			}
+			log.Info("reshare transaction sent", "txHash", txHash)
+		}
+
+		indexOfSharing, err := c.indexOfSharing(&c.signer, state, h)
+		if err != nil {
+			return fmt.Errorf("failed to call indexOfSharing: %w", err)
+		}
+		if indexOfSharing > 0 {
+			msgs, pvss, err := c.amevKeystore.OnReshareFinish()
```
txhsl: I wish the period order gets changed in `antimev/keystore.go`, make `OnValidatorList` and `OnReshareFinish` something as `OnReshareStart` and `OnShareStart`.

We can wait #326 to get fully tested and merged, or rebased to it to have a rework.

consensus/dbft/dbft.go
```diff
@@ -2020,3 +2041,533 @@ func (c *DBFT) getValidators(blockNum *uint64, state *state.StateDB, header *typ
...
+		c.consensusList = consensusList
+		c.needRecover = false
+	}
+
+	shareStartHeight := c.targetHeight - 2*c.shareDuration
+	recoverStartHeight := shareStartHeight + c.shareDuration
+	consensusSize := uint64(len(c.consensusList))
+
+	// send share and reshare tx when currentHeight == shareStartHeight+1
+	if currentHeight == shareStartHeight+1 {
```
txhsl: Why we use `shareStartHeight+1` instead of `shareStartHeight`?

consensus/dbft/dbft.go
```diff
@@ -233,10 +235,19 @@ type DBFT struct {
 
 	// various native contract APIs that dBFT uses.
 	ethAPI          *ethapi.BlockChainAPI
+	txAPI           *ethapi.TransactionAPI
 	validatorsCache *lru.Cache[uint64, []common.Address]
 
 	// The fields below are for testing only
 	fakeDiff bool // Skip difficulty verifications
+
+	// fields for dkg
```
txhsl: To have consistent format, I think we need comments to start with capital letters, unless there is a name of values or method.

consensus/dbft/dbft.go
```diff
@@ -2020,3 +2041,533 @@ func (c *DBFT) getValidators(blockNum *uint64, state *state.StateDB, header *typ
...
+		state, err := c.chain.StateAt(h.Root)
+		if err != nil {
+			return fmt.Errorf("failed to call StateAt: %w", err)
+		}
+		c.shareReady, err = c.isShareReady(state, h)
+		if err != nil {
+			return fmt.Errorf("failed to call isShareReady: %w", err)
+		}
+		if !c.shareReady {
+			return fmt.Errorf("share msgs not enough: %w", err)
```
txhsl: The `err` would be `nil` here.

consensus/dbft/dbft.go
```diff
@@ -2020,3 +2041,533 @@ func (c *DBFT) getValidators(blockNum *uint64, state *state.StateDB, header *typ
...
+			return nil
+		}
+
+		// only indexesNeedRecover <= (consensusSize - threshold) can recover
+		threshold := consensusSize - (consensusSize-1)/3
+		if len(indexesNeedRecover) > 0 && len(indexesNeedRecover) <= int(consensusSize-threshold) {
+			c.needRecover = true
+		}
+		if !c.needRecover {
+			return fmt.Errorf("reshare msgs not enough: %w", err)
```
txhsl: As well as here, a `nil` `err`.

consensus/dbft/dbft.go
```diff
@@ -2020,3 +2041,533 @@ func (c *DBFT) getValidators(blockNum *uint64, state *state.StateDB, header *typ
...
+		}
+	}
+
+	// if share not ready, no need to recover
+	if !c.shareReady {
+		return nil
+	}
+
+	// send reshareRecovered at height recoverStartHeight+1+c.shareDuration/3
+	if c.needRecover && currentHeight == recoverStartHeight+1+c.shareDuration/3 {
```
txhsl: Then we will have 3 stages in recovering, can we optimize this?

consensus/dbft/dbft.go
txhsl: I wish we may use the address in `amevKeystore`, even though they should be the same one. I export it in #326.

consensus/dbft/dbft.go
```diff
@@ -233,10 +235,19 @@ type DBFT struct {
...
 	// various native contract APIs that dBFT uses.
 	ethAPI          *ethapi.BlockChainAPI
+	txAPI           *ethapi.TransactionAPI
 	validatorsCache *lru.Cache[uint64, []common.Address]
 
 	// The fields below are for testing only
 	fakeDiff bool // Skip difficulty verifications
+
+	// fields for dkg
+	lastHeight    uint64
```
AnnaShaleva: We already have `lastIndex` - the index of the latest block processed by dBFT. Currently it is updated only by dBFT callbacks from a single thread. If you use this field in the same way (from a single thread from dBFT callbacks), then reuse `lastIndex` instead of `lastHeight`. But concurrency is important, it needs to be checked.

consensus/dbft/dbft.go
```diff
@@ -1769,6 +1785,11 @@ func (c *DBFT) handleChainBlock(h *types.Header, checkForSync bool) error {
 		return nil
 	}
 
+	// handle DKG
+	if !c.syncing.Load() {
```
AnnaShaleva: Hi, deadlocks! You don't need this condition here, because the condition above already guarantees proper behaviour:
```
if checkForSync && c.syncing.Load() {
```

consensus/dbft/dbft.go
```diff
@@ -1769,6 +1785,11 @@ func (c *DBFT) handleChainBlock(h *types.Header, checkForSync bool) error {
...
+		c.handleDKG(h)
```
AnnaShaleva: Move it under `if uint32(h.Number.Uint64()) >= c.dbft.BlockIndex {` condition below, place it before the block processing itself.

consensus/dbft/dbft.go
```diff
@@ -2020,3 +2041,533 @@ func (c *DBFT) getValidators(blockNum *uint64, state *state.StateDB, header *typ
 func (c *DBFT) shouldUpdateCommitteeAt(blockNum uint64) bool {
 	return blockNum%uint64(len(c.config.StandByValidators)) == 0
 }
+
+// handleDKG handle the transaction send for dkg process.
+// It calls and send transaction to KeyManagement contract at right height with amev store.
```
AnnaShaleva:
```suggestion
// handleDKG handles the transaction submission for DKG process.
// It constructs and sends transaction to KeyManagement contract using amev store.
```

consensus/dbft/dbft.go
```diff
@@ -2020,3 +2041,533 @@ func (c *DBFT) getValidators(blockNum *uint64, state *state.StateDB, header *typ
...
+func (c *DBFT) handleDKG(h *types.Header) error {
+	// just return when antiMEV not enabled
+	currentHeight := h.Number.Uint64()
+	if int64(currentHeight) < c.config.antiMEVEnablingHeight {
+		return nil
+	}
```
AnnaShaleva: Remove it, this check will be presented at the caller's level.

consensus/dbft/dbft.go
```diff
@@ -2020,3 +2041,533 @@ func (c *DBFT) getValidators(blockNum *uint64, state *state.StateDB, header *typ
...
+
+	// avoid duplicate heights
+	if currentHeight <= c.lastHeight {
+		return nil
+	}
```
AnnaShaleva: Remove it, caller's code will provide it.

consensus/dbft/dbft.go
AnnaShaleva: Move it under `c.chain.Config().IsNeoXAMEV()`

AnnaShaleva: OK, and you actually have to move it inside `postBock` after all corresponding indexes update. Because it's guaranteed that `postBlock` is called *once* per height and *every* time when it's needed.

consensus/dbft/dbft.go
AnnaShaleva: It's a matter of https://github.com/bane-labs/go-ethereum/issues/323. And no, we can't just update this code, because it's a part of existing network. We need to keep NextConsensus-related code as is, and what should be changed is https://github.com/bane-labs/go-ethereum/blob/10c914d56e43e2b2f715a39f24e0dfe2064f9fe4/consensus/dbft/dbft.go#L769
In this line we just need to convert message validator index to DKG index by using current consensus list from Governance contract.

AnnaShaleva: And BTW, a question to @txhsl regarding sorting order of CN addresses in the Governance contract: currently Governance contract sort them by votes. But when votes are equal it seems to me that we still have to sort them by alphabet order, exactly like in N3.

consensus/dbft/dbft.go
AnnaShaleva: Let's merge #326 firstly.

consensus/dbft/dbft.go
```diff
@@ -2020,3 +2041,533 @@ func (c *DBFT) getValidators(blockNum *uint64, state *state.StateDB, header *typ
...
+				pubs = append(pubs, pubKey)
+			}
+			msgs, pvss, err := c.amevKeystore.OnValidatorList(c.consensusList, pubs)
+			if err != nil {
+				return fmt.Errorf("failed to call OnValidatorList(), err: %w", err)
+			}
+			// send reshare tx
+			txHash, err := c.reshare(pvss, msgs)
+			if err != nil {
+				return fmt.Errorf("failed to send reshare transaction, err: %w", err)
```
AnnaShaleva: We need a retry mechanism for all these transaction sending processes, because anything may go wrong. I suggest to make a flag that denotes whether transaction was successfully sent for every period, and if not, then we have to retry during the next block processing.

consensus/dbft/dbft.go
```diff
@@ -2020,3 +2041,533 @@ func (c *DBFT) getValidators(blockNum *uint64, state *state.StateDB, header *typ
 func (c *DBFT) shouldUpdateCommitteeAt(blockNum uint64) bool {
 	return blockNum%uint64(len(c.config.StandByValidators)) == 0
 }
+
+// handleDKG handle the transaction send for dkg process.
+// It calls and send transaction to KeyManagement contract at right height with amev store.
+func (c *DBFT) handleDKG(h *types.Header) error {
```
AnnaShaleva: It's 2K+ lines of code in this file already. Please, let's move `handleDKG` function and all contract reader helpers to a separate file, something like `dkg.go`.

core/systemcontracts/contracts.go
AnnaShaleva: Also please, don't remove the existing methods.

consensus/dbft/dbft.go
txhsl: > Currently Governance contract sort them by votes. But when votes are equal it seems to me that we still have to sort them by alphabet order, exactly like in N3.

Well, we used a Top-K algorithm in Governance contract, so even in the elected CN list, they are not strictly sorted by votes. Instead, our Governance website does this. So my question is, is the order of CNs that important, in dBFT or anything else?

core/systemcontracts/contracts.go
chenquanyu: I used `IGovernance` and `IKeyManagement` file in interfaces folder, I think we should keep this abi string consistent with the interface version.

txhsl: We have already removed a big part of it, plz refer https://github.com/bane-labs/go-ethereum/pull/330/files#diff-222215cf4ecad61a1c003cb08c4159e8b0350452f7266bbbc2e8b6df88954759R13.

consensus/dbft/dbft.go
chenquanyu: Considering reorg, it may be safer to lag behind by one block.

txhsl:
I merged #326, and will open a PR to this, to modify some behaviors of antimev keystore for DKG integration. @chenquanyu Please, a rebase will be helpful.

consensus/dbft/dbft.go
chenquanyu: OK, we should also skip `handleDKG` when `checkForSync` is false.

consensus/dbft/dbft.go
chenquanyu: I moved it to `postBlock ` and removed last height check.

core/systemcontracts/contracts.go
chenquanyu: Done.

consensus/dbft/dbft.go
chenquanyu: Fixed.

consensus/dbft/dbft.go
chenquanyu: Fixed by moving `handleDKG` to `postBlock`.

AnnaShaleva:
Can we rebase onto fresh master and fix remaining conversations? I'll review one more time after this.

txhsl:
I'm reworking some parts of the keystore, to include #335 in this PR.

consensus/dbft/dkg.go
```diff
@@ -0,0 +1,634 @@
...
+		sharePeriodDuration, err := c.sharePeriodDuration(state, h)
+		if err != nil {
+			return fmt.Errorf("failed to call currentEpochStartHeight: %w", err)
+		}
+		consensusList, err := c.getCurrentConsensus(state, h)
+		if err != nil {
+			return fmt.Errorf("failed to call getCurrentConsensus: %w", err)
+		}
+		c.targetHeight = currentEpochStartHeight + epochDuration
+		c.preEpochStartHeight = currentEpochStartHeight - epochDuration
```
txhsl: If the current ongoing DKG is targeting `currentEpochStartHeight + epochDuration`, then the latest shared DKG for resharing and recovery should be fetched by `currentEpochStartHeight`.

consensus/dbft/dkg.go
```diff
@@ -0,0 +1,634 @@
...
+
+		var shareErr, reshareErr error
+		if isCurrentConsensus {
+			rMsgs, rPvss, err := c.amevKeystore.DKGReshare()
+			if err != nil {
+				return fmt.Errorf("failed to call amevKeystore.DKGReshare, err: %w", err)
+			}
+			// send reshare tx
+			txHash, err := c.reshare(rPvss, rMsgs)
+			if err != nil {
```
txhsl: I don't think there will be such an error that can be solved by this retry. The `reshare`, `share` and `recover` operation is state-independent between each sender but only depend on history data, so if the op failed, possibly because i) `rMsgs` or `rPvss` is invalid; ii) `reshare` finished or hasn't started. Well, since `currentHeight == shareStartHeight+1`, then any retry will fail, because nothing is changed to make it successful.

What I'm thinking about retry is e.g. transaction cannot get confirmed for a long time, so we have to improve its priority in mempool. But how do you think? @AnnaShaleva

consensus/dbft/dkg.go
```diff
@@ -0,0 +1,634 @@
...
+			return fmt.Errorf("failed to call indexCurrentNeedRecovering, err: %w", err)
+		}
+		// reshare finished, no need to recover
+		if len(indexesNeedRecover) == 0 {
+			return nil
+		}
+
+		// only indexesNeedRecover <= (consensusSize - threshold) can recover
+		threshold := consensusSize - (consensusSize-1)/3
+		if len(indexesNeedRecover) > 0 && len(indexesNeedRecover) <= int(consensusSize-threshold) {
```
txhsl: I see a redundancy here. Let me fix it.

consensus/dbft/dkg.go
AnnaShaleva: I was thinking more about some network-related problems, e.g. if CN is disconnected from its peers for a long time or transaction is not relayed for some reason. We have mempool reannouncement mechanism, but it may be not enough to reliably relay transaction and include it into the block. So my suggestion was retry with the following rules: during the first block of `Reshare` period CN sends its reshare transaction. Then at block `Reshare + X` CN checks if this transaction was persisted to the chain, and if not, then CN should send one more reshare transaction (similar to the previous one). CN should take attempts to send reshare transaction every X blocks until any of transactions is persisted or until `Reshare` period is over.

> so we have to improve its priority in mempool. 

I don't think it's needed because `c.reshare` in fact adds transaction to the node's pool as  "local" tx, and local transactions are natively prioritized over non-local transactions by the node's mempool itself. What may be implemented is kind of "high priority" transactions (similar to what we have in N3, the `HightPriority` transaction attribute), but it's a protocol-level change, and it needs to be discussed. So it's definitely not a matter of this PR and not for the current release.

AnnaShaleva: > So it's definitely not a matter of this PR and not for the current release.

Ref. https://github.com/bane-labs/go-ethereum/issues/344.

chenquanyu: I have added "txWatchList" for checking tx status and retrying sending.

consensus/dbft/dbft.go
chenquanyu: Add `getDKGIndex` for this.

consensus/dbft/dbft.go
chenquanyu: Now we have 2 stages in recovering.

txhsl: Yep, although the contract only has 1 `recovering` period. I still don't think this is the best handling, but the code itself is reasonable.

params/config.go
```diff
@@ -349,7 +349,8 @@ type ChainConfig struct {
 	// Fork scheduling was switched from blocks to timestamps here
 
 	ShanghaiTime  *uint64  `json:"shanghaiTime,omitempty"`  // Shanghai switch time (nil = no fork, 0 = already on shanghai)
-	NeoXAMEVBlock *big.Int `json:"neoXAMEVBlock,omitempty"` // Block-based switch to anti-MEV related logic for dBFT, system contracts and processing engine (nil = no fork, 0 = already activated)
+	NeoXDKGBlock  *big.Int `json:"NeoXDKGBlock,omitempty"`  // Block-based switch to DKG related logic for dBFT, system contracts and processing engine (nil = no fork, 0 = already activated)
+	NeoXAMEVBlock *big.Int `json:"neoXAMEVBlock,omitempty"` // Block-based switch to anti-MEV related logic for dBFT (nil = no fork, 0 = already activated)
```
AnnaShaleva, line 353: Regarding this change: we're strongly against of splitting DKG and AMEV forks into two *on config level*. We need to introduce as less changes as possible to the node config. AMEV fork won't work without DKG fork pre-enabled anyway.

Our suggestion is:
1. Keep only NeoXAMEVBlock in the node's config.
2. Bind initial DKG setup to `NeoXAMEVBlock` height.
3. Bind dBFT switch and block signatures scheme switch to `NeoXAMEVBlock + H` where `H` is some constant number of blocks.

As a result, we'll have a single fork in the node's config and all AMEV-related functionality bound to this fork. But the functionality itself is split into multiple stages.

AnnaShaleva (commented):
Give me some time, I'll review the rest of the code in a couple of days once I publish a draft of blocks signature scheme switch.

params/config.go
txhsl: Before implementing this, may I ask for advice? Should we define the constant variable `H` in `consensus/dbft` or `params/config`?

AnnaShaleva: To me it looks more like config-level helper, because then it may be implemented the same way as `(c *ChainConfig) IsNeoXAMEV`.

consensus/dbft/dbft.go
txhsl: Still, reorg can happen at any height, and the tx list is not affected, so this might make no difference.

params/config.go
txhsl: Sorry, I'm now back to work on this.

> 3. Bind dBFT switch and block signatures scheme switch to NeoXAMEVBlock + H where H is some constant number of blocks.

So, this may cause an issue since the value is determined based on compilation. The value `H` can be consistent on both Mainnet and Testnet, but a large `H` doesn't gonna work on Privnet for a fast check.

consensus/dbft/dbft.go
txhsl: Removed.

txhsl:
Sorry for a lot of commits, this PR should be squashed to get merged, but I think it is now ready for testing. @songb2

consensus/dbft/dbft.go
```diff
@@ -1937,6 +1959,13 @@ func (c *DBFT) handleChainBlock(h *types.Header, checkForSync bool) error {
 			"primary", h.Primary(),
 			"coinbase", h.Coinbase,
 			"mix digest", h.MixDigest.String())
+		// handle DKG
+		if height >= uint64(c.config.dkgEnablingHeight) && height > c.lastIndex {
```
txhsl: @AnnaShaleva Ref https://github.com/bane-labs/go-ethereum/pull/332#discussion_r1806313490.

In our testing, we found that sometimes the state was missing in the database, when we called `c.chain.StateAt(h.Root)` in `c.postBlock(h)`, and the code has to give up the whole logic. So it is not safe enough to ensure keystore epoch change.

I tried to solve this issue, and moved `c.handleDKG(h)` here, but I still need your opinion. Is it a safe place to do that, so that it can get checked every height without any error? Or is there any other better place to this?

At least, I think we need to ensure that the new state is used after `c.blockQueue.PutBlock()`.

AnnaShaleva: Let me check this issue more precisely and review the whole PR. In general, by the moment of a call to `postBlock` I'd expect the state to be stored in the DB, because it's called either after the node receives (and processes) new block via P2P or when the block submitted to the block queue by dBFT (and processing is finished). So both cases assume that the state is present in the DB.

txhsl: Maybe @songb2 can show you the error log, it happens more frequently on machines which are lack of performance.

Also, this PR may need #355 before rebased to main branch, because there is a conflict in `tpke.PublicKey.Bytes()`. It is related to the keystore persistent format, although not that important for current test, and can be modified later.

songb2: The error sometimes appears in the log:
```
…
ERROR[11-19|19:18:17.465] handleDKG error                          height=130 err="failed to call StateAt: missing trie node 42ae22cbc87a74b71f9fb22af5affaf783898748daa286722e1b212d4937a97a (path ) state 0x42ae22cbc87a74b71f9fb22af5affaf783898748daa286722e1b212d4937a97a is not available, not found"
…
2024-11-19T19:20:00.893+0800	INFO	dbft@v0.3.1-0.20241030192742-82ad77871e88/check.go:120	approving block	{"height": 140, "hash": "0xac26724b63b3895a28c2bd0509e2d41740a02591b6c72ce4bba45c7adeef67ff", "tx_count": 0, "merkle": "0x6619d217f3fdb841c2b68ba9edf972460006b20a1f1dea0991727d9bccb50477", "prev": "0xad429f7d77dbf85880bcb686a04eacaacb06f6d4e86f6e8d05dc01bc8f7e32c1"}
ERROR[11-19|19:20:00.924] handleDKG error                          height=140 err="DKG failed, share msgs not enough"
```

antimev/keygroup.go
```diff
@@ -91,67 +86,46 @@ func (tkg *thresholdKeyGroup) newTemplateForReshare(validators []common.Address,
...
-func (tkg *thresholdKeyGroup) dkgAggregate(scaler int) error {
-	// Compute public key S=sum(A0)
-	scs := make([]*tpke.Commitment, len(tkg.holders))
-	for i, pvss := range tkg.pvsses {
-		if pvss == nil {
-			return ErrSecretShareNotEnough
+// aggregate aggregates received secrets and commitments to get global
+// public key and a local piece of secret key for message decryption and signing.
+func (tkg *thresholdKeyGroup) aggregate(scaler int, aggregatedCmt []byte, selfIndex int) error {
+	if selfIndex > 0 {
```
AnnaShaleva: Why have you added this `selfIndex` check? Is it possible that `selfIndex` is negative? If we're talking about watch-only node, then this path should probably be handled at the dBFT level (although let's discuss it after your clarification).

antimev/keystore.go
```diff
@@ -102,6 +118,27 @@ func (ks *KeyStore) MessagePubKey() string {
...
+
+// LastGlobalPubKey returns last round global public key.
+func (ks *KeyStore) LastGlobalPubKey() (*tpke.PublicKey, error) {
+	ks.mu.RLock()
+	defer ks.mu.RUnlock()
+	if ks.reshared == nil {
+		return nil, ErrNoPubKey
+	}
+	return ks.reshared.globalPubKey, nil
+}
```
AnnaShaleva: It's good that you've added these APIs because they are needed for DKG-based block signature scheme during NextConsensus calculation in the last block of every epoch. Could you please name the cases when LastGlobalPubKey is nil? I mean, if CNs has successfully completed sharing/resharing process and OnEpochChange was successfully executed and CurrentGlobalPubKey is not nil, then may I rely on fact that LastGlobalPubKey is also not nil?

antimev/keystore.go
```diff
@@ -131,248 +175,282 @@ func (ks *KeyStore) HasShared() bool {
...
 	ks.mu.Lock()
 	defer ks.mu.Unlock()
+	// Check if out of the period
+	if ks.sharing == nil {
+		return ErrKeyGroupNotExists
+	}
+	// If is a member of receiver
+	selfIndex := ks.sharing.holderIndex(ks.address)
+	if selfIndex < 1 {
+		return ErrNotParticipant
```
AnnaShaleva: I think we need a check for that at the dBFT level as far. A call to `ReceiveSecretShare` should be explicitly skipped for WatchOnly nodes. But then WatchOnly nodes won't be able to properly handle some of the consensus messages, for example, they won't be able to verify PreCommits, decrypt encrypted transactions or to construct threshold block signatures. Is there a way for non-DKG-member to decrypt message or construct signature based on received shares of other valid GKG-members?

AnnaShaleva: Let's discuss this topic before any implementation changes.

antimev/keystore.go
```diff
@@ -131,248 +175,282 @@ func (ks *KeyStore) HasShared() bool {
...
+	err = ks.sharing.receiveShareMessage(fromIndex, ss, p, selfIndex)
+	if err != nil {
+		return err
+	}
+	return ks.saveStoreAndReInitialize()
 }
 
 // ReceiveSecretReshare tries to verify a resharing message array and store their data.
-func (ks *KeyStore) ReceiveSecretReshare(from common.Address, ers [][]byte, pvss []byte) error {
+func (ks *KeyStore) ReceiveSecretReshare(fromIndex int, ers [][]byte, pvss []byte) error {
```
AnnaShaleva: Could you please add a note to method's documentation regarding `fromIndex`: is it an index of CN in the list sorted by address alphabet order or sorted the same way as in the Governance contract?

antimev/signature.go
```diff
@@ -14,6 +14,8 @@ var (
 
 // SignShare tries to sign a message with local private key.
 func (ks *KeyStore) SignShare(msg []byte) (*tpke.SignatureShare, error) {
+	ks.mu.RLock()
+	defer ks.mu.RUnlock()
```
AnnaShaleva: Why have you added mutex to the KeyStore? dBFT is a state machine, every callback is guaranteed to be called sequentially, there should be no concurrent calls. And there are no users of KeyStore other than dBFT and unit tests, but unit tests should also be fine with mutex-less code because a single KeyStore instance is not shared between multiple unit-tests.

cmd/geth/antimevcmd.go
```diff
@@ -105,6 +106,30 @@ For non-interactive use, password can be specified with --antimev.password flag:
...
+					utils.AntiMEVKeyStoreFlag,
+					utils.AntiMEVPasswordFlag,
+				},
+				Description: `
+    geth antimev reset
+
+Reset the existing antimev keystore.
+
+The keystore is reset to the initial state, all DKG data are deleted to recover
+from a stuck epoch change.
```
AnnaShaleva, line 126: It sounds scary. Have you faced with the situation when epoch transition is stuck? I thought that we have reshare phase specifically for cases when shearing is not successful.

cmd/geth/antimevcmd.go
```diff
@@ -205,3 +241,13 @@ func antimevUpdate(ctx *cli.Context) error {
 	}
 	return nil
 }
+
+// antimevReset cleans all DKG data but keep the message keypair and other parameters.
+func antimevReset(ctx *cli.Context) error {
+	ks := makeAntimevKeystore(ctx)
```
AnnaShaleva, line 249: Please, ask the user for confirmation before reset. It's crucial.

consensus/dbft/dbft.go
```diff
@@ -2119,6 +2147,26 @@ func (c *DBFT) APIs(chain consensus.ChainHeaderReader) []rpc.API {
...
+
+func (c *DBFT) shouldUpdateCommitteeAt(blockNum uint64) bool {
+	return blockNum%uint64(len(c.config.StandByValidators)) == 0
+}
+
+// getOriginalValidators returns validators chosen in the result of the latest
+// finalized voting epoch. It calls Governance contract under the hood. The call
+// is based on the provided state or (if not provided) on the state of the block
+// with the specified height. Validators returned from this method are not
+// sorted with original order from Governance contract. This method uses cached values in case of validators
```
AnnaShaleva: > Validators returned from this method are not sorted with original order from Governance contract.

Validators returned from this method are sorted in the original order used by Governance contract.

consensus/dbft/dbft.go
```diff
@@ -2119,6 +2147,26 @@ func (c *DBFT) APIs(chain consensus.ChainHeaderReader) []rpc.API {
 // sorted in another way). This method uses cached values in case of validators
 // requested by block height.
 func (c *DBFT) getValidators(blockNum *uint64, state *state.StateDB, header *types.Header) ([]common.Address, error) {
+	res, err := c.getOriginalValidators(blockNum, state, header)
```
AnnaShaleva: Rename `getValidators` to `getValidatorsSorted` and then rename `getOriginalValidators` to `getValidators`.

consensus/dbft/dbft.go
```diff
@@ -2175,6 +2222,24 @@ func (c *DBFT) getValidators(blockNum *uint64, state *state.StateDB, header *typ
...
+func (c *DBFT) getDKGIndex(validatorIndex int, blockNum uint64) (int, error) {
+	originValidators, err := c.getOriginalValidators(&blockNum, nil, nil)
+	if err != nil {
+		return -1, err
+	}
+	if validatorIndex < 0 || validatorIndex >= len(originValidators) {
+		return -1, fmt.Errorf("invalid validator index: validators count is %d, requested %d", len(originValidators), validatorIndex)
+	}
+	orderedValidators := make([]common.Address, len(originValidators))
+	copy(orderedValidators, originValidators)
```
AnnaShaleva: We're at Go 1.22, use `slices.Clone()`.

consensus/dbft/dbft.go
```diff
@@ -2119,6 +2147,26 @@ func (c *DBFT) APIs(chain consensus.ChainHeaderReader) []rpc.API {
 // sorted in another way). This method uses cached values in case of validators
 // requested by block height.
 func (c *DBFT) getValidators(blockNum *uint64, state *state.StateDB, header *types.Header) ([]common.Address, error) {
+	res, err := c.getOriginalValidators(blockNum, state, header)
+	if err != nil {
+		return nil, err
+	}
+
+	slices.SortFunc(res, common.Address.Cmp)
```
AnnaShaleva: `res` is cached, so you can't just sort the same slice. Make a clone of slice before sorting.

consensus/dbft/dbft.go
```diff
@@ -2175,6 +2222,24 @@ func (c *DBFT) getValidators(blockNum *uint64, state *state.StateDB, header *typ
...
+	copy(orderedValidators, originValidators)
+	slices.SortFunc(orderedValidators, common.Address.Cmp)
+	addr := orderedValidators[validatorIndex]
+	for i := range originValidators {
+		if orderedValidators[i] == addr {
+			return i + 1, nil
+		}
+	}
+	// impossible case
+	return -1, nil
```
AnnaShaleva: Panic here instead of silent return. It's a program bug if execution flow reaches this place.

consensus/dbft/dbft.go
```diff
@@ -2175,6 +2222,24 @@ func (c *DBFT) getValidators(blockNum *uint64, state *state.StateDB, header *typ
...
+	originValidators, err := c.getOriginalValidators(&blockNum, nil, nil)
+	if err != nil {
+		return -1, err
+	}
+	if validatorIndex < 0 || validatorIndex >= len(originValidators) {
+		return -1, fmt.Errorf("invalid validator index: validators count is %d, requested %d", len(originValidators), validatorIndex)
+	}
+	orderedValidators := make([]common.Address, len(originValidators))
+	copy(orderedValidators, originValidators)
+	slices.SortFunc(orderedValidators, common.Address.Cmp)
```
AnnaShaleva, line 2411: For now the implementation is sufficient, but let's have https://github.com/bane-labs/go-ethereum/issues/357 included to the next-next release.

consensus/dbft/dkg.go
```diff
@@ -0,0 +1,843 @@
...
+	CurrentCNs         []common.Address `json:"currentCNs"`
+	PendingCNs         []common.Address `json:"pendingCNs"`
+	IndexNeedRecover   []uint64         `json:"indexNeedRecover"`
+	ShareOped          bool             `json:"shareOped"`
+	RecoverOped        bool             `json:"recoverOped"`
+	ReshareRecoverOped bool             `json:"reshareRecoverOped"`
+}
+
+// newSnapshot creates a new snapshot with the specified startup parameters.
+func (c *DBFT) newSnapshot(h *types.Header, state *state.StateDB, height uint64) (*Snapshot, error) {
```
AnnaShaleva, line 54: Initially, when we designed dBFT engine, I was explicitly against of any state stored for dBFT in the node's database, because otherwise there's a need for additional data management and these data must be strictly aligned with the current chain's state. I see that you introduce DKG state here, but is there any strong reason why snapshots should be stored? From my understanding, all necessary information is stored inside KeyManagement contract, so could you please explain what is missing in the KeyManagement and what is the reason for DKG snapshots introducing?

consensus/dbft/dkg.go
```diff
@@ -0,0 +1,843 @@
...
+			if err := c.prepareDKG(c.snapshot.CurrentCNs, state, h); err != nil {
+				return fmt.Errorf("failed to sync shared DKG, err: %v", err)
+			}
+			// If is a member of current consensus, then try sync secrets
+			if slices.Contains(c.snapshot.CurrentCNs, amevAddress) {
+				// Only a warning here, since it doesn't destroy dBFT and anti-mev,
+				// a new DKG can perform and the next reshare is recoverable.
+				// But it is dangerous if more than 1/3 CNs reinit their keystores
+				// or change their message key.
+				if err := c.syncLastRoundSecrets(c.snapshot, state, h); err != nil {
```
AnnaShaleva: Why error may happen here?

consensus/dbft/dkg.go
```diff
@@ -0,0 +1,843 @@
...
+				if err = c.taskReshare(currentHeight, recoverStartHeight); err != nil {
+					c.snapshot.ShareOped = true
+					return fmt.Errorf("failed to task DKG reshare, err: %v", err)
+				}
+			}
+		}
+		c.snapshot.ShareOped = true
+	}
+
+	if currentHeight >= recoverStartHeight && !c.snapshot.RecoverOped {
```
AnnaShaleva: Are you sure that `RecoverOped` doesn't contain a typo? I understand the responsibility of this field, but not sure that I understand its naming.

consensus/dbft/dkg.go
```diff
@@ -0,0 +1,843 @@
...
+			if currentHeight < item.EndHeight && !item.ConfirmedSuccess {
+				needRetry := false
+				// Send failed, just resend and set txHash
+				if item.TxHash == nil {
+					needRetry = true
+				}
+
+				// Send successfully, wait 3 blocks to check tx status
+				if item.TxHash != nil && currentHeight-item.SendHeight == 3 {
+					receipt, err := c.txAPI.GetTransactionReceipt(context.Background(), *item.TxHash)
```
AnnaShaleva: Is there something similar to N3's `ValidUntilBlock` in Ethereum protocol?

consensus/dbft/dkg.go
```diff
@@ -0,0 +1,843 @@
...
+					if err != nil {
+						needRetry = true
+						log.Error("DKG get transaction receipt failed", "err", err, "txHash", item.TxHash)
+					}
+					if receipt["status"] != uint(1) {
```
AnnaShaleva: There should be a special designated constant declared somewhere in the core for that. I mean, already existing constant.

consensus/dbft/dkg.go
```diff
@@ -0,0 +1,843 @@
...
+						log.Error("DKG get transaction receipt status error", "txHash", item.TxHash, "status", receipt["status"])
+					}
+				}
+
+				var err error
+				if needRetry {
+					item.TxHash, err = c.sendTxToKeyManagement(item.Method, item.Params...)
+					if err != nil {
+						retryList = append(retryList, item)
+						log.Error("retry sending transaction failed", "currentHeight", currentHeight, "method", item.Method, "err", err)
```
AnnaShaleva: What happens on the contract side if transaction with the same data is accepted twice? For example, if first transaction is not accepted after 3 blocks, then we try to send the second one, and meanwhile the first transaction is accepted along with the second one.

There should be nothing critical, but we'd better ensure that it's properly handled on the contract side. Also, is it possible to distinguish if second transaction is failed due to the fact that KeyManagement already contains data submitted by the first transaction from any other failure reason?

consensus/dbft/dkg.go
```diff
@@ -0,0 +1,843 @@
...
+	return nil
+}
+
+// syncRecoveredSecrets downloads DKG recoverings and related PVSS, and sends to keystore
+func (c *DBFT) syncRecoveredSecrets(snap *Snapshot, selfIndex uint64, state *state.StateDB, header *types.Header) error {
+	pvss, err := c.spvsses(snap.Round-1, selfIndex, state, header)
+	if err != nil {
+		return err
+	}
+	for i := uint64(1); i <= uint64(len(snap.CurrentCNs)); i++ {
```
AnnaShaleva: I'd say that for the sake of simplicity it's better be `for i := range snap.CurrentCNs`, then we may use `i+1` for a call to `recoverMsgs` and use `i` for `ReceiveRecoverShare`.

core/state_processor.go
```diff
@@ -226,26 +235,41 @@ func ProcessOnPersist(vmenv *vm.EVM, statedb *state.StateDB) error {
...
+		GasFeeCap: common.Big0,
+		GasTipCap: common.Big0,
+		To:        &systemcontracts.KeyManagementProxyHash,
+		Data:      data,
+	}
+	vmenv.Reset(NewEVMTxContext(msg), statedb)
+	statedb.AddAddressToAccessList(systemcontracts.KeyManagementProxyHash)
+	_, _, err = vmenv.Call(vm.AccountRef(msg.From), *msg.To, msg.Data, 30_000_000, common.U2560)
+	if err != nil {
+		return fmt.Errorf("onPersist call failed: %w", err)
```
AnnaShaleva, line 255: A lot of duplicating code, should be unified.

params/config.go
AnnaShaleva: @roman-khimov, need your resolution here.

privnet/four/genesis_privnet.json
```diff
@@ -148,6 +149,30 @@
...
+        "0x0dd6059bd488b7a63261e65b70ed9a4d3faf1661c64b272f9b5f7af023fefb3f": "0x6161333231343763383433366364663932383661346134343330316639363632",
+        "0x0dd6059bd488b7a63261e65b70ed9a4d3faf1661c64b272f9b5f7af023fefb40": "0x6162000000000000000000000000000000000000000000000000000000000000",
+        "0x5db24d05b56c3a5a71258d938a65f2cc28411a79253c52669d85833a02a0e964": "0x0105",
+        "0x7526037888f049a7306c4c81937364f790d4fd49b810bdbd431ffd92d1cff4ce": "0x3034396536633330333930333766333236636561346135663831633866343636",
+        "0x7526037888f049a7306c4c81937364f790d4fd49b810bdbd431ffd92d1cff4cf": "0x3239623462623533363837636334383739626335373961616161356363616538",
+        "0x7526037888f049a7306c4c81937364f790d4fd49b810bdbd431ffd92d1cff4d0": "0x6330303831303635353131393965373561383734313032343437346265656137",
+        "0x7526037888f049a7306c4c81937364f790d4fd49b810bdbd431ffd92d1cff4d1": "0x3039386566663664393063623333393364326537666530393736346664616363",
+        "0x7526037888f049a7306c4c81937364f790d4fd49b810bdbd431ffd92d1cff4d2": "0x3865000000000000000000000000000000000000000000000000000000000000",
```
AnnaShaleva, line 175: Could you please clarify, these are ready-to-use shares generated using privnet anti-MEV keustore, right? So that privnet works properly even without DKG transactions submission?

consensus/dbft/dbft.go
```diff
@@ -1938,7 +1960,13 @@ func (c *DBFT) handleChainBlock(h *types.Header, checkForSync bool) error {
 			"coinbase", h.Coinbase,
 			"mix digest", h.MixDigest.String())
 		c.postBlock(h)
-
+		// Handle DKG synchronization on anti-mev keystore and task transaction sending
+		if c.lastIndex >= uint64(c.config.dkgEnablingHeight) {
+			err := c.handleDKG(h)
+			if err != nil {
+				log.Error("handleDKG error", "height", height, "err", err)
+			}
```
AnnaShaleva: Also, we need to test if this code works properly in the following situation: the CN1 is being stopped for a while, other CNs move to the next DKG round, then CN1 gets back and starts syncing. During node sync all dBFT callbacks are intentionally inactive because we need the node state to be up-to-date. Once the CN1 is in sync we need to check that local anti-MEV keystore setup for the new DKG round works properly.

consensus/dbft/dbft.go
```diff
@@ -1938,7 +1960,13 @@ func (c *DBFT) handleChainBlock(h *types.Header, checkForSync bool) error {
 			"coinbase", h.Coinbase,
 			"mix digest", h.MixDigest.String())
 		c.postBlock(h)
-
+		// Handle DKG synchronization on anti-mev keystore and task transaction sending
+		if c.lastIndex >= uint64(c.config.dkgEnablingHeight) {
+			err := c.handleDKG(h)
```
AnnaShaleva: I'd say it's not exactly the proper place for this call, because `waitForNewSealingProposal` call below may go far beyond the height `h`. So I'd propose to get back to the previous solution and move `c.handleDKG` to `(c *DBFT) postBlock`. To deal with missing block state problem, let's reuse existing state DB as much as possible:
1. Change signature of `postBlock` to: `func (c *DBFT) postBlock(h *types.Header, state *state.StateDB) {`.
2. Provide the copy of existing state to `postBlock` from `WithProcessBlock` dBFT callback - it's guaranteed not to be nil.
3. Pass the state from `postBlock` to `handleDKG`; in `handleDKG` if the state is not given then retrieve it from the DB.
4. `WithProcessBlock` is not the only source of new blocks for dBFT engine, but I suspect that the problem is with blocks inserted by `WithProcessBlock`. If not, then sometimes block may arrive earlier via P2P interaction. We need to run the node with changes 1, 2 and 3 and check if the state sometimes is still missing. If so, then we need to check the synchronisation module code and see how new block notifications are emitted (before/after state persist)

consensus/dbft/dbft.go
AnnaShaleva: Regarding missing state problem, ref. https://github.com/bane-labs/go-ethereum/pull/332#discussion_r1869364276

Regarding https://github.com/bane-labs/go-ethereum/issues/355: I'll create a PR with these changes, it's rather simple.

AnnaShaleva: > Regarding https://github.com/bane-labs/go-ethereum/issues/355: I'll create a PR with these changes, it's rather simple. 

Ref. https://github.com/bane-labs/go-ethereum/pull/358.

txhsl: I've not seen related codes in #358, but I can revert to move it back to `postBlock()`.

cmd/geth/antimevcmd.go
txhsl: Well, OK.

privnet/four/genesis_privnet.json
txhsl: Nope, they are the message public keys for standby validators, so that the privnet works without initial registration.

But in production, we need to do this manually.

consensus/dbft/dbft.go
txhsl: Sounds good.

consensus/dbft/dbft.go
txhsl: That is also the reason we need a `reset` option in antimev commands. When a CN gets stopped and becomes out-of-date, the keystore can not be easily synced since `handleDKG` is not enforced in state processing but only works when dBFT starts "mining".

consensus/dbft/dkg.go
txhsl: The contract will accept and update the data.

After a deadline, no share or reshare will be accepted, but before that, the uploaded data can be updated again and again.

cmd/geth/antimevcmd.go
txhsl: Ref https://github.com/bane-labs/go-ethereum/pull/332#discussion_r1870567422.

When a node stops dBFT "mining" for a long time, then the keystore becomes out-of-date. The DKG moves forward, but the node still thinks it is the latest, so its keystore will get stuck, its sharing transaction will be rejected.

And if the node is elected to be a CN, then the new DKG round gets failed and revert.

> I thought that we have `reshare` phase specifically for cases when shearing is not successful.

Reshare and share are independent, reshare can be `recovered` by 5/7, but share needs 7/7 correct.

So my solution is to clean up the keystore data and sync from the latest. The node's local secret will be deleted as well, but DKG resharing has a tolerance of 2 failed CN, they can be recovered by the other 5.

antimev/keystore.go
txhsl: Anything in DKG and AMEV is sorted the same way as in the Governance contract. I will.

consensus/dbft/dkg.go
txhsl: Ref https://github.com/bane-labs/go-ethereum/pull/332#discussion_r1869330592.

Yes, the world state of DKG is maintained in contract, but the local DKG progress of the node itself is not.

Try image such a case, I'm in the first period of DKG, so I send a share transaction and a reshare transaction. But I reboot the node, without any record in database, I don't know where I'm when I connect again because the state is not continuous. All the local state is maintained in memory in the first implementation, so we may lose them easily.

The snapshots can help this situation, but yes, not perfectly solve it.

If the current implementation looks quite heavy with too much rarely used code, I’d say move the integration to block state processor is also an option.

consensus/dbft/dkg.go
txhsl: It is possible that the user inits a new keystore and change his message key, then the shares and reshares in contract are not decryptable.

consensus/dbft/dkg.go
txhsl: Just `recover operated`.

antimev/keystore.go
txhsl: > I think we need a check for that at the dBFT level as far. A call to ReceiveSecretShare should be explicitly skipped for WatchOnly nodes.

We have.

> Is there a way for non-DKG-member to decrypt message or construct signature based on received shares of other valid GKG-members?

`AggregateAndDecryptWithShare` and `AggregateAndVerifySig` only need the global public key to use, regardless of be a CN or not. So any node enabled dBFT mining should be able to use them.

antimev/keygroup.go
txhsl: `selfIndex` will be 0 if an address is not in the DKG participant list, it comes from `thresholdKeyGroup.holderIndex()`.

I didn't add a new one, it just replaced the following codes.

```golang
// Compute local secret key
if tkg.holderIndex(tkg.selfAddr) > 0 {
	tkg.localPrvKey = tpke.NewPrivateKey(tkg.receivedSecrets)
}
```

> If we're talking about watch-only node, then this path should probably be handled at the dBFT level.

Do you mean, let dBFT, instead of the keystore, maintain the address list of each DKG group?

antimev/keystore.go
txhsl: > Could you please name the cases when LastGlobalPubKey is nil?

Before the second round DKG succeeds, it's nil. I'm not sure if this `LastGlobalPubKey()` is necessary in signature verification, because I export it to allow a threshold decryption on past old envelope tx in mempool (we don't have implemented it yet).

antimev/signature.go
txhsl: Because I'm not sure if there is a concurrent call, e.g. signing and decrypt, decrypt and encrypt, dkg and signing, dkg and decryption. If there is not, then let's remove it.

consensus/dbft/dbft.go
txhsl: Anyway, reverted.

antimev/keygroup.go
txhsl: > Do you mean, let dBFT, instead of the keystore, maintain the address list of each DKG group?

It is possible, but what we will do is moving some codes from antimev to dbft.

antimev/keystore.go
AnnaShaleva: > I'm not sure if this LastGlobalPubKey() is necessary in signature verification

It's definitely necessary during NextConsensus construction, for NextConsensus we need both current (fresh) and previous (old) keys.

antimev/keygroup.go
AnnaShaleva: > Do you mean, let dBFT, instead of the keystore, maintain the address list of each DKG group?

Not exactly. Maintain the address list at the antimev level is good. Initially I meant DKG-related operations for watch-only nodes, but seems that everything is OK here, because of https://github.com/bane-labs/go-ethereum/pull/332#discussion_r1870621085.

antimev/keystore.go
AnnaShaleva: > AggregateAndDecryptWithShare and AggregateAndVerifySig only need the global public key to use

Have you tested the DKG setup process for non-CN node? Correct me if I'm wrong, but in general non-CN node should be able to watch the new blocks and once shearing phase is finished, the node should get an access to global public key to keep the proper state of its anti-MEV keystore.

antimev/signature.go
AnnaShaleva: From what I see, every DKG-related operation is performed sequentially. Please, double-check it, may be I'm missing something. From my side I can guarantee that all dBFT callbacks are sequential (as far as `postBlock`).

consensus/dbft/dkg.go
AnnaShaleva: > I don't know where I'm when I connect again because the state is not continuous. 

But you may easily call KeyManagement contract and see if KeyManagement has data that you've previously sent (share and reshare data). If KeyManagement has these data, then the node may reuse it to rebuild its local keystore state. Can it?

consensus/dbft/dkg.go
AnnaShaleva: May be then something like `RecoverDone` (and similarly for the rest of bool flags)? It's a nitpicking, but still.

params/config.go
AnnaShaleva: BTW, may be we may store `H` in the contract storage, then only NeoXAMEV fork is required to be set in config, and for privnet we'll use a small value for `H`. But this way is much more complicated, because we need a state to check is hardfork enabled, and I'm not sure that it's always possible.

Hence, I'm OK with two `NeoXDKGBlock` and `NeoXAMEVBlock` in config. But let's wait for Roman's opinion.

consensus/dbft/dbft.go
AnnaShaleva: And that's why I'm against of additional persistance for DKG. We already have KeyManagement contract with its state, and it should be possible to rebuild local node's amev keystore based on KeyManagement state. Then, once you move `handleDKG` to `postPersist`, then it will be called exactly when the node is "ready" for operation. Let's firstly see the outcome of https://github.com/bane-labs/go-ethereum/pull/332#discussion_r1871148138 and then decide on this conversation.

consensus/dbft/dkg.go
AnnaShaleva: > The contract will accept and update the data.

Then I'd say that the way of tracking "if transaction is successful" may be adjusted a bit. Right now this code checks the execution status of known transaction that was sent ~3 blocks ago. If execution status is not OK or transaction is not yet accepted, then retry happens. Instead of this, we may track the desired side-effects of the transaction. For example, if transaction submits share data to KeyManagement, then we should check if KeyManagement has share data persisted in the storage. If not, then retry.

It's just a suggestion, but I consider side-effects tracking as more reliable way to maintain this service. Because there might be interrupts, or transaction may get lost or something else happens which prevent KeyManagement from accepting share. If the node tracks side-effect, then it will retry until KeyManagement has proper data. We use the same approach for NeoFS system contracts auto-deploy procedure, because it has to be reliable.

But the problem that it's relatively hard to implement properly and it requires a lot of testing, so it's just a suggestion, I don't insist on implementation. Overall, current scheme with transactions tracking LGTM.

antimev/keygroup.go
txhsl: Implemented, but not tested yet, I will do it tomorrow. I think this will reduce the keystore size and execution time.

antimev/keystore.go
txhsl: > Have you tested the DKG setup process for non-CN node?

We have. But only watch-only nodes can get synced, seed nodes don't.

antimev/signature.go
txhsl: DKG itself is moving forward step-by-step, so no worries.

consensus/dbft/dkg.go
txhsl: OK, let me try it.

consensus/dbft/dkg.go
```diff
@@ -0,0 +1,833 @@
...
+				return fmt.Errorf("failed to task DKG reshare recover, err: %v", err)
+			}
+		}
+		c.snapshot.ReshareRecoverOped = true
+	}
+	return c.snapshot.store(c.db)
+}
+
+// loopTaskList retries every task in tx watch list
+func (c *DBFT) loopTaskList(header *types.Header) {
```
AnnaShaleva: I've just noticed one potential problem: from what I see `loopTaskList` is run synchroniousely (not in a separate routine). We've decided to move `handleDKG` to the `postBlock` callback, but the problem is that the node can't wait until all transactions are accepted in `loopTaskList`, because it blocks other blocks processing and dBFT functioning itself. I'd say it should be moved to a separate routine, but we need to be careful with `loopTaskList` communication with the main routine.

antimev/keystore.go
txhsl: So, the thing is same as governance epoch change, only after the contract is persisted, and after `postBlock`, we can have the new key.

consensus/dbft/dkg.go
AnnaShaleva: The overall idea is to avoid introducing the DKG persistence DB at all, and use KeyManagement state for everything state-related. But may be I'm just not aware of some details and it's just not possible. Check it please.

antimev/keystore.go
AnnaShaleva: >  seed nodes don't.

It's OK, they should not.

consensus/dbft/dkg.go
txhsl: Implemented, going to test.

cmd/geth/antimevcmd.go
txhsl: Ref https://github.com/bane-labs/go-ethereum/pull/332#discussion_r1872725951. Most of the case may get solved by dBFT (that code will do the reset), but I still leave the command here in case we need it.

core/state_processor.go
txhsl: Merged.

params/config.go
AnnaShaleva: We've discussed it with Roman, the result is we're good with two forks. But we have one additional suggestion regarding Testnet/Mainnet: 
1. Since we inevitably have two forks, let's firstly schedule and enable only `NeoXDKG` (and keep `NeoXAMEV` disabled). Let's wait for successful `NeoXDKG` transition and wait for at least several DKG rounds to pass successfully. If needed, we may even register/vote for new CN on Testnet and check that DKG setup is properly handled for several rounds and for updated set of CNs.
2. Once we're sure that `NeoXDKG` fork works properly and DKG passes successfully and CNs have their amev keystores in the proper state, then we'll schedule `NeoXAMEV` fork to particular height in the future. `NeoXAMEV` enables encrypted transactions processing and threshold block signatures.

So the main idea is to firstly try `NeoXDKG`, ensure that it works. and only after that schedule `NeoXAMEV` height.

@txhsl please discuss it with @steven1227 and write your solution.

consensus/dbft/dkg.go
AnnaShaleva: We discussed this issue with Roman, the final decision is to try to implement this now, while we're working on this PR. Side-effects tracking is more reliable way, and it works inevitably even if the CN was down for a while. We may perform tracking in a separate routine that will subscribe for new blocks (`c.chain.SubscribeChainHeadEvent`), receive Block notifications, check for KeyManagement storage updates and see if it contains proper data.

@txhsl do you think it's possible/worth to implement in the same PR?

params/config.go
txhsl: That's the best solution we can have, I'm OK with it.

And the good news is we don't have to change anything so far in this PR. What we need is to set `NeoXAMEVBlock` to `nil` in the genesis file, and I think `IsNeoXAMEV` will always return `false` before we give it a value.

consensus/dbft/dkg.go
txhsl: > We may perform tracking in a separate routine that will subscribe for new blocks (c.chain.SubscribeChainHeadEvent), receive Block notifications, check for KeyManagement storage updates and see if it contains proper data.

It is definitely worth, I have the same idea as well. For the retry list, we have ZK task to work on quite following, so there undoubtedly will be another PR to change this. It might be a new routine as well, so a good start in this PR will be helpful.

Let me try to implement it. Just too many changes here already, I'd be careful.

params/config.go
AnnaShaleva: Agree. But we still need both `NeoXBKGBlock: X` and `NeoXAMEVBlock: Y` enabled in privnet for testing. And we need this "hack" that you've implemented for KeyManagement storage to properly setup privnet antimev keystore and to be able to test transactions decryption on privnet.

consensus/dbft/dkg.go
txhsl: No, just wait for three blocks before checking the receipt.

consensus/dbft/dbft.go
txhsl: Sorry, but it's not working, seems the block state in `WithProcessBlock` when `OnCommit` is nil, so I got a panic.

txhsl:
Merged `bane-main` into this branch. Still several things to do,

- [x] Fix missing state when `postBlock`, ref https://github.com/bane-labs/go-ethereum/pull/332#discussion_r1869364276;
- [x] Rework transaction retry task, ref https://github.com/bane-labs/go-ethereum/pull/332#discussion_r1871194018 and https://github.com/bane-labs/go-ethereum/pull/332#discussion_r1871182099;
- [x] Correct DKG index fetching in dBFT, ref https://github.com/bane-labs/go-ethereum/pull/332#discussion_r1877896702.

consensus/dbft/dbft.go
```diff
@@ -1140,8 +1156,12 @@ func (c *DBFT) getBlockWitness(pub *tpke.PublicKey, block *Block) ([]byte, error
 		// Take all available shares since some of them may be invalid.
 		for i := 0; i < len(vals); i++ {
 			if p := dctx.CommitPayloads[i]; p != nil && p.ViewNumber() == dctx.ViewNumber {
+				dkgIndex := slices.Index(c.dkgSnapshot.CurrentCNs, vals[i]) + 1
```
txhsl: @chenquanyu I didn't find a good way to use `getDKGIndex` properly, need a fix. Also, should we use the latest height or just height-1?

https://github.com/bane-labs/go-ethereum/blob/d780f6912e9d27c808f341be419f7bcb071681e4/consensus/dbft/dbft.go#L884

chenquanyu: I think `height -1` should be used, because the current consensus list should be based on the result of the previous block.

txhsl: Fixed.

txhsl:
@songb2 Now the DKG, threshold signature and dynamic envelope transaction should work properly, please feel free to test them before we merge. To know about the details, please also refer to related issues that will be closed by this PR.

consensus/dbft/dkg.go
chenquanyu: I think we can put the whole `handleDKG` method into a separate routine so that it does not block `postBlock`. In this case, we need to add a `Mutex` to protect the state of `dkgSnapshot` and `amevKeystore`.

txhsl: > In this case, we need to add a `Mutex` to protect the state of `dkgSnapshot` and `amevKeystore`.

If we only move `loopTaskList`, then we don't.

consensus/dbft/dkg.go
chenquanyu: >  Side-effects tracking is more reliable way, and it works inevitably even if the CN was down for a while.

Do you mean we need to get share pvss and messages from the contract to verify whether the data is consistent with the local node or do we only need to verify the existence of the corresponding data to confirm the success of the share transaction? In my opinion, the successful transaction status can confirm that there is no need to resend the transaction, because currently each type of transaction is only constructed once, and as long as the transaction is successfully uploaded to the chain, it proves that there is corresponding data in the contract.

consensus/dbft/dkg.go
chenquanyu: > If we only move `loopTaskList`, then we don't.

We may also need to protect `txWatchList`.

txhsl: > We may also need to protect `txWatchList`.

The communication should be maintained by channel, then `txWatchList` is isolated with main routine.

consensus/dbft/dkg.go
txhsl: @AnnaShaleva We now have the `loopTaskList` in a separate routine, and it seems working as expected.

For the side effects tracking, we've referred the implementation of NeoFS, but the data here we have to verify against contract are more than necessary, and our case is a bit different (there is a deadline of period). Our DKG only needs following checking, and they are checked by `KeyManagement`.

- The contract has enough data (7 pieces) for both sharing and resharing when epoch change;
- These data are uploaded by permitted DKG participants.

If the above requirements are not satisfied, then the DKG reverts, as well as local keystores (offline ones will be overwritten by the next round, or sync to the latest when they connect again).

The only case I can image that causes an inconsistency is, a node first sends a transaction and uploads data correctly, then gets shut down and reboot. Without a side effects tracking, the node will send a transaction again to upload, but that one fails (e.g. keep pending/computing ZK till the deadline).

Well, but a side effects tracking will not solve this case thoroughly, if the first transaction is pending in the mempool, and the node reboots before this transaction get confirmed, it will still send the second transaction (the pool nonce will increase, so it will not be rejected by transaction API). And if the second transaction fails, then we have the same result.

We still have 1/3 fault tolerance here, so the current loose check still works. If we can't find a perfect way to implement this kind of side effects check, especially without a DB persistence, then I suggest delaying it to a future PR.

txhsl: Ref #367.

params/config.go
txhsl: I just tried to set `neoXDKGBlock` to some value and `neoXAMEVBlock` to `nil`, and it returned a `failed to create dBFT engine` error, with a message `NewPreBlockFromContext is set, but AntiMEVExtensionEnablingHeight is not specified`.

I think the root reason is the `checkConfig()` in the dBFT package, ref https://github.com/nspcc-dev/dbft/blob/be64b5f47672278bf4763d1eeb1e34cb2e7ad71f/config.go#L167, maybe we should open an issue for this.

songb2:
I tested that the pr looks good now.
I tested this pr with some cases like nodes become offline, switch to new censensus nodes and so on, they all passed.

txhsl:
Will implement https://github.com/bane-labs/go-ethereum/issues/367 in a further PR, but now things work quite well for most cases, let's get this merged first.

params/config.go
txhsl: Opened #370.

PR #334, opened 2024-10-22
antimev: update keystore behaviors for integration

txhsl:
@chenquanyu Update the keystore behaviors according to #312 and the implementation in branch antimev-integration-contract`.

- `OnValidatorList` starts a DKG and `OnEpochChange` ends it;
- `ReceiveSecretShare`, `ReceiveSecretReshare`, `ReceiveRecoverShare` and `ReceiveRecoveredReshare` handle different kinds of DKG messages;
- `StartRecover` starts a recovering period and `TryRecoverReshare` tries to get a result from recovery;
- Any CN can call above methods in correct periods without error, but the behaviors are different, depends on their roles in DKG:
  - If the caller is not a sender of sharing, resharing or recovering, the returned messages of `OnValidatorList` and `StartRecover` will be `nil`;
  - If the caller is not a receiver of sharing, resharing or recovering, only public PVSS of `ReceiveSecretShare`, `ReceiveSecretReshare` and `ReceiveRecoveredReshare` will be handled. In `OnEpochChange`, only public keys will be generated in keystore, but the local private key will be `nil`.

### DIFF
--- a/antimev/keygroup.go
+++ b/antimev/keygroup.go
@@ -185,10 +185,12 @@ func (tkg *thresholdKeyGroup) dkgReshare(target *thresholdKeyGroup) ([][]byte, *
 	return generateShareMessages(tkg.localSecret.Renovate(), target.holders)
 }
 
+// dkgReshareRecovered tries to recover a dkg secret, and returns an error if
+// the attempt fails when shares are not enough.
 func (tkg *thresholdKeyGroup) dkgReshareRecovered(threshold int, target *thresholdKeyGroup) ([][]byte, *tpke.PVSS, error) {
-	// Do nothing if not a receiver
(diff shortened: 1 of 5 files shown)


antimev/keygroup.go
```diff
@@ -185,6 +185,8 @@ func (tkg *thresholdKeyGroup) dkgReshare(target *thresholdKeyGroup) ([][]byte, *
 	return generateShareMessages(tkg.localSecret.Renovate(), target.holders)
 }
 
+// dkgReshareRecovered tries to recover a dkg secret, and returns an error if
+// there is no secret gets recoverd.
```
vncoelho: there is no secret gets recoverd 
there is no secret that gets recovered.

chenquanyu (approved):

PR #337, opened 2024-10-25
contracts: privnet: add an initializer in Governance for DKG period setting and initial values to privnet genesis files

### LINKED ISSUE #336: Add an initializer for the DKG period setting in Governance contract
# Rationale

The current deployed Governance contracts on both Mainnet and Testnet don't have an initial value for `sharePeriodDuration` setted by genesis file. An initializer is necessary.

# Implementation

Add a setter method in Governance that can be only executed by `GovProxyAdmin`'s `scheduleUpgrade` through `upgradeAndCall`.

txhsl:
Close #336. The duration of DKG sharing and recovery for privnet are both `10 blocks`, so `10 blocks` free, `10 blocks` sharing, and `10 blocks` recovery.


chenquanyu (approved):

AnnaShaleva (commented):
It's already merged, but still LGTM, we need some way to set period duration for existing network.

PR #339, opened 2024-10-30
dbft: properly verify all types of transactions

### LINKED ISSUE #314: Properly verify transacctions in `VerifyPrepareRequest`/`ProcessPreBlock`/`WithNewBlockFromContext` dBFT callbacks
1. #287 moves block state processing from `VerifyPrepareRequest` to `ProcessPreBlock` which means that in-block transactions are not verified in any way in PrepareRequest. Verification must be performed for initial set of transactions even if some of them are Envelopes. Verificaiton may be performed either by state processing (`ChainHeaderReader.ProcessState` returns an error if there are some troubles with tx nonces/balances/etc) or by allocating new mempool and adding proposed transactions to this pool one-by-one.
2. #307 introduces legacy pool into dBFT engine so that verification of decrypted transactions becomes possible. However, it's not correct to use shared node's pool to verify proposed transactions because pool content may have conflicts with block content. We need to consider allocating a new pool in-place every time when we need to process PreBlock and check PreBlock's transaction using this isolated pool.
3. #287 processes state of final `Block` with decrypted transactions and, if error is returned, then the node stops operation. We should consider to replace this behaviour: if error happens, then we may fall back to initial set of transactions (with Envelopes) and process them instead of decrypted transactions. It's better than nothing.

AnnaShaleva:
1. Remove verification with the help of the node's mempool. It's not correct to use an existing memory pool for in-block transactions verification, because node's pool may contain some pending transactions. Close 2-nd point from #314.
2. Use local mempool to check transactions before state processing at PreBlock verification and PreBlock processing levels. Close 1-st point from #314.
3. Properly verify Envelopes and Decrypted transactions. Dynamically rebuild the transactions list if some of Decrypted transactions are invalid or conflict with existing in-block transactions. Close 3-d point  from #314.
4. Fix bugs in Decrypted transactions iteration index. These bugs wehre discovered during my privnet testing.

Close #314. Depends on https://github.com/nspcc-dev/dbft/pull/133.

TODO:
- [X] Manage a single local mempool instance instead of allocating new local mempool every time it's needed. The caveat is mempool reset, because Geth's mempool is not designed for multiple resets at the same height. Ref. https://github.com/bane-labs/go-ethereum/issues/341.
- [X] Properly setup pool's parameters in order not to conflict with the chain's/node's pool settings.
- [X] Resolve minor TODOs in code.
- [X] Test real envelopes processing on privnet, ensure that different combinations of envelopes/standard txs are covered.

@txhsl, @chenquanyu, @songb2 ready for review and tests.


go.sum
```diff
@@ -493,6 +493,8 @@ github.com/nats-io/nkeys v0.0.2/go.mod h1:dab7URMsZm6Z/jp9Z5UGa87Uutgc2mVpXLC4B7
 github.com/nats-io/nuid v1.0.1/go.mod h1:19wcPz3Ph3q0Jbyiqsd0kePYG7A95tJPxeL+1OSON2c=
 github.com/nspcc-dev/dbft v0.3.1-0.20240912152105-ba329883f471 h1:EpFQ5tHFDuGPdMI0CK31MuH5YB5WDPHf7czF1pMVh64=
 github.com/nspcc-dev/dbft v0.3.1-0.20240912152105-ba329883f471/go.mod h1:BNvJkPKTE28r+qRaAk2C3VoL2J9qzox3fvEeJbh7EWE=
+github.com/nspcc-dev/dbft v0.3.1-0.20241030134634-5d6954c9e5f9 h1:ZWHE/ZqPNxm1TEh4LS2ToKptbjHCFxW/JtlCRUbqLGQ=
+github.com/nspcc-dev/dbft v0.3.1-0.20241030134634-5d6954c9e5f9/go.mod h1:BNvJkPKTE28r+qRaAk2C3VoL2J9qzox3fvEeJbh7EWE=
```
roman-khimov: go mod tidy

consensus/dbft/preblock.go
```diff
@@ -23,10 +24,16 @@ type PreBlock struct {
 	// envelopesData is the ordered list of decoded content of Envelopes
 	// transactions of the proposed PreBlock.
 	envelopesData []envelopeData
+
```
roman-khimov, line 27: :eyes:

consensus/dbft/preblock.go
```diff
@@ -23,10 +24,16 @@ type PreBlock struct {
...
 	// finalTransactions is the cached final list of transactions formed after TPKE
 	// decryption of Envelopes content. This list includes both simple standard and
 	// decrypted transactions.
 	finalTransactions []*types.Transaction
+	// final* fields below represent an information got after finalTransactions
```
roman-khimov: s/an//

consensus/dbft/dbft.go
```diff
@@ -667,7 +657,32 @@ func New(chainCfg *params.ChainConfig, _ ethdb.Database) (*DBFT, error) {
 				return false
 			}
 			ethBlock := dbftBlock.ToEthBlock()
-			_, _, err := c.chain.VerifyBlock(ethBlock, false)
+
+			// TODO: consider reusing a single instance of local pool bound to the engine.
```
roman-khimov: It's an optimization that can be left for another issue.

consensus/dbft/dbft.go
```diff
@@ -667,7 +657,32 @@ func New(chainCfg *params.ChainConfig, _ ethdb.Database) (*DBFT, error) {
...
+			localPool := legacypool.New(legacypool.Config{
+				NoLocals: true,
+				Journal:  "",
+			}, c.chain)
+			localPool.Init(1, parent, func(addr common.Address, reserve bool) error {
+				return nil
+			})
+
+			errs := localPool.Add(dbftBlock.transactions, false, true) // TODO: consider replacing sync:true by sync:false, because it's not for prod usage?
```
roman-khimov: Given that the pool is empty there shouldn't be a lot of difference. Also, for our use case we should be ok with `false`, we only need to validate the set and it's done anyway, `Add()` will refuse invalid transactions.

consensus/dbft/dbft.go
```diff
@@ -754,86 +769,209 @@ func New(chainCfg *params.ChainConfig, _ ethdb.Database) (*DBFT, error) {
 		}),
 		dbft.WithProcessPreBlock(func(b dbft.PreBlock[common.Hash]) error {
 			var (
-				ctx = c.dbft.Context
-				pre = ctx.PreBlock().(*PreBlock)
+				ctx             = c.dbft.Context
+				pre             = b.(*PreBlock)
+				hasDecryptedTxs bool
 			)
+
```
roman-khimov: :eyes:

consensus/dbft/dbft.go
```diff
@@ -754,86 +769,209 @@ func New(chainCfg *params.ChainConfig, _ ethdb.Database) (*DBFT, error) {
...
+			// Process state with constructed list of transactions to fill state-dependent
+			// block fields.
+			finalBlock := &PreBlock{
+				header:       pre.header,
+				transactions: pre.finalTransactions,
+				withdrawals:  pre.withdrawals,
+			}
+			ethBlock := finalBlock.ToEthBlock()
+			pre.finalState, pre.finalReceipts, _, pre.finalGASUsed, err = c.chain.ProcessState(ethBlock, nil)
```
roman-khimov: Can't this be avoided if `!hasDecryptedTxs`?

consensus/dbft/dbft.go
```diff
@@ -843,6 +981,31 @@ func New(chainCfg *params.ChainConfig, _ ethdb.Database) (*DBFT, error) {
 	return c, nil
 }
 
+// ValidateDecryptedTx checks the validity of the transaction to determine whether the outer envelope transaction should be replaced.
+func (c *DBFT) validateDecryptedTx(head *types.Header, decryptedTx *types.Transaction, envelope *types.Transaction) error {
+	// Make sure the transaction is signed properly and has the same sender and nonce with envelope
+	if decryptedTx.Nonce() != envelope.Nonce() {
+		return fmt.Errorf("decryptedTx nonce mismatch: decryptedNonce %v, envelopeNonce %v", decryptedTx.Nonce(), envelope.Nonce())
+	}
+	// Ensure the gasprice is high enough to replace the envelope transaction
```
roman-khimov, line 990: Economic requirements like this need to be checked against our economic model. @txhsl?

txhsl: We only define a minimum gas tip and a static base fee, and it's still an Ethereum price market, so this rule of replacement LGTM.

But we may need a tip gap to charge the fee of antimev service.

txhsl: As we have discussed, we may add an `antimevFee` to Policy contract, which is evaluated in fixed GAS and gets divided by execution gas to be applied and checked here.

But in this solution, the transaction sender can cheat by declaring a high gas limit. Because here we can only get the `decryptedTx.Gas()` (which is the gas limit) for verification, but it can be very different from the value which is really used by EVM.

So I think it is better to take the same strategy as transaction replacement, which is charging by percentage. It will be a bit "unfair", but is more practical. How do you think? @steven1227 @doubiliu @AnnaShaleva @roman-khimov.
https://github.com/bane-labs/go-ethereum/blob/1abb2c7c192e0d4e1a6c8cd1684aa346e7546304/core/txpool/legacypool/list.go#L305-L328

go.sum
AnnaShaleva: Fixed.

consensus/dbft/preblock.go
AnnaShaleva: It's intentionally to split `final*` fields from the rest of block fields.

consensus/dbft/preblock.go
AnnaShaleva: Fixed.

consensus/dbft/dbft.go
AnnaShaleva: https://github.com/bane-labs/go-ethereum/issues/341.

consensus/dbft/dbft.go
AnnaShaleva: I reviewed the pool's code, you're right, `false` works OK for us.

consensus/dbft/dbft.go
AnnaShaleva: It's also intentionally, but fixed.

consensus/dbft/dbft.go
AnnaShaleva: Sure it can, fixed.

consensus/dbft/dbft.go
AnnaShaleva: > Economic requirements like this need to be checked against our economic model. 

I moved this code almost as is from the existing implementation, it was added in https://github.com/bane-labs/go-ethereum/pull/307.

But in general, I think that it should be carefully discussed, and even moved to a separate issue. Let's use https://github.com/bane-labs/go-ethereum/issues/342 for further conversations.

AnnaShaleva:
Ready for review, privnet tests are still in progress.

AnnaShaleva:
Works with simple transactions; testing on Envelopes is in progress.

AnnaShaleva:
@roman-khimov, @txhsl, @chenquanyu ready for review and merge. Tested with the help of generator (I attached a naive version of this generator as a test in a separate commit). I tested "good" cases of encrypted transaction acceptance (single decrypted tx per block, various combinations of decrypted transaction with standard transactions), e.g.:
```
INFO [11-06|19:57:06.342] Envelope data decrypted                  "envelope hash"=dd1a60..d7eb1a "envelope index"=0 data=f869808509502f90008252089474f4effb0b538baec703346b03b6d9292f53a4cd0180850113a4538ea09f03664a56ae11cd9cfb5efedc1648c374d51ae91982f7b836893383e4c52f0aa067c4771d614d7c296aecddfd25aae6fab3276e8a48114f87953b5c03c75bad76
...
2024-11-06T19:57:06.354+0300	INFO	dbft@v0.3.1-0.20241030192742-82ad77871e88/check.go:120	approving block	{"height": 9, "hash": "0x543c0f2757f30e4df567791feb557d6f4ffd3dc57b0a15aa5faff5ac41fda57e", "tx_count": 2, "merkle": "0xa3b72c19a67f838b365e46c859e9d1c50e2e901db16fe8ca1b043fac9da66a98", "prev": "0x1f5164e5df314e5af51c630332effc17ef7d11e62ba3839060d1698d9a37febd"}
INFO [11-06|19:57:06.363] Successfully wrote new block with state  number=9 hash=368799..0059ba
```
And I tested some basic error cases where Envelope is accepted instead of its decrypted content because of various verification errors, like this:
```
2024-11-06T20:11:37.311+0300	INFO	dbft@v0.3.1-0.20241030192742-82ad77871e88/check.go:66	processing PreBlock	{"height": 5, "view": 0, "tx_count": 2, "preCommit_count": 3}
INFO [11-06|20:11:37.315] Envelope data decrypted                  "envelope hash"=6c38cc..dc2681 "envelope index"=1 data=f869808509502f90008252089474f4effb0b538baec703346b03b6d9292f53a4cd01808501139e3377a06e45723e9e6dd4e4123209e93cbba0bc5a9c5aa0563ff9cb19434b104c718881a017a065da644f69570773a97594e4adb2f325d09c10afebe114b986d4e80e3695
INFO [11-06|20:11:37.316] Decrypted transaction is invalid, fallback to envelope "envelope hash"=6c38cc..dc2681 "envelope index"=1 data=f869808509502f90008252089474f4effb0b538baec703346b03b6d9292f53a4cd01808501139e3377a06e45723e9e6dd4e4123209e93cbba0bc5a9c5aa0563ff9cb19434b104c718881a017a065da644f69570773a97594e4adb2f325d09c10afebe114b986d4e80e3695 error="decryptedTx nonce mismatch: decryptedNonce 0, envelopeNonce 1"
...
2024-11-06T20:11:37.318+0300	INFO	dbft@v0.3.1-0.20241030192742-82ad77871e88/check.go:120	approving block	{"height": 5, "hash": "0xf1ee4bd98b06184c80187f81af1f1913d32e3097122481413ac8037bd66bc747", "tx_count": 2, "merkle": "0x3604e8be2a0108449f0bcf1283a3fdd911aa70e62e0b105929f3f29b8ea4b99d", "prev": "0x0d883be70ec9ac1d21b5d428ef4fb1b1b000a32d4415ffe20107934740c0521c"}
```
So at least it works for "correct" envelopes/encrypted transactions and woks for Envelopes with malformed data inside. But it's crucial to make tests for mempool-related verification code, I'm leaving this task to @songb2.

consensus/dbft/dbft.go
```diff
@@ -754,90 +756,199 @@ func New(chainCfg *params.ChainConfig, _ ethdb.Database) (*DBFT, error) {
...
-					log.Info("Decrypted transaction decoding failed",
-						"envelope hash", pre.transactions[i].Hash(),
-						"envelope index", i,
-						"data", hex.EncodeToString(decryptedTxsBytes[j]),
-						"error", err.Error())
-					txx[i] = pre.transactions[i]
-					j++
-					continue
+					// Some shares are invalid, valid shares isn't enough to decrypt, wait for more shares to be collected.
+					return fmt.Errorf("failed to decrypt Encrypted transactions, not enough valid shares: %w", err)
```
roman-khimov, line 826: Looks like we can't fix it, but from what I see we can get enough precommits and still fail here and thus require more of precommits which potentially can never happen (node dead, for example). This makes the protocol somewhat weaker. We can theoretically just move on with unencrypted transactions, but then the problem is that someone might assemble the right combination and we gonna have a different interpretation of the same transaction set. Do I understand correctly that this happens when node sends an invalid share in an otherwise correct precommit? And we have no way to check it at the individual message level?

consensus/dbft/dbft.go
```diff
@@ -754,90 +756,199 @@ func New(chainCfg *params.ChainConfig, _ ethdb.Database) (*DBFT, error) {
...
+						if errs[0] != nil {
+							txx = pre.transactions
+							hasDecryptedTxs = false
+							break
+						}
+						txx[i] = pre.transactions[i]
+						continue
+					}
+					if decryptedTxsBytes[j] == nil { // pre.transactions[i] is Envelope, but its content failed to be decrypted, use Envelope as-is.
+						errs := localPool.Add([]*types.Transaction{pre.transactions[i]}, false, false)
```
roman-khimov: Duplicates a lot of the `if` above, can be merged.

consensus/dbft/dbft.go
```diff
@@ -754,90 +756,199 @@ func New(chainCfg *params.ChainConfig, _ ethdb.Database) (*DBFT, error) {
...
+						"data", hex.EncodeToString(decryptedTxsBytes[j]))
+					var decryptedTx = new(types.Transaction)
+					err := decryptedTx.DecodeRLP(rlp.NewStream(bytes.NewReader(decryptedTxsBytes[j]), 0))
+					if err != nil {
+						log.Info("Decrypted transaction decoding failed, fallback to envelope",
+							"envelope hash", pre.transactions[i].Hash(),
+							"envelope index", i,
+							"decrypted data", hex.EncodeToString(decryptedTxsBytes[j]),
+							"error", err.Error())
+						errs := localPool.Add([]*types.Transaction{pre.transactions[i]}, false, false)
```
roman-khimov: Maybe this can be a local function.

consensus/dbft/dbft.go
```diff
@@ -754,90 +756,199 @@ func New(chainCfg *params.ChainConfig, _ ethdb.Database) (*DBFT, error) {
...
+							txx = pre.transactions
+							hasDecryptedTxs = false
+							break
+						}
+						txx[i] = pre.transactions[i]
+						j++
+						continue
+					}
+					errs := localPool.Add([]*types.Transaction{decryptedTx}, false, false)
+					if errs[0] != nil {
```
roman-khimov, line 899: Or maybe you can chain the happy path into a single `if` in some way since most of unhappy ones are the same.

songb2:
@AnnaShaleva Below test case failed
```
--- FAIL: TestGenerateEncryptedTx (0.55s)
    tpke_test.go:137: 
                Error Trace:    /home/bruce/new/main/go-ethereum/antimev/tpke_test.go:137
                Error:          Received unexpected error:
                                could not decrypt keystore
                Test:           TestGenerateEncryptedTx
FAIL
FAIL    github.com/ethereum/go-ethereum/antimev 549.168s
```

consensus/dbft/dbft.go
AnnaShaleva: > require more of precommits which potentially can never happen (node dead, for example)

dBFT tolerance to F faulty nodes is still preserved. It's just that for NeoX case the definition of "faulty" node is extended a bit: we consider node as "faulty" if it sends PreCommit that contains malformed share. `dbft` library itself properly handles this case.

> This makes the protocol somewhat weaker.

I won't say so. Same guarantees if M out of N nodes are "good".

> We can theoretically just move on with unencrypted transactions

Only if PreCommits from all N nodes are received and it's still impossible to decrypt the messages. But it means that more than F nodes are faulty. I may add this fallback path, but do we really need to continue if more than F nodes are faulty?

> Do I understand correctly that this happens when node sends an invalid share in an otherwise correct precommit? 

Exactly.

> And we have no way to check it at the individual message level?

Absolutely no way. Moreover, we can't say for sure which PreCommit is invalid. We can only say that a particular subset of PreCommits can (or can't) decrypt the message.

consensus/dbft/dbft.go
AnnaShaleva: I made it intentionally to make the code more readable, but OK, let's refactor it.

consensus/dbft/dbft.go
roman-khimov: The difference is that previously M (out of N) messages were enough to move on, now it can be insufficient. But yeah, if there are M good shares we're OK and there should be M of them anyway. So OK, let's keep it as is without introducing any additional complexity.

consensus/dbft/dbft.go

consensus/dbft/dbft.go

consensus/dbft/dbft.go
AnnaShaleva: It looks better with local function.

AnnaShaleva:
> Below test case failed

It's because the list of 7-privnet CN accounts is modified by other unit-tests. If run this test in a stand-alone mode, it will pass, and I just didn't run it concurrently with other tests. Fixed by 6f5905034622d52632ae735eee69e934292f8b10.

roman-khimov (approved):

PR #358, opened 2024-12-04
tpke: take some care of Signature and PublicKey

### LINKED ISSUE #355: Migrate block signatures onto compressed TPKE public key form
# Rationale

Ref. https://github.com/bane-labs/go-ethereum/pull/347#discussion_r1861449326.

# Implementation

The implementation itself is quite straightforward, see 773028baf1de50a68c64e7267867de9af8a5121f. But it needs cache, otherwise it's useless.

AnnaShaleva:
Add a bit of documentation, fix formatting, extend the list of constants and close #355.

Required for #332. @txhsl, @chenquanyu, @songb2 please review.

### DIFF
--- a/crypto/tpke/public_key.go
+++ b/crypto/tpke/public_key.go
@@ -66,6 +66,24 @@ func NewPublicKeyFromBytes(b []byte) (*PublicKey, error) {
 	return pk, nil
 }
 
+// NewPublicKeyFromPaddedBytes deserializes PublicKey from the given byte slice.
+// It expects uncompressed padded PublicKey representation as an input with length of
+// [PublicKeyLenPadded] (see [PublicKey.Bytes] documentation).
+func NewPublicKeyFromPaddedBytes(b []byte) (*PublicKey, error) {
+	if len(b) != PublicKeyLenPadded {
+		return nil, fmt.Errorf("invalid public key length: expected %d, got %d", PublicKeyLenPadded, len(b))
+	}
+	var (
+		pk  = new(PublicKey)
+		err error
+	)
+	pk.pg1, err = decodePointG1(b)
+	if err != nil {
+		return nil, err
+	}
+	return pk, nil
+}
+
 // Bytes serializes PublicKey into byte slice using compressed [bls12381.G1Affine]
 // representation. The resulting byte slice has the length of
 // [PublicKeyLen].
@@ -78,17 +96,6 @@ func (pk *PublicKey) Bytes() []byte {
 // resulting byte slice has the length of [PublicKeyLenPadded].
 func (pk *PublicKey) Encode() []byte {
 	return encodePointG1(pk.pg1)
-}
-
-// Decode decodes PublicKey from the given byte slice in an uncompressed form with
-// padding. It expects the slice of [PublicKeyLenPadded] length as an input.
-func (pk *PublicKey) Decode(b []byte) (*PublicKey, error) {
-	pg1, err := decodePointG1(b)
-	if err != nil {
-		return nil, err
-	}
-	pk.pg1 = pg1
-	return pk, nil
 }
 
 // Equal compares if two public keys are the same.
--- a/crypto/tpke/public_key.go
+++ b/crypto/tpke/public_key.go
@@ -8,9 +8,15 @@ import (
 	lru "github.com/hashicorp/golang-lru/v2"
 )
 
-// PublicKeyLen is the length of PublicKey in serialized compressed representation.
-const PublicKeyLen = bls12381.SizeOfG1AffineCompressed
+const (
+	// PublicKeyLen is the length of PublicKey in serialized compressed representation.
+	PublicKeyLen = bls12381.SizeOfG1AffineCompressed
+	// PublicKeyLenPadded is the length of PublicKey in a serialized uncompressed
+	// padded representation.
+	PublicKeyLenPadded = 128
+)
 
+// PublicKey is a BLS12-381 based public key implementation used in TPKE scheme.
 type PublicKey struct {
 	pg1 *bls12381.G1Affine // A public value for tpke encryption
 }
@@ -24,7 +30,7 @@ func init() {
 	keycache, _ = lru.New[string, *PublicKey](32)
 }
 
-// NewGlobalPublicKey aggregates and returns a PublicKey
+// NewGlobalPublicKey aggregates and returns a PublicKey.
 func NewGlobalPublicKey(cs []*Commitment, scaler int) *PublicKey {
 	pg1 := new(bls12381.G1Affine).Set(cs[0].coeff[0])
 	// Add up A0
@@ -38,9 +44,9 @@ func NewGlobalPublicKey(cs []*Commitment, scaler int) *PublicKey {
 }
 
 // NewPublicKeyFromBytes deserializes PublicKey from the given byte slice. It expects
-// compressed PublicKey representation as an input (see [PublicKey.Bytes]
-// documentation). It uses cache under the hood, hence returned value must not be
-// changed.
+// compressed PublicKey representation as an input with length of [PublicKeyLen] (see
+// [PublicKey.Bytes] documentation). It uses cache under the hood, hence returned
+// value must not be changed.
 func NewPublicKeyFromBytes(b []byte) (*PublicKey, error) {
 	pk, ok := keycache.Get(string(b))
 	if ok {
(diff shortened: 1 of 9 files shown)


crypto/tpke/public_key.go
```diff
@@ -5,15 +5,25 @@ import (
...
 	pg1 *bls12381.G1Affine // A public value for tpke encryption
 }
 
+// keycache is a simple lru cache for bls12381 keys that avoids Y calculation
+// overhead for known keys serialized in compressed form.
+var keycache *lru.Cache[string, *PublicKey]
+
+func init() {
+	// Less than 100K, probably enough for our purposes.
+	keycache, _ = lru.New[string, *PublicKey](1024)
```
roman-khimov: Probably can be even lower than that, we're using it for CN keys only and we have just 7 CNs usually. Something like 32 maybe.

consensus/dbft/dbft-TODO.md
```diff
@@ -0,0 +1,107 @@
+1. Split NeoXAMEV intto two:
```
roman-khimov: ?

consensus/dbft/x_md
```diff
@@ -0,0 +1,36 @@
+node 1:
```
roman-khimov: ?

consensus/dbft/x2_md
```diff
@@ -0,0 +1,101 @@
...
...51128dbaec554494f58919ee641b7455e8f80464 "uncle hash"=0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347 coinbase=0x1212000000000000000000000000000000000003 root=0x1d490454712c577ace6a93785421a97e1a42c391e57e13c89c46d972ef3da946 "tx hash"=0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421 "receipt hash"=0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421 bloom=00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000 difficulty=2 number=5 "gas limit"=30000000 "gas used"=0 time=1732110126 extra=01 "mix digest"=0xce2c47e402f65838a2251f779218a3e69debb1236a1099d6fb7e65c04a7065f7 nonce=1 "base fee"=20000000000 "withdrawals hash"=0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421
```
roman-khimov: ?

crypto/tpke/public_key.go
```diff
@@ -66,6 +66,24 @@ func NewPublicKeyFromBytes(b []byte) (*PublicKey, error) {
 	return pk, nil
 }
 
+// NewPublicKeyFromUncompressedBytes deserializes PublicKey from the given byte slice.
+// It expects uncompressed PublicKey representation as an input with length of
+// [PublicKeyLenUncompressed] (see [PublicKey.UncompressedBytes] documentation).
+func NewPublicKeyFromUncompressedBytes(b []byte) (*PublicKey, error) {
```
roman-khimov: I think https://pkg.go.dev/github.com/consensys/gnark-crypto/ecc/bls12-381#G1Affine.SetBytes handles uncompressed too, btw.

crypto/tpke/public_key.go
```diff
@@ -74,23 +92,12 @@ func (pk *PublicKey) Bytes() []byte {
 	return res[:]
 }
 
-// Encode encodes PublicKey to byte slice in an uncompressed form with padding. The
-// resulting byte slice has the length of [PublicKeyLenUncompressed].
-func (pk *PublicKey) Encode() []byte {
+// UncompressedBytes encodes PublicKey to byte slice in an uncompressed form with
+// padding. The resulting byte slice has the length of [PublicKeyLenUncompressed].
+func (pk *PublicKey) UncompressedBytes() []byte {
```
txhsl: `encodePointG1` is not the uncompressed format, which makes it wrong here.
- A compressed G1 point has 32 bytes;
- An uncompressed G1 point has 64 bytes;
- An encoded G1 point in EVM has 128 bytes.

If you want an uncompressed G1 point in a 64-byte array, then please use `pk.pg1.Marshal()` or `pk.pg1.RawBytes()`

crypto/tpke/public_key.go
txhsl: > I think https://pkg.go.dev/github.com/consensys/gnark-crypto/ecc/bls12-381#G1Affine.SetBytes handles uncompressed too, btw.

But `SetBytes` checks `len(buf) >= SizeOfG1AffineCompressed` instead of `len(buf) == SizeOfG1AffineUncompressed`, so compressed G1 point is included as well, however it should not be `decodePointG1`.

crypto/tpke/public_key.go
AnnaShaleva: Well, it's just a bad naming. I want a nice-name wrapper for `decode/encodePointG1` unified with existing `Bytes()`, because it's easy to misuse `Bytes()` and `Encode()`. And `Uncompressed*` is just a wrong word. May be you have some suggestions? Maybe something like  `EVMBytes` (although it sounds strange a bit).

Although it's not that important, we may keep it as it is.

crypto/tpke/public_key.go
AnnaShaleva: Fixed.

consensus/dbft/dbft-TODO.md
AnnaShaleva: Some of my local changes are leaking, fixed.

crypto/tpke/public_key.go
AnnaShaleva: Shili is right, and it's not actually an uncompressed form, the naming is misleading.

roman-khimov (dismissed):

AnnaShaleva:
@roman-khimov it's not ready for the merge yet, we need to decide about naming issue. Let's wait for @txhsl opinion on https://github.com/bane-labs/go-ethereum/pull/358#discussion_r1871573506 and then I'll rename or revert the changes.

roman-khimov:
That's exactly why I left it open for now.

crypto/tpke/public_key.go
txhsl: I would say `Encode()` and `Decode()` is the current Geth style naming in `v1.13`.

https://github.com/bane-labs/go-ethereum/blob/e91c2e32f1c16facdb9c533c77734036f6af5ebf/crypto/bls12381/g1.go#L136-L159

AnnaShaleva (commented):
Let's have it this way.

crypto/tpke/public_key.go
AnnaShaleva: Good, then let's keep `Encode`, check it one more time please.

roman-khimov (approved):

crypto/tpke/public_key.go
```diff
@@ -5,16 +5,32 @@ import (
...
 
+// PublicKey is a BLS12-381 based public key implementation used in TPKE scheme.
 type PublicKey struct {
 	pg1 *bls12381.G1Affine // A public value for tpke encryption
 }
 
-// NewGlobalPublicKey aggregates and returns a PublicKey
+// keycache is a simple lru cache for bls12381 keys that avoids Y calculation
+// overhead for known keys serialized in compressed form.
+var keycache *lru.Cache[string, *PublicKey]
```
txhsl, line 26: I'd expect this to be defined in dBFT, although it works here. 🤔

AnnaShaleva: This cache is global, it's done intentionally to cover possible other usages of compressed keys deserialisation.

AnnaShaleva: We use  the same approach for NeoGo, see for example https://github.com/nspcc-dev/neo-go/blob/2bc41dbe30eea58616380ed2c7e0e14a054cf59c/pkg/crypto/keys/publickey.go#L122

txhsl (approved):

PR #369, opened 2024-12-18
config: reduce dBFT block time to 5s

### LINKED ISSUE #368: Reduce block time to 5 seconds
# Rationale

In fact, we've already applied this change to T4 TestNet, and @steven1227 suggests now we can copy this change to MainNet as well.

The only difference I think this change will make is reducing the epoch time duration from about 1 week to about half a week. Since we will not have any Governance lock period for DKG and users will start to receive rewards ASA they vote after #332, the UX influence is also quite slight.

The biggest concern we may have after this is #295, which means the time we reserve for ZK-DKG is reduced as well.

# Implementation

Update the genesis files and reinit our databases.

txhsl:
Close #368. Seems no doc needs to get updated.

### DIFF
--- a/config/genesis_mainnet.json
+++ b/config/genesis_mainnet.json
@@ -16,7 +16,7 @@
     "grayGlacierBlock": 0,
     "shanghaiTime": 0,
     "dbft": {
-      "period": 10,
+      "period": 5,
       "standbyValidators": [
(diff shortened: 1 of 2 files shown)


AnnaShaleva (approved):
Confirmed, no additional changes are needed.

PR #375, opened 2024-12-26
workflow: add https settings for accessing private packages under bane-labs

txhsl:
Ref #363 and #373.

This is an HTTPS based scheme to allow `go get` in GitHub Action to access any possible private packages under organization `bane-labs`.
- A non-code GitHub App [`Bane Cross Repo Access`](https://github.com/organizations/bane-labs/settings/apps/bane-cross-repo-access) is created by `bane-labs`;
- This App has been installed, and can only be installed by `bane-labs`;
- This App uses its private key to generate tokens, which permission is `Read access to code, metadata, and packages`;
- The workflow set this token to `./netrc`, so that cmds `make all` and `go get` can work with private repositories.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -60,8 +60,26 @@ jobs:
           go-version: '1.22'
           cache: true
 
+      - name: Create GitHub App Token
+        uses: actions/create-github-app-token@v1.11.1
+        id: app-token
+        with:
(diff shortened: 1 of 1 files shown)


gsmachado (approved):
I think this approach is way cleaner and better than before. It maintains a higher level of security.

Approved.

Issue #376, opened 2024-12-26
BLS12381 scheme lacks subgroup check in v1.13.15


txhsl:
# Rationale

The current Geth version `v1.13.15` we are using doesn't perform subgroup check, only expect before `bls12381Pairing`. Ethereum introduced this fix through https://github.com/ethereum/EIPs/pull/8456, and finally implemented in `v1.14.2` through https://github.com/ethereum/go-ethereum/pull/29637.

# Implementation

Since we are using `Consensys/gnark-crypto` for most of the encoding/decoding, the BLS signature and Envelope transaction are well-checked, so that no need to update.

But the EVM format related decoding only checks `IsOnCurve()`. This happens in `decodePointG1()` and `decodePointG2()` of both EVM precompiles and DKG.

I'm going to add BLS12381 subgroup check to our `crypto/tpke`, but leave the code in `core/vm/contracts.go` for `v1.14.2` upgrade.

dusmart:
Thank you for the detailed description of the issue! Please keep this issue open until we merge v1.14.2, after which we can close it. I truly appreciate the effort you put into outlining the details.

txhsl:
Fixed in #495.

PR #377, opened 2024-12-26
crypto/tpke: apply subgroup check on any G1/G2 data from EVM

txhsl:
Part of #376.


dusmart:
LGTM

This is excellent work! While I wasn’t ultimately able to identify an PoC that demonstrates a feasible attack, implementing this preventative measure in a timely manner will be highly effective to further attacks. Thank you for addressing this issue so thoroughly and preventing it from hack!

PR #390, opened 2025-01-15
dbft: disable decryption shares construction if enforceECDSASignatures

### LINKED ISSUE #382: Prevent encrypted transactions decryption with `--antimev.force-ecdsa-signatures` enabled
https://github.com/bane-labs/go-ethereum/pull/366 introduces fallback to multisignature block signing scheme. However, if something wrong happens with DKG setup, nodes with `--antimev.force-ecdsa-signatures` flag will still try to construct `PreCommit` with potentially missing or invalid shares.

We need to disable shares construction and decoding if `--antimev.force-ecdsa-signatures` is enabled.

Thanks to @txhsl for the report.

AnnaShaleva:
If enforceECDSASignatures is on, then:

1. Don't decode Envelopes.
2. Disable construction of decryption shares at PreCommit constructor level.
3. Verify PreCommits of other CNs against the actual list of actual decoded Envelops count.

Close #382.

Tested on 4-nodes privnet, works as expected (Envelope is accepted instead of decrypted inner tx). @songb2, @txhsl  please ensure/test that it also works for your case.

### DIFF
--- a/consensus/dbft/preblock.go
+++ b/consensus/dbft/preblock.go
@@ -22,8 +22,13 @@ type PreBlock struct {
 	localShares  []byte
 
 	// envelopesData is the ordered list of decoded content of Envelopes
-	// transactions of the proposed PreBlock.
+	// transactions of the proposed PreBlock. In case if enforceECDSASignatures
+	// is enabled, an attempt to decode Envelope transactions won't be taken, hence
+	// envelopesData will be empty.
 	envelopesData []envelopeData
+	// enforceECDSASignatures reflects whether dBFT uses backup multisignature
+	// block signing scheme.
+	enforceECDSASignatures bool
 
 	// finalTransactions is the cached final list of transactions formed after TPKE
 	// decryption of Envelopes content. This list includes both simple standard and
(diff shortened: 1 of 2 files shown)


consensus/dbft/preblock.go
```diff
@@ -41,13 +46,17 @@ func (p *PreBlock) Data() []byte { return p.localShares }
...
 func (p *PreBlock) SetData(pk dbft.PrivateKey) error {
-	encryptedKeys := make([]*tpke.CipherText, len(p.envelopesData))
-	for i := range p.envelopesData {
-		encryptedKeys[i] = p.envelopesData[i].encryptedKey
-	}
-	shares, err := pk.(*Signer).AmevKeystore.DecryptWithShare(encryptedKeys)
-	if err != nil {
-		return fmt.Errorf("failed to construct shares: %w", err)
+	var shares []*tpke.DecryptionShare
+	if !p.enforceECDSASignatures {
```
roman-khimov, line 50: Can't you detect this via `AmevKeystore` or `pk` itself? I don't exactly like this field in `PreBlock`.

AnnaShaleva: > I don't exactly like this field in PreBlock.

Why?

> Can't you detect this via AmevKeystore or pk itself?

It's not the property of antimev keystore or private key; it's the node's setting. And this setting is on the `PreBlock` level because we need to use it not only for PreBlock-level shares generation, but also for in-block Envelopes decoding.

roman-khimov (approved):

consensus/dbft/preblock.go
roman-khimov: I know it's a node setting, but the resulting structure resembles `StateRootInHeader` too much, this modality is not obvious. On the other hand relying on keystore (it can be like `key == nil` or smth, we're forcing non-TPKE anyway) can be more natural. But it's a beast already, so can handle one more thing.

The structure itself can be unexported though, no need for it to be public.

PR #396, opened 2025-01-17
dbft: support cross-epoch Envelope decryption

AnnaShaleva:
Close #362, close #151.

@txhsl, @songb2 you need to update your JS library that generates Envelope transactions because we've changed the format of Envelope's data to include 4-bytes prefix (DKG epoch of transaction encryption, uint32 in LE form).

PR #400, opened 2025-01-23
contracts: make GovReward contract accept special envelope transfers

### LINKED ISSUE #399: Add a fallback handling for envelope transaction in system contract
# Rationale

Our envelope transactions are supposed to have address `0x1212000000000000000000000000000000000003` as its `to` address, and special message data begins with `0xffffffff`.

Meanwhile, the current governance reward contract only have a `receive()` method to handle pure GAS transfer, so decrypt-failed envelopes will now get rejected by contract.

# Implementation

We need a `fallback()` method to allow GAS transfer with specific messages.

txhsl:
Close #399.


contracts/solidity/GovReward.sol
```diff
@@ -13,6 +13,12 @@ contract GovReward is IGovReward, GovProxyUpgradeable {
 
     receive() external payable {}
 
+    fallback() external payable {
+        if (msg.sig != bytes4(0xffffffff)) {
```
AnnaShaleva, line 17: May be I'm missing something, but could you please clarify, why do you check `msg.sig` instead of `msg.data`? Because encrypted transaction is a part of Envelope's data.

AnnaShaleva: @songb2, test it please, ensure that Envelope can be accepted and successfully executed (you may intentionally construct invalid encrypted transaction to enforce Envelope acceptance).

txhsl: > May be I'm missing something, but could you please clarify, why do you check `msg.sig` instead of `msg.data`? Because encrypted transaction is a part of Envelope's data.

Maybe ref https://docs.soliditylang.org/en/latest/units-and-global-variables.html, `msg.sig` is the method selector.

> - msg.data (bytes calldata): complete calldata
> - msg.sig (bytes4): first four bytes of the calldata (i.e. function identifier)

txhsl: This works, but we might have some issue with #396.

txhsl: #406

PR #401, opened 2025-01-23
antimev: import ZK to improve DKG security

### LINKED ISSUE #295: Add ZKP to DKG contract

txhsl:
Close #295. Only a rough implementation, need long term testing and verification.

This is scheduled for `v0.4.0`, ~~but if possible, we can have it in `v0.3.0`~~.


txhsl:
Will reopen when this is ready and basically tested.

PR #402, opened 2025-01-25
DKG: fix initialization process

AnnaShaleva:
Otherwise it doesn't work correctly and may lead to deadlocks during the node's start.

@txhsl, @chenquanyu, @songb2 please review&test.


roman-khimov (approved):

txhsl (approved):
I'm good with this.

PR #407, opened 2025-02-06
antimev: fix preblock dkg index and update generation test

txhsl:
Close #406. @songb2 Please test the features introduced by #396.

PR #412, opened 2025-02-10
consensus/dbft: fix cross epoch decryption start round

txhsl:
Relate to #407.

Since the `dkgRound` is now `c.dkgSnapshot.Round - 1`, the first reshare round is finished when `c.dkgSnapshot.Round` becomes `3`, so `crossEpochDecryptionStartRound` should be `2`.

### DIFF
--- a/consensus/dbft/dbft.go
+++ b/consensus/dbft/dbft.go
@@ -93,8 +93,8 @@ const (
 	// crossEpochDecryptionStartRound is the number of DKG round (as denoted in KeyManagement
 	// system contract) starting from which continuous cross-epoch Envelopes decryption is supported.
 	// First DKG round setups sharing key, second DKG round setups resharing key, hence resharing
-	// key may be used for decryption starting from the third round.
-	crossEpochDecryptionStartRound = 3
+	// key may be used for decryption starting from the second round.
+	crossEpochDecryptionStartRound = 2
 )
 
 // Various error messages to mark blocks invalid. These should be private to


consensus/dbft/dbft.go
```diff
@@ -93,8 +93,8 @@ const (
 	// crossEpochDecryptionStartRound is the number of DKG round (as denoted in KeyManagement
 	// system contract) starting from which continuous cross-epoch Envelopes decryption is supported.
 	// First DKG round setups sharing key, second DKG round setups resharing key, hence resharing
-	// key may be used for decryption starting from the third round.
-	crossEpochDecryptionStartRound = 3
+	// key may be used for decryption starting from the second round.
```
AnnaShaleva, line 96: It's still may be used from the *third* round, it's just we use zero-based indexes starting from https://github.com/bane-labs/go-ethereum/pull/407.

txhsl: It is zero-based, yes, but in contract, `0` is an empty round. And the number of our first DKG round is `1`. So the round `2` is the second DKG round we have.

In the following code, the Governance contract checks if the next epoch points to a new DKG. For the first check, it's from `0` to `1`.

https://github.com/bane-labs/go-ethereum/blob/41c9a7ad80d32c2ff2919cb0cad01650e26c1ea8/contracts/solidity/Governance.sol#L322-L329

If it's from `0` to `0`, then the first round fails and the next epoch points to an empty DKG.

Issue #342, opened 2024-11-01
Build decrypted transaction verification rules


AnnaShaleva:
# Rationale

The problem described under https://github.com/bane-labs/go-ethereum/pull/339#discussion_r1823325166.

# Implementation

Ensure that decrypted transactions don't have a way to "cheat" with the fees. Adjust corresponding verification logic implemented as `(c *DBFT) validateDecryptedTx`.

vncoelho:
I did not see the comment, where is it?

AnnaShaleva:
> I did not see the comment, where is it?

It's here: https://github.com/bane-labs/go-ethereum/pull/339#discussion_r1823325166.

txhsl:
Still, my opinion is https://github.com/bane-labs/go-ethereum/pull/339#discussion_r1825541900.

txhsl:
@steven1227 @AnnaShaleva @roman-khimov @gsmachado @chenquanyu @doubiliu @igormcoelho 

Maybe we are going to implement this quite soon, so let me describe a bit about the current solutions we have. It's welcome if there is any other better idea.

## Performance Cost

So first we need to know about the cost of an envelope transaction. The encryption is executed by the user, so we should only take the decryption into consideration.

Each envelope decryption includes a BLS12381 decryption and an AES256 decryption. The BLS12381 part costs the most but stable, meanwhile the AES256 is quick but related to the size of ciphertext. Since the AES computation is always fast, we can regard the cost of handling an envelope transaction is a **const** value.

## Solution 1: Charge a fixed value

So, based on the above evaluation, we firstly have the idea to require users to send a fixed value in his envelope transaction (not the inner transaction) to our `GovernanceReward` contract.

However, in our protocol, if the decryption succeeds, the envelope will be replaced but its inner transaction, so this value cannot be successfully charged in dBFT. To reflect this fee in the inner transaction, a convert becomes necessary.

As we all know, the fee of a transaction is calculated based on `fee=price*gas`. So if we want to add this value into the fee of
the inner transaction, we have to compute `price=fee/gas`. But the `gas` is the maximum gas limit that declared by sender, so it is possible that a transaction passes our check, but the final charged fee is much lower than we expect.

The block gas limit is also a point we need to take care. By declaring a very high gas limit, a sender may prevent other transactions from getting packed, or break the rule of block verification. But we don't have a proper method to detect this kind of attack, or handle such a transaction after decrypt it (because the inner transaction is encrypted).

[...]

Issue #313, opened 2024-09-04
Add Envelopes verification to mempool level


AnnaShaleva:
# Rationale

Envelope transactions currently are not verified in any way at the mempool level. They treated as simple transactions whereas we need a special verification rules for them which include at least additional fee check. We also need to think about others additional verification rules that may be applied to Envelopes.

# Implementation

We agreed not to use blob transactions as Envelopes, thus we need to extend only legacy pool. Definition of Envelope should be reused from `dbft` package (#287), it may be moved from `dbft` to some other package to avoid cyclic deps problem.

txhsl:
Depends on #342.

PR #413, opened 2025-02-11
antimev: add envelope fee policy and check

chenquanyu:
Part of #342 and #313.
We need to charge for encrypted transactions, so we add the envelopeFee setting in the policy contract. The gas price of envelope transactions needs to be increased by envelopeFee based on the original policy minimum value before they can enter the transaction pool. Because the decrypted gas price has been checked to be higher than the envelope transaction after the current decryption transaction, there is no need to add additional checks on the decrypted transaction.


txhsl:
The decrypted transaction will replace its envelope in dBFT. I see the price check in `validateDecryptedTx`, but I think we need to make sure this replacement will not break the rule of building a block.

E.g. the decrypted transaction has a much larger gas limit or consumes more gas than the envelope, and finally break the block gas limit. Is this possible in our code?

chenquanyu:
> E.g. the decrypted transaction has a much larger gas limit or consumes more gas than the envelope, and finally break the block gas limit. Is this possible in our code?

Yes, it's possible, we may add check before [`Process state with constructed list of transactions`](https://github.com/bane-labs/go-ethereum/blob/03095a21e124a5ef5cba25bff8c0c124794839ff/consensus/dbft/dbft.go#L1153).

txhsl:
The original path of constructing seems complex.
https://github.com/bane-labs/go-ethereum/blob/1635e6ea424bdb46066ba63b9abd5db87758ed42/miner/worker.go#L1100-L1129

txhsl:
> Yes, it's possible, we may add check before [`Process state with constructed list of transactions`](https://github.com/bane-labs/go-ethereum/blob/03095a21e124a5ef5cba25bff8c0c124794839ff/consensus/dbft/dbft.go#L1153).

How about #417.

BTW, in this price model, even there are only simplest transfers (21000 gas each) in the block, our TPKE can still handle the decryption and aggregation in less than 0.5s in the best case.

consensus/dbft/dbft.go
```diff
@@ -1232,6 +1232,12 @@ func (c *DBFT) validateDecryptedTx(head *types.Header, decryptedTx *types.Transa
 	if envelopeFrom != decryptedFrom {
 		return fmt.Errorf("decryptedTx from mismatch: decryptedFrom %v, envelopeFrom %v", decryptedFrom, envelopeFrom)
 	}
+	if decryptedTx.Type() == types.BlobTxType {
+		return fmt.Errorf("%w: decryptedTx is a blob transaction", types.ErrInvalidTxType)
+	}
+	if decryptedTx.Gas() > envelope.Gas() {
+		return fmt.Errorf("decryptedTx required gas more than envelope")
```
AnnaShaleva: > decryptedTx required gas more than envelope

decryptedTx requires more gas than envelope

consensus/dbft/dbft.go
```diff
@@ -1232,6 +1232,12 @@ func (c *DBFT) validateDecryptedTx(head *types.Header, decryptedTx *types.Transa
 	if envelopeFrom != decryptedFrom {
 		return fmt.Errorf("decryptedTx from mismatch: decryptedFrom %v, envelopeFrom %v", decryptedFrom, envelopeFrom)
 	}
+	if decryptedTx.Type() == types.BlobTxType {
+		return fmt.Errorf("%w: decryptedTx is a blob transaction", types.ErrInvalidTxType)
+	}
```
AnnaShaleva: This check seems to be redundant, because we verify decrypted transactions using local memory pool:
https://github.com/bane-labs/go-ethereum/blob/4dc35d385d4d2ad18f7842c87195f70c45adb3e1/consensus/dbft/dbft.go#L1133

Local mempool is legacy pool, it doesn't accept blob transactions:
https://github.com/bane-labs/go-ethereum/blob/4dc35d385d4d2ad18f7842c87195f70c45adb3e1/core/txpool/legacypool/legacypool.go#L707-L713

core/state_transition.go
```diff
@@ -328,7 +328,8 @@ func (st *StateTransition) preCheck() error {
 			// For LegacyTx, GasFeeCap and GasPrice are equal, so checking GasTipCap and GasFeeCap is enough.
 			if st.evm.ChainConfig().DBFT != nil {
 				var minGasTipCap = st.state.GetState(systemcontracts.PolicyProxyHash, systemcontracts.GetMinGasTipCapStateHash()).Big()
-				if cmath.BigMin(msg.GasTipCap, new(big.Int).Sub(msg.GasFeeCap, st.evm.Context.BaseFee)).Cmp(minGasTipCap) < 0 {
+				effectiveGasTip := cmath.BigMin(msg.GasTipCap, new(big.Int).Sub(msg.GasFeeCap, st.evm.Context.BaseFee))
+				if effectiveGasTip.Cmp(minGasTipCap) < 0 {
```
AnnaShaleva: Nothing is changed compared to the previous version. What we need here is to get envelope fee from Policy contract and ensure that it's included into transaction's fee.

core/txpool/validation.go
```diff
@@ -215,10 +216,19 @@ func ValidateTransactionWithState(tx *types.Transaction, signer types.Signer, op
...
 			ErrUnderpriced, minGasTipCap, baseFee, tx.GasTipCap(), tx.GasFeeCap())
 	}
+	// Apply policy envelope fee check
+	if antimev.IsEnvelope(tx) {
+		var envelopeFee = opts.State.GetState(systemcontracts.PolicyProxyHash, systemcontracts.GetEnvelopeFeeStateHash()).Big()
+		if effectiveGasTip.Cmp(new(big.Int).Add(minGasTipCap, envelopeFee)) < 0 {
+			return fmt.Errorf("%w: policy envelopeFee needed %v, minGasTipCap needed %v, baseFee needed %v, gasTipCap %v, gasFeeCap %v ",
+				ErrUnderpriced, envelopeFee, minGasTipCap, baseFee, tx.GasTipCap(), tx.GasFeeCap())
+		}
+	}
```
AnnaShaleva: In fact, here we treat envelope fee as a part of gas tip. So I'd suggest:
```suggestion
	// Apply policy envelope fee check
	if antimev.IsEnvelope(tx) {
		var envelopeFee = opts.State.GetState(systemcontracts.PolicyProxyHash, systemcontracts.GetEnvelopeFeeStateHash()).Big()
		minGasTipCap.Add(minGasTipCap, envelopeFee)
	}
	if math.BigMin(tx.GasTipCap(), new(big.Int).Sub(tx.GasFeeCap(), baseFee)).Cmp(minGasTipCap) < 0 {
		return fmt.Errorf("%w: policy minGasTipCap (including Envelope fee for Envelopes) needed %v, baseFee needed %v, gasTipCap %v, gasFeeCap %v ",
			ErrUnderpriced, minGasTipCap, baseFee, tx.GasTipCap(), tx.GasFeeCap())
	}
```

core/state_transition.go
AnnaShaleva: I see that you've added this check to the pool, but we also need the same check to be present at the state transition level.

AnnaShaleva: One thing that is missing from the current PR is a convenient way for the external user to retrieve Envelope fee from Policy contract. For example, we have RPC handlers that suggest gas tip capacity and GAS price, but don't have any RPC that allows to easily get current envelope fee. Do you think we need to add one? It may be done via reading the contract state on the user's side, but IMO, a separate RPC handler is better since it might be a hot path.

How does JS library (that helps user to generate Envelope transaction) handles envelope fee? It has to be included into final tx fee.

contracts/solidity/Policy.sol
```diff
@@ -120,7 +121,8 @@ contract Policy is IPolicy, GovernanceVote, GovProxyUpgradeable {
             keccak256(abi.encode(_candidateLimit))
         )
     {
-        if (_candidateLimit < IGovernance(GOV).consensusSize()) revert Errors.InvalidCandidateLimit();
+        if (_candidateLimit < IGovernance(GOV).consensusSize())
+            revert Errors.InvalidCandidateLimit();
```
AnnaShaleva: Not related to the current PR, revert it.

AnnaShaleva (commented):
> E.g. the decrypted transaction has a much larger gas limit or consumes more gas than the envelope, and finally break the block gas limit. Is this possible in our code?

No, it's not possible even without this PR, because in the described case there will be an error on transactions processing during final Block construction in dBFT which will result in a fallback to the original set of transactions for the whole block. And we always know that the original set of transactions (included non-decrypted Envelopes) is always valid since it's verified at the PrepareRequest level. @txhsl @chenquanyu please check the following code and ensure it's not possible, otherwise please submit a bug:
https://github.com/bane-labs/go-ethereum/blob/4dc35d385d4d2ad18f7842c87195f70c45adb3e1/consensus/dbft/dbft.go#L1080-L1089

> How about https://github.com/bane-labs/go-ethereum/pull/417.

Regarding 417: it works, but it imposes too harsh restrictions on the inner encrypted transaction, I see the following pitfall with this approach: consider that user wants to encrypt some transaction that requires more gas than a simple Envelope transfer. Then the user inevitably has to increase gas for outer Envelope to be able to pay for the decrypted transaction execution. But if decryption is failing for some reason, then Envelope will be accepted spending all gas for nothing.

> but it imposes too harsh restrictions on the inner encrypted transaction

May be I'm missing something, but in general we probably need to discuss this issue more carefully. Because in general I'd expect from Envelope to include gas only for transfer + encryption fee (+ optional prioritization fee). @roman-khimov @txhsl, need your thoughts on this.

consensus/dbft/dbft.go
txhsl: Yes, then it's redundant.

consensus/dbft/dbft.go
txhsl: > No, it's not possible even without this PR, because in the described case there will be an error on transactions processing during final Block construction in dBFT which will result in a fallback to the original set of transactions for the whole block.

Yes, but.

> Regarding 417: it works, but it imposes too harsh restrictions on the inner encrypted transaction, I see the following pitfall with this approach: consider that user wants to encrypt some transaction that requires more gas than a simple Envelope transfer. Then the user inevitably has to increase gas for outer Envelope to be able to pay for the decrypted transaction execution.

Then if several such transactions break the block gas limit, then we fall back the whole block? I know the gas limit check here is restricted, but we'd better not allow the transaction execution results to be affected by other transactions.

For our user, although it requires some more calculation, but they can declare a higher `gaslimit` for their envelopes. It's only for allocating the block space, left gas will pay back.

core/state_transition.go
txhsl: > How does JS library (that helps user to generate Envelope transaction) handles envelope fee? It has to be included into final tx fee.

A separate RPC handler will be more convenient, yes.

consensus/dbft/dbft.go
AnnaShaleva: > Then if several such transactions break the block gas limit, then we fall back the whole block?

Yes. But please note, I’m not saying that this PR is useless; we definitely need this PR. I’m just saying that in emergency situation dBFT won’t accept transactions that violate block gas limit constraint.

> but we'd better not allow the transaction execution results to be affected by other transactions.

Agree.

> but they can declare a higher gaslimit for their envelopes.

Then users have to know about this rule. Does JS library offers correct suggested gas price (with included gas of decrypted tx) for the Envelope construction?

> It's only for allocating the block space, left gas will pay back.

Could you please clarify, how? If the user is able to claim the gas back, then decryption fee should be taken from it anyway (irrespectively if it’s an Envelope or inner transaction), I’m not sure that I saw this in the code.

txhsl: > I’m just saying that in emergency situation dBFT won’t accept transactions that violate block gas limit constraint.

OK.

> Then users have to know about this rule. Does JS library offers correct suggested gas price (with included gas of decrypted tx) for the Envelope construction?

It's possible, said by the developer, and both the gas price and gas limit are customizable for Metamask. I checked that code, it's now a demo rather than a library.

> Could you please clarify, how? If the user is able to claim the gas back, then decryption fee should be taken from it anyway (irrespectively if it’s an Envelope or inner transaction), I’m not sure that I saw this in the code.

First, we know that the `tx.Gas()` returns the gas limit of a transaction. It is a declared maximum value that the user is willing to pay for transaction execution.

https://github.com/bane-labs/go-ethereum/blob/c11a70a1027c8d6afb3b8cdacf27f1b442ea9587/core/types/transaction.go#L288-L289

If the gas limit is all used in EVM execution, then nothing will be returned. But if there is left gas after execution, they will be added back to the address balance.

> For example, if you put a gas limit of 50,000 for a simple ETH transfer, the EVM would consume 21,000, and you would get back the remaining 29,000.

So declaring a higher gas limit for envelope transaction will not increase the real cost.

The whole picture of this PR is to charge in price. The decryption fee is now included in gas tip, and this part will be finally sent to governance reward, somehow equivalent to sending the value to the contract directly. In this scheme, envelopes are charged based on data length (case fail), and decrypted transactions are charged based on execution complexity (case success). 

https://github.com/bane-labs/go-ethereum/blob/8e7472e4298654ae7836e5393abdfc1221d78947/miner/worker.go#L817

But I'm checking the miner worker, it seems filling the block based on exact cost instead of gas limit. So even if the gas limit check here passes, there is still a possibility to break the block limit. 🤔So how can we protect the block gas limit then?

core/state_transition.go
txhsl: Ref [8e7472e](https://github.com/bane-labs/go-ethereum/pull/413/commits/8e7472e4298654ae7836e5393abdfc1221d78947).

consensus/dbft/dbft.go
roman-khimov: > we'd better not allow the transaction execution results to be affected by other transactions.

:+1: One deliberately created transaction will affect a number of valid ones then, not acceptable.

> But if decryption is failing for some reason, then Envelope will be accepted spending all gas for nothing.

1. It's not expected for valid transactions.
2. It's not that big of a problem to me. At least it's much less of an issue than the one above. The rule is pretty simple to follow, can be explained to people.

AnnaShaleva: > it seems filling the block based on exact cost instead of gas limit. So even if the gas limit check here passes, there is still a possibility to break the block limit. 

I need to check Geth code more precisely. The straightforward solution would be to compare the exact execution cost instead of comparing gas limit, but we can't do it easily because it requires re-execution of the Envelope during PreBlock processing, it's too inefficient.

txhsl: > I need to check Geth code more precisely. The straightforward solution would be to compare the exact execution cost instead of comparing gas limit, but we can't do it easily because it requires re-execution of the Envelope during PreBlock processing, it's too inefficient.

Or maybe we change the miner logic to reserve enough space for envelopes.

AnnaShaleva: > to reserve enough space for envelopes.

@txhsl, it looks to me that the only proper way to do it is to reserve the whole `decryptedTx.Gas()`. To implement that we also need to store `decryptedTx.Gas()` value in an unencrypted part of Envelope's data (like we did for hash in https://github.com/bane-labs/go-ethereum/pull/418).

Reservation of less amount makes no sense to me because decrypted transaction is allowed to spend up to `decryptedTx.Gas()`, so if we reserve less then there's a possibility of block gas limit overflow. But this approach contains a possible vector of attack: consider that all encrypted transactions use very large value of `decryptedTx.Gas()`whereas in fact their execution consumes only a very tiny part of available gas. The rest will be returned back to the user's account, but the problem that these Envelopes will take a lot of block space and prevent other transactions from entering the block due to the block gas limit. So malicious Envelopes senders may misuse this mechanism to occupy the block's space, and it won't cost a lot. What do you think?

txhsl: > But this approach contains a possible vector of attack: consider that all encrypted transactions use very large value of `decryptedTx.Gas()` whereas in fact their execution consumes only a very tiny part of available gas.

Yes, it's possible. But IMO, this attack is a normal use case as well, some transaction may require a high gas limit but consume really low. We need to allow it, but the sender should pay for the price.

And you are right, we'd better keep the miner logic, so then we need to update GovReward contract and improve the price scheme, so that the Geth code can handle them correctly without potential risks.

AnnaShaleva: We discussed this issue with @roman-khimov, so the final solution is following:
1. Require Envelope to have the same gas limit as inner encrypted transaction.
2. Implement extension of GovernanceReward contract: if payment from Envelope is received, then burn the rest of available gas up to Envelope's gas limit.
3. Don't change the miner logic.

This scheme allows inner transaction to use as much gas as needed and do not restrict user with some artificial limits. If inner transaction is successfully decrypted and accepted, then after execution the remaining gas will be returned back to the user's account (so nothing is changed is this pat). But if decryption is failed and Envelope is accepted, then all available gas limit will be used by the system (partially for execution and partially just burnt by GovernanceReward contract). So the user won't have a chance to beat the system. Also this approach incentivise users to precisely calculate inner transaction's gas limit. And we suppose that our primary users are exchanges, so the difference between Envelope's and inner's execution gas is not that high assuming that both inner and Envelope are token transfers.

So to sum up, in "good" scenario when inner tx is decrypted and accepted, there will be no difference in behaviour for the user (comparing to simple transactions, except for the extra decryption fee). But in "bad" scenario when Envelope is accepted, it will be costly for the user, but it's a necessary trade-off to protect the system from the possible attack.

@txhsl do you agree with the proposed solution?

txhsl: Yes, and I've implemented this code. Ref #422.

txhsl: > But this approach contains a possible vector of attack: consider that all encrypted transactions use very large value of `decryptedTx.Gas()` whereas in fact their execution consumes only a very tiny part of available gas.

I have to say that this attack is still possible even after this pat. Are we OK with it, or just allow it for now?

AnnaShaleva: Hm, you're right, and I missed that. This solution only solves the problem if Envelope is accepted, but we need to prevent this kind of attack also for cases when inner transactions accepted, so we're not OK with it.

PR #414, opened 2025-02-11
Fix memory leak

### LINKED ISSUE #386: Possible memory leak with commits after v0.2.2
#### System information

OS & Version: Linux
Commit hash : [a0e8c9c](https://github.com/bane-labs/go-ethereum/commit/a0e8c9cab2f6ec88371d7bc6667bc075af766e54)

#### Expected behaviour

The node memory usage is quite stable and low with `v0.2.2`.

#### Actual behaviour

But when we are using the latest commit code to upgrade an existing network, @songb2 confirmed a significant memory usage pump.

bd613d507327b8b9f97f456a95ee5b8

The above graph tracks the memory usage of one of our servers, it runs 3 CNs as a part of our privnet.

We stopped the network and started to try to upgrade it from Jan 8th 15:00, and finally started it up in Jan 10th 18:00.

So regardless of the middle, we can see an obvious increase, and finally the server crashed.

Set up a privnet with `make privnet_start_seven`, and keep running it, maybe more than 1 day.

#### Backtrace

The first graph is from a node running on [a0e8c9c](https://github.com/bane-labs/go-ethereum/commit/a0e8c9cab2f6ec88371d7bc6667bc075af766e54) for half a week.
388523caf39f7ffb6fc3957743c8f52
The following two are from another node running on [93810b5](https://github.com/bane-labs/go-ethereum/commit/93810b58010f15da845e485ecd82e71105790727) for about half a month.
77941d8db08354796cadb4acfa0637b
When only select `/consensus` as scope, it shows that 75% memory usage came from here.
cbead5f61d21e3d190ed55f6364bcc7

roman-khimov:
Fixes #386. Or maybe not, this is just a hypothesis at this stage. @songb2, @txhsl, please check. Also includes a patch to fix Go 1.23 builds (cherry-picked from upstream).

### DIFF
--- a/cmd/geth/main.go
+++ b/cmd/geth/main.go
@@ -359,8 +359,6 @@ func geth(ctx *cli.Context) error {
 // it unlocks any requested accounts, and starts the RPC/IPC interfaces and the
 // miner.
 func startNode(ctx *cli.Context, stack *node.Node, backend ethapi.Backend, isConsole bool) {
-	debug.Memsize.Add("node", stack)
-
 	// Start up the node itself
 	utils.StartNode(ctx, stack, isConsole)
 
--- a/consensus/dbft/dbft.go
+++ b/consensus/dbft/dbft.go
@@ -858,6 +858,7 @@ func (c *DBFT) verifyPreBlockCb(b dbft.PreBlock[common.Hash]) bool {
 	ethBlock := dbftBlock.ToEthBlock()
 
 	localPool := c.newLocalPool(parent)
+	defer localPool.CloseSilently()
 	errs := localPool.Add(dbftBlock.transactions, false, false)
 	for i, err := range errs {
 		if err != nil {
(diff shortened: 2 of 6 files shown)


txhsl:
Let's have a try.

songb2:
I've run about 2 days, it's better now, but the memory is still increasing.
image
Let's wait more time to see if it can be stable at a level.

txhsl (approved):

AnnaShaleva (commented):
ACK, should be a valid fix.

songb2:
This is the 6 days memory consumption trend, we can see that it's still increasing slowly, let's wait more time to see if it will make the server down (because previous commits will make server down).
image
image

songb2:
It finally make server OOM after running 7 days or so.
image


image

Here is the geth mem dump:
[memdump219.txt](https://github.com/user-attachments/files/18857376/memdump219.txt)

PR #418, opened 2025-02-17
antimev: return declared transaction hash for envelope transactions

### LINKED ISSUE #411: Metamask doesn't trace the decrypted transaction but the envelope
#### System information

OS & Version: Linux
Commit hash : [28a4058](https://github.com/bane-labs/go-ethereum/commit/28a4058845e358d5f63a2fcd720ba4500b00fb6e)

#### Expected behaviour

We expect user wallets (e.g. Metamask) can work with our Envelope transactions.

#### Actual behaviour

As we tested with #316 and #287, it was found that Metamask could not trace decrypted transactions correctly.

The reason is, we are sending the outside Envelope to the mempool, but it gets replaced by the decrypted transaction in dBFT.

For Metamask, the transaction it sends is dropped. It doesn't know how and is not going to trace the decrypted transaction any further.

It doesn't destroy the user experience, but somehow an issue.

#### Backtrace

Image

txhsl:
Close #411. This changes the Geth behavior of transaction submission.


internal/ethapi/api.go
```diff
@@ -1889,7 +1890,14 @@ func SubmitTransaction(ctx context.Context, b Backend, tx *types.Transaction) (c
 	} else {
 		log.Info("Submitted transaction", "hash", tx.Hash().Hex(), "from", from, "nonce", tx.Nonce(), "recipient", tx.To(), "value", tx.Value())
 	}
-	return tx.Hash(), nil
+
+	// If the transaction is an antimev envelope, then return the declared transaction hash instead of its own.
```
vncoelho, line 1894: Perhaps now it is going to be as discussed on the issue,right?
The original hash will be pushed instead.

txhsl: > Perhaps now it is going to be as discussed on the issue,right? The original hash will be pushed instead.

Yes, will test soon.

It works.

internal/ethapi/api.go
```diff
@@ -1889,7 +1890,14 @@ func SubmitTransaction(ctx context.Context, b Backend, tx *types.Transaction) (c
...
+	if antimev.IsEnvelope(tx) {
+		hashOffSet := antimev.EncryptedDataPrefixLen + antimev.EncryptedDataRoundLen
+		return common.Hash(tx.Data()[hashOffSet : hashOffSet+antimev.EncryptedDataHashLen]), nil
+	} else {
+		return tx.Hash(), nil
+	}
```
AnnaShaleva, line 1900: This change solves the original problem, however consider the situation when decryption fails for some reason and the original Envelope is accepted instead of inner transaction. In this case, it seems that we end up in the reverted situation: wallet still tracks the inner transaction (by its hash) whereas original Envelope is accepted. What is the Metamask behaviour in this case?

txhsl: Before sending a transaction, Metamask will try to get address nonce from RPC `eth_getTransactionCount`, compare it with the local value, and choose the higher one.

But as for other wallet, I'm not sure.

vncoelho: It is a good point, Anna.

Decryption can fail but then we just set that original secret hash as failed sent by that users.

Internally you publish the envelope hash as an additional field for a sack of curiosity, this has no practical use.

AnnaShaleva (approved):
Discussed on the meeting. It's an invasive change and I'm not fond of it (likely @roman-khimov feels the same way), but I can't suggest any better alternative to make it work properly with metamask. In defense of this solution I can say that it doesn't change the node's behaviour for simple (non-amev) transactions.

vncoelho (approved):

roman-khimov (approved):
Compatibility has its cost, we can't ignore the #411 problem.

PR #420, opened 2025-02-18
dbft: dispose intermediate state DB copies

### LINKED ISSUE #386: Possible memory leak with commits after v0.2.2

AnnaShaleva:
Properly finalize prefetcher routines, otherwise they consume memory and lead to OOM.

Close #386.

WIP: it's just an idea and I need to check how some parts of StateDB work, but in general I'd expect this patch at least partially solve #386. @songb2, @txhsl could you please run this patch for a while using the same environment as in #386 and #414 and check the resulting memory consumption?

### DIFF
--- a/consensus/dbft/dbft.go
+++ b/consensus/dbft/dbft.go
@@ -462,7 +462,9 @@ func (c *DBFT) processBlockCb(b dbft.Block[common.Hash]) error {
 		return fmt.Errorf("failed to construct block witness: %w", err)
 	}
 	dbftBlock.header.Extra = append(dbftBlock.header.Extra, witness...) // Extra version isn't changed, validators addresses and signatures are added.
+	// TODO: here snapshot usage is impossible since PutBlock modifies&finalize state.
 	state := dbftBlock.state.Copy()
+	defer state.Dispose()
 	res := types.NewBlockWithHeader(dbftBlock.header).WithBody(dbftBlock.transactions, nil).WithWithdrawals(dbftBlock.withdrawals)
 
 	// Firstly, notify chain about new block.
@@ -2730,7 +2732,9 @@ func (c *DBFT) getGlobalPublicKey(h *types.Header, s *state.StateDB) (*tpke.Publ
 	keystore := c.amevKeystore.Copy()
 	c.lock.RUnlock()
 
-	err := c.handleDKG(snapshot, keystore, h, s.Copy(), true)
+	cp := s.Copy()
+	defer cp.Dispose()
+	err := c.handleDKG(snapshot, keystore, h, cp, true)
 	if err != nil {
 		return nil, fmt.Errorf("failed to handle DKG at %d: %w", h.Number.Uint64(), err)
 	}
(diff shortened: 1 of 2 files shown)


songb2:
Have run 2 days, here is the data trend, the memory consumption is creasing. Let's continue to wait for more days.
image
image

songb2:
5 days or so
image


image

AnnaShaleva:
Yes, thank you for testing, it's still ~2GB/day, so I'm continuing investigation.

AnnaShaleva:
@songb2 can you share the memory profile as far?

songb2:
@AnnaShaleva Below is the latest mem dump :

[memdump2.25.txt](https://github.com/user-attachments/files/18953234/memdump2.25.txt)
image
image

AnnaShaleva:
@songb2 I mean, Golang's pprof flame graph, I'd like to see if traces are different comparing to https://github.com/bane-labs/go-ethereum/issues/386#issuecomment-2664503184.

txhsl:
> @songb2 I mean, Golang's pprof flame graph, I'd like to see if traces are different comparing to [#386 (comment)](https://github.com/bane-labs/go-ethereum/issues/386#issuecomment-2664503184).

image

songb2:
The server has memory alerts now, here is the 7 day's mem consumption data and trend:
[memdump2.26.txt](https://github.com/user-attachments/files/18975932/memdump2.26.txt)

image
image
image

songb2:
@AnnaShaleva 
8 day's trend 
[memdump2-27.txt](https://github.com/user-attachments/files/19000695/memdump2-27.txt)
image
image
image

txhsl:
@songb2 Please have a try with https://github.com/bane-labs/go-ethereum/pull/420/commits/c75652019d4811c74c71cc27cf66b26d08270157, and maybe we decrease the block time to speed up verification.

AnnaShaleva:
>Please have a try with https://github.com/bane-labs/go-ethereum/commit/c75652019d4811c74c71cc27cf66b26d08270157

It won't work, snapshot reverting mechanism fails with NPE, I'm still digging into the source of problem, so I'd suggest to release without this PR.

AnnaShaleva:
This approach doesn't work, will investigate other solutions.

PR #431, opened 2025-02-28
contracts: move encrypted tx gas limit check to consensus level

AnnaShaleva:
This commit solves two problems:
1. By the moment of GovReward contract execution we must be sure that encrypted transaction gas limit satisfies minimum required value.
2. By the same moment we must be sure that Envelope transaction has enough gas limit to pay for execution of decrypted transaction.

If these two constrains are violated, contract itself can't do anything about it, the Envelope execution will be aborted, but still, Envelope will be accepted to the chain since we don't filter out failed Envelopes (and we shouldn't do that, ref. https://github.com/bane-labs/go-ethereum/pull/422#discussion_r1971545809).

However, the check is still needed because otherwise it's easy for the user to accidently send an invalid Envelope spending user's funds for nothing. I won't say that it's a vector of attack because small Envelope fee will lead to the fact that encrypted tx execution itself will be restricted by this value and hence it's not critical for the system and doesn't bring any profit to the user.

This commit moves these two checks to mempool and state transition level so that it's impossible to submit Envelope that is improperly constructed.

@txhsl need your review.

### DIFF
--- a/antimev/envelope.go
+++ b/antimev/envelope.go
@@ -9,7 +9,13 @@ import (
 	"github.com/ethereum/go-ethereum/core/systemcontracts"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/crypto/tpke"
+	"github.com/ethereum/go-ethereum/params"
 )
+
+// MinEncryptedGasLimit is the minimum required gas limit for encrypted transaction.
+// It is set to be equal to a simple transfer execution cost since it is assumed that
+// minimum valid encrypted transaction structure is a simple transfer.
+const MinEncryptedGasLimit = uint32(params.TxGas)
 
 var (
 	// EncryptedDataPrefix is the prefix of Envelope transaction's data. It's used to
@@ -75,7 +81,7 @@ func GetEncryptedHash(envelope *types.Transaction) common.Hash {
 
 // GetEncryptedGas returns the gas limit of inner encrypted transaction specified in an
 // unencrypted part of Envelope data. Passing non-Envelope as an argument is a no-op.
-func GetEncryptedGas(envelope *types.Transaction) uint32 {
+func GetEncryptedGas(envelopeData []byte) uint32 {
 	gasOffset := EncryptedDataPrefixLen + EncryptedDataRoundLen
-	return binary.BigEndian.Uint32(envelope.Data()[gasOffset : gasOffset+EncryptedDataGasLen])
+	return binary.BigEndian.Uint32(envelopeData[gasOffset : gasOffset+EncryptedDataGasLen])
 }
--- a/core/state_transition.go
+++ b/core/state_transition.go
@@ -335,6 +335,16 @@ func (st *StateTransition) preCheck() error {
 						return fmt.Errorf("%w: address %v, gasLimit %v, policy maxEnvelopeGasLimit %v, ", ErrGasLimitReached, msg.From.Hex(), msg.GasLimit,
 							envelopeGasLimit)
 					}
+					// Check that encrypted transaction gas limit satisfies the minimum required gas limit.
+					encryptedGasLimit := antimev.GetEncryptedGas(msg.Data)
+					if encryptedGasLimit < antimev.MinEncryptedGasLimit {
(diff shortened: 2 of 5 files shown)


contracts/solidity/GovReward.sol
```diff
@@ -20,11 +20,8 @@ contract GovReward is IGovReward, GovProxyUpgradeable {
         }
         // burn required gas amount internally
         uint32 requiredSpace = Bytes.decodeUint32(msg.data[8:12]);
-        if (requiredSpace < 21000) {
-            revert Errors.InvalidGasLimit();
```
txhsl: From the contract perspective, this check is still necessary, otherwise there is a potential risk that `requiredSpace < 21000`.

I know the new code in Geth can guarantee that the underflow doesn't happen, but they are two separated things. We'd better leave it as it is.

core/state_transition.go
```diff
@@ -335,6 +335,16 @@ func (st *StateTransition) preCheck() error {
 						return fmt.Errorf("%w: address %v, gasLimit %v, policy maxEnvelopeGasLimit %v, ", ErrGasLimitReached, msg.From.Hex(), msg.GasLimit,
 							envelopeGasLimit)
 					}
+					// Check that encrypted transaction gas limit satisfies the policy.
```
txhsl: But it's not a Neo X policy, maybe we change the description.

antimev/envelope.go
```diff
@@ -11,6 +11,11 @@ import (
 	"github.com/ethereum/go-ethereum/crypto/tpke"
 )
 
+// MinEncryptedGasLimit is the minimum required gas limit for encrypted transaction.
+// It is set to be equal to a simple transfer execution cost since it is assumed that
+// minimum valid encrypted transaction structure is a simple transfer.
+const MinEncryptedGasLimit uint32 = 21000
```
txhsl: We can use `params.TxGas` here.

https://github.com/bane-labs/go-ethereum/blob/5b787b6806f340f1f978d7998d9eca24b5f65888/params/protocol_params.go#L36

contracts/solidity/GovReward.sol
AnnaShaleva: > otherwise there is a potential risk that requiredSpace < 21000.

But how? Consensus will reject such Envelopes as invalid,so they won't even be processed. If you still think that there's a possibility of consensus accepting an invalid Envelope, then we need to fix it on the consensus side.

antimev/envelope.go
AnnaShaleva: Good catch, I wasn't aware of this constant.

core/state_transition.go
AnnaShaleva: Fixed.

antimev/envelope.go
AnnaShaleva: Fixed. But I'd still suggest to keep `MinEncryptedGasLimit` in a separate constant because it's convinient for us to keep use it as `uint32` (`TxGas` is `uint64`, so conversion is needed every time we use it).

contracts/solidity/GovReward.sol
txhsl: I know it will be impossible, but from the contract perspective, it is an underflow risk in coding. You can regard it as a double check, it will be helpful for first reading, without any further knowledge about our Geth code.

antimev/envelope.go
txhsl: > Fixed. But I'd still suggest to keep `MinEncryptedGasLimit` in a separate constant because it's convinient for us to keep use it as `uint32` (`TxGas` is `uint64`, so conversion is needed every time we use it).

Yes, good enough.

contracts/solidity/GovReward.sol
AnnaShaleva: The only thing that comes to me is that this constraint may be violated by gas estimator because estimator doesn't verify transaction. So for gas estimator this check is still needed.

Do you have any other use-cases in mind?

txhsl: > Do you have any other use-cases in mind?

No more.

txhsl (approved):

contracts/solidity/GovReward.sol
AnnaShaleva: Good, then let's keep it for estimator. Reverted.

AnnaShaleva:
Only commit description is changed with the last update, the commit content is still the same.

PR #433, opened 2025-03-07
contracts: register key when register candidate and refactor tests

### LINKED ISSUE #432: Bind message key registration to Governance
# Rationale

We're now registering the message keys for existing member nodes in the KeyManagement contract. But further, we'd better require this key when new candidates interacting with the Governance contract.

It seems that the compiler version doesn't affect the key registration before `NeoXDKGBlock`, which means it won't destroy the Governance service.

# Implementation

WIP.

txhsl:
Close #432. The changes in `core/systemcontracts/contracts.go` doesn't affect it functionality, so we can enable this new registration before `v0.3.1`.


txhsl:
https://github.com/bane-labs/go-ethereum/pull/433/commits/7ad1569cb4a23206ea76503f3ff7ff9d565fac26#diff-3ee4350df529a3267300200ca837fc6ee7f2c3058fdbd6452b02459c1186c256R72

Well, the entry I left for Governance has a bug (direct registration is not affected, so no effect on T4 now), so we may have to update both two T4 contracts to enable this feature (first KeyManagement then Governance).

contracts/solidity/KeyManagement.sol
```diff
@@ -69,7 +69,7 @@ contract KeyManagement is GovProxyUpgradeable, IKeyManagement {
             revert Errors.SideCallNotAllowed();
         if (tx.origin != candidate) revert Errors.OnlyEOA();
         if (pubkey.length != 65) revert Errors.InvalidMessageKey();
-        messagePubkeys[msg.sender] = pubkey;
+        messagePubkeys[candidate] = pubkey;
```
AnnaShaleva, line 72: > Well, the entry I left for Governance has a bug

I see, so it won't work correctly for the case of `GOV` sender. Agree, let's update both contracts on testnet before NeoXDKG fork.

AnnaShaleva (approved):

contracts/solidity/KeyManagement.sol
txhsl: > Agree, let's update both contracts on testnet before NeoXDKG fork.

Let's do it when the Governance frontend is ready, it doesn't prevent us from `NeoXDKG`.

Issue #372, opened 2024-12-24
Refactor amev transaction management inside mempool


AnnaShaleva:
# Rationale

Ref. https://github.com/bane-labs/go-ethereum/pull/316#discussion_r1873097905. We should consider encrypted transactions management to be performed in a separate subpool instead of using the same legacy pool for both envelopes and encrypted transactions.

# Implementation

The node should be able to maintain both legacy pool (for envelopes) and encrypted pool (for encrypted transactions). Account's reservation logic should be adjusted as far:
https://github.com/bane-labs/go-ethereum/blob/9b63444dbd3343389533e5c31bfe1126342b4004/core/txpool/txpool.go#L111.

Issue #329, opened 2024-10-04
Always use Policy-based block BaseFee


AnnaShaleva:
# Rationale

f06dad44ee26d71688d2ea356919c36893cfbf09 (#328) removes NeoXBurn fork and makes block's BaseFee to be Policy-based for dBFT-enabled nodes. However, nodes with another consensus engine still use an old-fashioned London way of BaseFee calculation.

# Implementation

Always use Policy-based fee calculation behaviour irrespectively of consensus engine settings. It allows to get rid of some useless code that is never invoked in NeoX. Pitfalls: a lot of tests should be migrated onto new BaseFee calculation scheme or disabled.

roman-khimov:
It's OK for us, NeoX does not have any other consensus mechanism.

AnnaShaleva:
Yes, I know it’s OK, but it requires unit-tests rework, that’s why I moved it to a separate issue.

txhsl:
I tried to run a full test of Geth, and got the following failures.

```
--- FAIL: TestGenerateEncryptedTx (7.74s)
    tpke_test.go:151: 
        	Error Trace:	/home/linux/Desktop/Bane/go-ethereum/antimev/tpke_test.go:151
        	Error:      	Received unexpected error:
        	            	no global public key
        	Test:       	TestGenerateEncryptedTx
FAIL

......

--- FAIL: TestUnderpricingDynamicFee (0.10s)
    legacypool_test.go:1977: adding underpriced pending transaction error mismatch: have <nil>, want transaction underpriced
--- FAIL: TestUnderpricing (0.08s)
    legacypool_test.go:1796: adding underpriced pending transaction error mismatch: have <nil>, want transaction underpriced
--- FAIL: TestDualHeapEviction (0.15s)
    legacypool_test.go:2070: transaction count mismatch: have 30, want 10
FAIL
```

Yes, we need to fix them. But, hmm, if the contract storage of the test chain is empty, then our fee policies should be all `0` so that not affecting the test results, so we don't need to rework them.

OK, so these UT failures are #316 and #372 related, since #316 increased the number of slots.
https://github.com/bane-labs/go-ethereum/blob/28a4058845e358d5f63a2fcd720ba4500b00fb6e/core/txpool/legacypool/legacypool.go#L810-L830

PR #441, opened 2025-03-20
params: start v0.3.3 release cycle and extend Envelope generator test

AnnaShaleva:
@txhsl, please review.

### DIFF
--- a/antimev/tpke_test.go
+++ b/antimev/tpke_test.go
@@ -2,6 +2,7 @@ package antimev
 
 import (
 	"bytes"
+	"context"
 	"encoding/binary"
 	"encoding/hex"
 	"fmt"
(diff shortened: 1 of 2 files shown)


antimev/tpke_test.go
```diff
@@ -132,18 +136,29 @@ func TestInitKeyStores(t *testing.T) {
...
-	// epoch is an epoch of Envelope data encryption. The resulting encrypted transaction can be
-	// decrypted only during this or next epoch.
-	const epoch = 1
+	const (
+		skip = true
+		send = false
+		// rpc1 is an RPC endpoint address of privnet's CN1.
+		rpc1 = "http://localhost:8562"
+	)
+	if skip {
```
roman-khimov, line 145: It was a proper test, better keep it and add another one that interacts with some node.

AnnaShaleva: This test was failing since those times we introduced dynamic DKG keys into privnet (yes, we have failing tests in master).

This test requires antimev keystore to be properly setup and to pass at least initial DKG round.

So no, it wasn’t a proper test.

roman-khimov: > yes, we have failing tests in master

I've not heard and seen this.

roman-khimov (approved):

txhsl (approved):

antimev/tpke_test.go
txhsl: https://github.com/bane-labs/go-ethereum/issues/329#issuecomment-2713147948

TestGenerateEncryptedTx is one of them, and the rest should be caused by #316. But I don't a good-enough idea to fix them, since it's about the slot management in txpool and we changed it.

Not related to this PR, but #372. @AnnaShaleva should we first improve the encrypted mempool, or just refactor it directly?

PR #444, opened 2025-03-28
consensus/dbft: add a new goroutine for ZK proof generation

### LINKED ISSUE #443: Add a new routine for ZK proof generation in dBFT
# Rationale

Since we are going to enforce a ZK verification in the DKG smart contract, a ZK prover becomes necessary to handle the output of anti-mev keystores after `handleDKG()`.

Our first idea is generating the proof data inside the Geth node, instead of a separated/centralized prover or docker, so that we can avoid related message exchange and protocols.

# Implementation

Working on https://github.com/bane-labs/go-ethereum/tree/zk-dkg.

txhsl:
Close #443. This PR is after #442, ready for review, and now waiting for ZK-related tests.

The basic idea is to add a new routine after `handleDKG()` but before `loopTaskList()` to deal with the parameter generation of our DKG transactions, so the message encryption/encode is also moved from keystore logic to this new routine.

The feature of zk-SNARK limits that we can only prepare the setup parameters for a fixed network size, so we may only enable ZK-DKG on 7-CN networks. Different setups require redundant codes, and we are not going to use them for production, but I still expect the Geth code compatible with all of our privnets.

As the result, a new option `zkdkg` about enabling ZK or not is added to the genesis file, in the same place as the contract allocation. The new code still works with 1-CN and 4-CN privnets with their past settings, but the 7-CN is updated to enable the ZK feature with a different contract allocation.

I tried to put everything about the contract in `0x1212100000000000000000000000000000000008`, but the compiled code was just too large for a proper deployment, so several new slots start with `0x12122` are used for verifier libraries.

@AnnaShaleva The code works on privnets without ZK, ~~but not yet verified on the new 7-CN privnet~~ and verified with ZK.


txhsl:
Seems too weak in proving performance, trying to solve.

txhsl:
I think we have maximized the proving efficiency under Groth16, with `zk-dkg@v0.2.1`. No compilation happens in Geth now.

@songb2 You can have tests on a 7-CN network now. 

1. Collect 7 machines with each at least 64 GB memory, and download the Geth code from this PR;
2. Use the genesis file, wallets and keystore files in `/privnet/seven` to init the new network;
3. Download the R1CS and the proving key from https://neochain1.blob.core.windows.net/fileshare/mengyu20250401.zip, which contains the necessary ZK parameters for DKG (only share and reshare without recover);
4. Add new flags `OneMsgR1CSPathFlag`, `TwoMsgR1CSPathFlag`, `SevenMsgR1CSPathFlag`, `OneMsgProvingKeyPathFlag`, `TwoMsgProvingKeyPathFlag` and `SevenMsgProvingKeyPathFlag` when starting these Geth CNs. Since we only have `R1CS_7` and `PK_7`, you can leave the left 4 flags to mock files;
5. Check if the network can perform ZK-DKG and change epochs correctly.

About the network settings.

1. The epoch time is ~~8~~ 2 hours and the window for each DKG period is ~~4 hours~~ 30 mins. ~~The latter is the same as T4, so if 4 hours is not enough for the 2nd round (has both share and reshare) and the enabling of BLS fails, you need to evaluate and extend the `sharePeriodDuration`. It also means we need to adjust the same setting of T4~~;
2. You may extend the `sharePeriodDuration` at the beginning to check the proving time, it's also a proper solution;
3. After we can have the network perform 2 rounds of successful DKG, you can ask @doubiliu to generate the left 4 files for DKG recover.

I'd expect the following results from this test.

1. The DKG with ZK proofs is whether practical;
2. The minimum hardware requirement of a CN, especially the memory usage (e.g. 64 GB or 32 GB);
4. The proper setting of `sharePeriodDuration` is 4 hours or not.

I will be in HongKong for the next week, so DM me online if there is any issue.

consensus/dbft/dbft.go
```diff
@@ -1480,6 +1489,20 @@ func (c *DBFT) WithTxPool(pool txPool) {
 	c.txpool = pool
 }
 
+// WithZKFiles initializes the paths of ZK required files.
+func (c *DBFT) WithZKFiles(r1cs1, r1cs2, r1cs3, pk1, pk2, pk3 string) error {
+	if c.config.ZKDKG && (len(r1cs1) < 1 || len(r1cs2) < 1 || len(r1cs3) < 1 || len(pk1) < 1 || len(pk2) < 1 || len(pk3) < 1) {
```
AnnaShaleva: Nitpicking, but usually it's `len(r1cs1) == 0` and etc.

consensus/dbft/dbft.go
```diff
@@ -236,6 +237,13 @@ type config struct {
 	antiMEVEnablingHeight  int64
 	enforceECDSASignatures bool
 	logLevel               *zap.AtomicLevel
+	// ZK-DKG required MPC files.
+	oneMsgR1CSFile         string
+	twoMsgR1CSFile         string
+	sevenMsgR1CSFile       string
+	oneMsgProvingKeyFile   string
+	twoMsgProvingKeyFile   string
+	sevenMsgProvingKeyFile string
```
AnnaShaleva: `config` is an Engine's constructor level structure, but these files are not setup during engine construction. Since we use a separate callback to setup these files (`WithZKFiles`), let's move them to a separate structure and bind it to `DBFT` structure directly.

consensus/dbft/dkg.go
```diff
@@ -164,44 +175,51 @@ func (c *DBFT) handleDKG(snapshot *Snapshot, keystore *antimev.KeyStore, h *type
...
-		// Send watch task list to loopTaskChan when handleDKG finished
+		// Send pending task list to execution when handleDKG finished.
 		defer func() {
-			c.loopTaskChan <- watchList
+			if len(taskList) > 0 {
+				if len(c.executeProofTaskChan) < cap(c.executeProofTaskChan) {
+					c.executeProofTaskChan <- &taskList
+				} else {
+					// Panic here since the channel is going to block DKG and dBFT.
+					panic(fmt.Errorf("proof task channel is full: cap %d, length %d", len(c.executeProofTaskChan), cap(c.executeProofTaskChan)))
```
AnnaShaleva: But are you sure we can panic here? Because it looks to me that we may easily end up in a situation when `executeProofTaskChan` already has two pending tasks in queue by the moment we'll try to enqueue one more.

What do you think about silently skipping this task instead of panic?

consensus/dbft/dbft.go
```diff
@@ -2283,7 +2307,8 @@ drainLoop:
 	close(c.messages)
 	close(c.txs)
 	close(c.chainHeadEvents)
-	close(c.loopTaskChan)
+	close(c.executeProofTaskChan)
+	close(c.loopTransactionTaskChan)
```
AnnaShaleva: We can't just close these channels, it may lead to the panic during the node stop. We need to wait until both `loopExecuteProofTask` and `loopTransactionTaskList` are finished their work and exit, and only after that close these channels.

> are finished their work and exit,

To implement that we may to watch over `c.quit` channel like it is done in DBFT's `eventLoop`. If this channel is closed, then it's a signal to `loopExecuteProofTask` and `loopTransactionTaskList` routines to finalize their work.

It's important because these two routines are heavy now, and we need to properly release resources on the node's shutdown.

AnnaShaleva:
I'll need some time to review it carefully.

consensus/dbft/dbft.go
AnnaShaleva: Or may be it's not needed, my primary concern is that no one should be able to send anything to these channels after we close them.

consensus/dbft/dkg.go
txhsl: > What do you think about silently skipping this task instead of panic?

We have 2 periods in every round of DKG, and CNs may have the following tasks to execute.

1. Share and Reshare in the 1st period. They happen at the same time, so arranged in the same list;
2. Recover in the 2nd period. This is the second list, but optional;
3. ReshareRecoverd in the 2nd. This relies on the result of the above Recover, so is the third list, but also optional.

As a result, a CN can have at most 3 task lists for each epoch. The channel cap is `2`, so when it's full, and we still try to enqueue one more, it must be after the beginning of the next DKG. E.g. on T4, it's 3.5 days after, and the computation of the first task list is still not finished (generally it should be done in 4 hours).

The panic here is preventing a CN to continue the DKG, like an OOM message, because its physical hardware is far away from the minimum requirement to compute ZK.

In a perfect DKG, this channel should get cleaned before the next task list comes. When it is full, then the node is already in an extreme condition.

> when `executeProofTaskChan` already has two pending tasks in queue by the moment we'll try to enqueue one more.

We can silently skip here, whatever we don't expect a full channel to be normal, but which one it better?

consensus/dbft/dbft.go
txhsl: Yes, definitely better.

Updated.

consensus/dbft/dkg.go
txhsl: Maybe we just return an error here, but it is not allowed in `defer`.

OK, I changed the code order. Now it returns an error rather than panic.

txhsl:
The current code is verified on privnet, working with the experimental ceremony mentioned in #442.

image

The recommended physical memory of our CN will increase to 32 GB, since the top usage here is about 21 GB.

The proof generation of a pair of Share and Reshare (which is the heaviest job) can be executed within 30 mins, so we don't have to change the T4 settings. The benchmark was run with an AMD EPYC 7B13, but each machine only used 4 cores of it.

image

Let me take three examples to explain what happened with the new code.

- The first wave is the Share of the 1st round DKG. The node read a R1CS file and a PK file into its memory, which pumped the memory usage to about 10 GB. And then is the ZK proving, the memory usage increased to around 17 GB, and got fully GCed in the end.

- The second wave is the Share and Reshare of the 2nd round DKG. The node read the same R1CS file and PK file into its memory, but computed two proofs one by one, so it increased to 21 GB.

- The third wave is the Recover of the 2nd round DKG. We made a CN offline, then it happened. The R1CS file and PK file were smaller, so the memory usage and execution time both decreased.

Since the execution time is acceptable, I don't keep the R1CS files and PK files in memory, otherwise the three pairs of them may cap 20 GB in total.

consensus/dbft/dbft.go
txhsl: I add the new one where we close `loopTaskChan`, based on `v0.3.2`.

I haven't seen any error in my testing, but the channel is closed before the loop quits.

```
INFO [04-11|17:29:43.932] dBFT event loop finished
INFO [04-11|17:29:43.933] DKG transaction dispatcher stopped
INFO [04-11|17:29:43.937] DKG proof dispatcher stopped
```

Is it an issue on T4 as well?

AnnaShaleva: I see that this code is implemented the same way as `loopTaskChan`, so originally I thought that there's a bug in `loopTaskChan`, but I re-checked the code, and there's no bug. The only routine that can submit a task to `loopExecuteProofTask` or to `loopTransactionTaskList` is `handleDKG` routine. The only caller of `handleDKG` is dBFT's `eventLoop`. So we close these two channels after main loop of `eventLoop` is exited, so everything is OK here and no changes are required.

consensus/dbft/dkg.go
```diff
@@ -337,35 +345,226 @@ func (c *DBFT) handleDKG(snapshot *Snapshot, keystore *antimev.KeyStore, h *type
...
+				}
+				for i, idx := range indexes {
+					idxsBigInt[i] = big.NewInt(int64(idx))
+				}
+				// Compute necessary inputs for circuit.
+				fisBytes, fisInts, bigFis, nonces, encryptedFis, rs, bigRs := circuit.PrepareEncryptedKeyShares(pubs, fis)
+				// Update the task parameters for sending a transaction.
+				msgs := encodeMessages(encryptedFis, bigRs, nonces)
+				// Send transactions based on ZK settings.
+				if c.config.ZKDKG {
```
AnnaShaleva: We need to move this logic to some `NeoXZKDKG` fork and enable it after `NeoXAMEV` fork. We need this  for two reasons:
1. For testnet safety. With this new fork the updated T4 CNs can start this ZK-based logic simultaneously from the particular height without any risk of divergence in behaviour (due to delays on CNs update or delays in config update).
2.  For mainnet safety. I remember that the original plan was to enable ZK-based DKG in mainnet without non-ZK transition period. But the problem is that this scheme is not tested on T4, and if something unexpected happens during mainnet update, we'll be in trouble. That's why we think that it's more safe to update mainnet using exactly the same algorithm that was used for T4 update: move ZK logic to a separate fork; enable NeoXDKG, then enable NeoXAMEV, then enable NeoXZKDKG. There's no need to wait for a long time between these forks; the main idea is to perform the upgrade step-by-step.

So we probably need to get @steven1227's opinion on that.

txhsl: Maybe a new fork is not that necessary. The ZK-based logic doesn't require any critical divergence in dBFT.

1. For testnet, the DKG behaviours rely on the verification of KeyManagement contract. The new ZK-based logic changes the ABI of all message methods, including `share`, `reshare`, `recover` and `reshareRecovered`, by only adding the following 3 new parameters. Once the contract is updated, its ABI changes, but any accepted DKG message before ZK will still be valid after upgrade. We have no risk of diverged behaviours, but maybe 1 or 2 rounds of failed DKG and the Governance won't change CNs, and the failure of DKG transactions won't block dBFT or cause panic on CNs;
```solidity
uint[8] calldata proof,
uint[2] calldata commitments,
uint[2] calldata commitmentPok
```
3. If we want to avoid failed rounds in a new-fork solution, then we need to add `V2` methods in the KeyManagement contract, instead of changing them. But it sounds worse to me, because these two kinds of verification should not exist at the same time;
4. For mainnet, if we can upgrade T4 without any new fork, then mainnet is safe to upgrade as well. I mean, the things we are going to change belong to `NeoXDKG`, which is a quite independent logic before `NeoXAMEV`. It doesn't enable any new feature which may affect the global state transition, and we can still wait 2 epochs to see if it works. Futhermore, the new code is still compatible with `NeoXAMEV`, so the old steps we used to upgrade T4 are still valid;
5. BTW, the switch of ZK is now in the genesis file, so if we upgrade to the new code on T4 without `geth init`, then it works as well. So we can first update the node code, then upgrade the contract to change the verification rules. Then, all CNs need to set `config.dbft.zkdkg` to enable the new DKG behaviours, otherwise the contract will reject all of their DKG transactions.

I know the safest way of enabling new features is a new fork. I just think it might be redundant since this doesn't affect state transition. It won't cause any case that CNs try to accept some illegal data into new blocks to cause a consensus fork, this is my consideration of using a new fork.

txhsl: OK… Now we have #450 to solve, it's possible to merge them in the same fork if necessary.

But still, I think we should avoid using hardfork for this, because it's possible that we need to update this again for optimization.

consensus/dbft/dkg.go
```diff
@@ -230,34 +234,36 @@ func (c *DBFT) handleDKG(snapshot *Snapshot, keystore *antimev.KeyStore, h *type
...
-			// If is a member of pending consensus
+			// If is a member of pending consensus.
 			indexOfSharing := slices.Index(snapshot.PendingCNs, amevAddress) + 1
 			if err != nil {
 				return fmt.Errorf("failed to get message keys, err: %v", err)
 			}
 			if indexOfSharing > 0 {
-				if err = c.taskShare(keystore, receiverMessageKeys, currentHeight, recoverStartHeight, watchList); err != nil {
+				taskList, err = taskList.taskShare(keystore, receiverMessageKeys, currentHeight, recoverStartHeight)
+				if err != nil {
```
AnnaShaleva, line 261: A nitpicking, but it's not the most common way in Go to work with slice. Consider this:
```
 // taskShare tries to prepare necessary inputs for a contract calling for DKG share.
-func (t *taskList) taskShare(keystore *antimev.KeyStore, receiverMessageKeys []*ecies.PublicKey, start uint64, end uint64) (taskList, error) {
+func (t *taskList) taskShare(keystore *antimev.KeyStore, receiverMessageKeys []*ecies.PublicKey, start uint64, end uint64) error {
        secrets, sPvss, err := keystore.DKGShare()
        if err != nil {
-               return nil, err
+               return err
        }
        // Set up a new task with keystore outputs.
-       return t.append(&task{SendHeight: start, EndHeight: end, Method: "share", Params: []interface{}{secrets, sPvss, receiverMessageKeys}}), nil
+       *t = append(*t, &task{SendHeight: start, EndHeight: end, Method: "share", Params: []interface{}{secrets, sPvss, receiverMessageKeys}})
+       return nil
 }
```

and then these two helpers can be removed:
```
-// append adds a new element to the task list.
-func (l *taskList) append(t *task) taskList {
-       return append(l.flatten(), t)
-}
-
-// flatten returns the tasks in the task list.
-func (l *taskList) flatten() []*task {
-       return ([]*task)(*l)
-}
```

consensus/dbft/dkg.go
```diff
@@ -337,35 +345,226 @@ func (c *DBFT) handleDKG(snapshot *Snapshot, keystore *antimev.KeyStore, h *type
...
 			return fmt.Errorf("failed to persist keystore, err: %v", err)
 		}
+		if len(taskList) > 0 {
+			if len(c.executeProofTaskChan) < cap(c.executeProofTaskChan) {
+				c.executeProofTaskChan <- &taskList
+			} else {
+				// Give up the tasks since the channel is going to block DKG and dBFT.
+				return fmt.Errorf("proof task channel is full: cap %d, length %d", len(c.executeProofTaskChan), cap(c.executeProofTaskChan))
+			}
+		}
```
AnnaShaleva: This check is racy by default. Use the following construction:
```
                if len(taskList) > 0 {
-                       if len(c.executeProofTaskChan) < cap(c.executeProofTaskChan) {
-                               c.executeProofTaskChan <- &taskList
-                       } else {
+                       select {
+                       case c.executeProofTaskChan <- &taskList:
+                       default:
                                // Give up the tasks since the channel is going to block DKG and dBFT.
-                               return fmt.Errorf("proof task channel is full: cap %d, length %d", len(c.executeProofTaskChan), cap(c.executeProofTaskChan))
+                               return fmt.Errorf("proof task channel is full")
                        }
                }
        }
```

consensus/dbft/dkg.go
```diff
@@ -337,35 +345,226 @@ func (c *DBFT) handleDKG(snapshot *Snapshot, keystore *antimev.KeyStore, h *type
...
+
+	for pendingList := range c.executeProofTaskChan {
+		log.Info("DKG loopExecuteProofTask", "pendingListLength", len(pendingList.flatten()))
+
+		// Share and reshare are always in the same task list for proving, so here is a cache to reduce file read.
+		// 1 and 2 message proving happen at most once per epoch, and not together with a 7, so no cache for them.
+		// A task list to prove is consist of 7+7 (length is 2), 1 or 2 or 7 (length is 1), no other case for now.
+		// So this cache should not alloc memory when a 1 or 2 message proving is going to happen.
+		var sevenMsgR1CS constraint.ConstraintSystem
+		var sevenMsgPK *groth16.ProvingKey
```
AnnaShaleva: Usually for multiple variables declaration we use:
```
var (
        ...
)
```

consensus/dbft/dkg.go
```diff
@@ -337,35 +345,226 @@ func (c *DBFT) handleDKG(snapshot *Snapshot, keystore *antimev.KeyStore, h *type
...
+
+		// Handle new tasks immediately but one by one.
+		watchList := taskList(make([]*task, 0))
+		for _, task := range pendingList.flatten() {
```
AnnaShaleva: s/`pendingList.flatten()`/`*pendingList` should work.

consensus/dbft/dkg.go
```diff
@@ -337,35 +345,226 @@ func (c *DBFT) handleDKG(snapshot *Snapshot, keystore *antimev.KeyStore, h *type
...
+			switch task.Method {
+			case "share", "reshare", "reshareRecovered":
+				// Prove the encryption of seven new messages.
+				ss := task.Params[0].([]*big.Int)
+				pvss := task.Params[1].([]byte)
+				pubs := task.Params[2].([]*ecies.PublicKey)
+				if len(ss) != len(pubs) {
+					log.Error("share task array not match", "secrets", len(ss), "keys", len(pubs))
```
AnnaShaleva: Can we also log method? And to be honest, it looks like a program bug (panic) rather than an error to me since task is constructed improperly.

consensus/dbft/dkg.go
AnnaShaleva: I'm wondering how big these files are in production? Can we read at least proving key once on dBFT engine start and then reuse these data? If so, then error on file reading may be handled right during the node startup.

consensus/dbft/dkg.go
```diff
@@ -337,35 +345,226 @@ func (c *DBFT) handleDKG(snapshot *Snapshot, keystore *antimev.KeyStore, h *type
...
+						provingKey, err = helper.ReadProvingKey(c.zkFiles.twoMsgProvingKeyPath)
+						if err != nil {
+							log.Error("invalid proving key file", "file", c.zkFiles.twoMsgProvingKeyPath, "err", err)
+							continue
+						}
+					default:
+						// The circuit setup is limited for proofs of 1-or-2-message tasks, other cases shouldn't happen.
+						panic(fmt.Errorf("invalid number of message inputs: expect 1 or 2, get %d", len(fis)))
+					}
+					proof, _, err := zkdkg.ProveMultipleKeyShareEncryption(r1cs, provingKey, pubs, rs, bigRs, fisBytes, fisInts, bigFis, encryptedFis, nonces)
```
AnnaShaleva, line 515: This is a time-consuming operation, IINM. Usually such operations use cancellation context so that if the node is interrupted by the operator in the middle of the proof generation, this routine won't hang in the air. Introducing cancellation context will require ZK library modification, so we can move it to a separate task, but in general I think it's a good practice to support graceful stop for this service.

consensus/dbft/dkg.go
```diff
@@ -337,35 +345,226 @@ func (c *DBFT) handleDKG(snapshot *Snapshot, keystore *antimev.KeyStore, h *type
...
+// loopTransactionTaskList retries every transaction task in transaction watch list.
+func (c *DBFT) loopTransactionTaskList() {
+	log.Info("DKG transaction dispatcher started")
+	defer log.Info("DKG transaction dispatcher stopped")
+
+	for watchList := range c.loopTransactionTaskChan {
+		currentHeight := (*c.backend).CurrentBlock().Number.Uint64()
+		log.Info("DKG loopTransactionTaskList", "currentHeight", currentHeight, "watchListLength", len(watchList.flatten()))
+		// Append watchList from loopTransactionTaskChan to c.transactionTaskList.
+		c.transactionTaskList = append(c.transactionTaskList, watchList.flatten()...)
```
AnnaShaleva: We need to untie `transactionTaskList` from `c` because it's a local variable for `loopTransactionTaskList` in fact. Otherwise it's not safe to use `c.transactionTaskList` from any other thread.

consensus/dbft/dkg.go
```diff
@@ -396,14 +595,14 @@ func (c *DBFT) loopTaskList() {
 					}
 				}
 			}
-			// Only keep retry failed and not reach max retry times
-			c.txWatchList = retryList
+			// Only keep retry failed and not reach max retry times.
+			c.transactionTaskList = retryList
```
AnnaShaleva: An attempt to resend tasks from `retryList` will be taken only after the next item will be received from `c.loopTransactionTaskChan` channel. We send tasks to `c.loopTransactionTaskCha` only from `handleDkg` routine only if `len(taskList) > 0` in fact:
https://github.com/bane-labs/go-ethereum/blob/5a505468681a57ba4c3c77cf72346dca7beff6f5/consensus/dbft/dkg.go#L348-L351
So my primary concern here is that resend action won't be triggered every block, it will be triggered much less frequently (correct me if I'm wrong please).

consensus/dbft/dkg.go
AnnaShaleva: > We have no risk of diverged behaviours, but maybe 1 or 2 rounds of failed DKG and the Governance won't change CNs, and the failure of DKG transactions won't block dBFT or cause panic on CNs;

But my concern is that we will have failing DKG transactions on testnet (and mainnet) during the transition period (the period after contract update but before DKG enabling in CNs configuration).

> If we want to avoid failed rounds in a new-fork solution, then we need to add V2 methods in the KeyManagement contract, instead of changing them.

Agree.

> But it sounds worse to me, because these two kinds of verification should not exist at the same time;

It's not a very big problem to me, because all these methods of KeyManagement contract are allowed to be called by current consensus nodes only, so it's a limited number of "trusted" callers during transition period.

I agree with your argument that this code won't affect chain states, but my concern is mostly about the fact that that this code will change the CNs behaviour in runtime, and if `config.dbft.zkdkg` won't be unified across all CNs, we will end up in a situation when some CNs are trying to use old way of DKG setup whereas other CNs will use this new ZK-based code. So some CN transactions will fail whereas other won't. In general it won't hurt much, because eventually `config.dbft.zkdkg` will be enabled on all CNs, but at the same time with hardfork we can implement a seamless switch in the nodes behaviour. Although, as you said, it will cost us additional methods support at the contract level and additional hardfork-based switch at the `handleDKG` routine level. I'm not sure does it worth what it cost, @roman-khimov, what do you think?

txhsl: OK, I see your concern.

So the best option seems like the node can choose its behavior based on the contract ABI.

> It's not a very big problem to me, because all these methods of KeyManagement contract are allowed to be called by current consensus nodes only, so it's a limited number of "trusted" callers during transition period.

However, allowing different versions of DKG sharing at the same time is not an option. These methods are called by not only current but also incoming consensus.

In our crypto assumption, they're not 100% trusted. That's also why we import ZK here to protect the DKG.

> In general it won't hurt much, because eventually `config.dbft.zkdkg` will be enabled on all CNs, but at the same time with hardfork we can implement a seamless switch in the nodes behaviour.

We have tried the hard fork solution once before, when we upgraded the Governance from V1 to V2. Consensus is forked because we allowed two different versions of Governance `OnPersist`, and this behavior influenced the final state hash.

But for the DKG side, if we don't have the hard fork in Governance `OnPersist` and still have latency in the genesis update, CNs also might not send their DKG transactions in time and cause a failed epoch change. 

It's the same situation as what we have now, a hard fork will not solve the misbehavior thoroughly. If the new fork doesn't affect the uniqueness of a new block, dBFT will not stop for us, but continue as it is.

If we want a consistent upgrade style as `NeoXDKG`, then we can have a new one for `NeoXZKDKG`. It seems better than a new dBFT config, I agree.

But again, the latency still matters, because the hard fork is not really “hard”. Your worry is not solved with it.

We have the flexibility of DKG upgrade without hard forks, but now we are going to give it up.

> But my concern is that we will have failing DKG transactions on testnet (and mainnet) during the transition period (the period after contract update but before DKG enabling in CNs configuration).

BTW, I have to mention these “failed” transactions will not really be sent and spend fees. Because the pre-execution will drop errors and prevent the CN from sending it.

They will be only some error messages in our logs. If my description is misleading, then I apologize for it.

https://github.com/bane-labs/go-ethereum/blob/c07c8500b3786c284cba343cd7cb40c79f04916e/internal/ethapi/api.go#L1942

Here contains a `DoEstimateGas` operation, so it must fail with error `execution reverted`, as well as the retry.

consensus/dbft/dkg.go
txhsl: Yes, it's important.

I've checked that `gnark` doesn't support such a usage now, so we need to request this feature, or in the worst case, we need to implement it on our own.

consensus/dbft/dkg.go
txhsl: You're right, let me fix it.

consensus/dbft/dkg.go
txhsl: The file size is smaller than 2 GB, but when it gets read into memory, the consumption is too heavy for us now. I have data for this `sevenMsgR1CS` and `sevenMsgPK`, which shows a cost about 10 GB if we want to reuse them. The proving key is huge as well (even larger than CSS), because the more constraints we have, the more commitments will be there in the key.

6ce5a3c8ca8657df8c5fb8d53b57cd7

And considering we have 3 pairs of pk and css, the reuse seems more impossible, since it will allocate about 20 GB memory.

But we are now developing the second version of our circuits, with plonk and recursion as optimization, ref https://github.com/bane-labs/zk-dkg/tree/recursion_plonk_groth16. I think we can implement this suggestion after that, because we will have only 1 pair of pk and css file, and they will be in the smallest size we can expect.

If we complete the optimization in `zk-dkg` quite fast, then I will upgrade to use it here directly.

consensus/dbft/dkg.go
txhsl: > Can we also log method?

Yes, of course.

> And to be honest, it looks like a program bug (panic) rather than an error to me since task is constructed improperly.

Yes, this is rather a bug check, then I panic here directly.

consensus/dbft/dkg.go
txhsl: I refactor this part with minor change. How about it now?

consensus/dbft/dkg.go
AnnaShaleva: > which shows a cost about 10 GB if we want to reuse them. 

Unacceptable, sure.

> If we complete the optimization in zk-dkg quite fast, then I will upgrade to use it here directly.

Will also depend on the final files size.

consensus/dbft/dkg.go
AnnaShaleva: We discussed this issue with Roman, so the key question is:

> So the best option seems like the node can choose its behavior based on the contract ABI.

Right now we store ABI as a constant in the node's code. Can we implement your suggestion as the following: every block in `handleDKG` retrieve the current ABI of KeyManagement contract (we have access to the node's state, so it should be possible, may be we can introduce some ContractVersion constant to the KeyManagement contract) and check if it is updated to the new version. Once it's updated, we switch to the new scheme of keys generation and stop KeyManagement updates monitoring. If it's possible then we agree that we don't need a fork and we don't even need a special `config.dbft.zkdkg` setting, it should be completely replaced by this new ABI autodetection.

consensus/dbft/dkg.go
doubiliu: Currently，a method that can be  used is to use  a recursive circuit to split the originally large-sized computing task into sub-tasks, and then insert the signal. For example: 7-node proof circuit = 7 * single-node proof circuit. I'm implementing related methods in zk-dkg. But if a single proof task is still time-consuming, this method may be useless

consensus/dbft/dkg.go
txhsl: Good idea. But the ABI is not a thing that stored in state, so we still need to hardcode it as a constant.

I'll try to implement this and let's see how can we optimize it.

AnnaShaleva: > But the ABI is not a thing that stored in state

We don't need the full ABI in fact, for that purpose I have two options in mind:
1. For versioning purpose we may implement some additional `GetVersion()` method on the contract side to return contract version, and perform test invocation of this method on dBFT side to make a behaviour switch depending on contract version. Then before ZK-related KeyManagement update we should manually upgrade the contract version in the contract's code.
2. Or we may go without any additional KeyManagement modifications and on dBFT side try to testinvoke old method (with old ABI) and check if invocation fails with specific error (due to method missing in contract). If it fails, then just invoke new method (with new ABI). On N3 it's possible, but I'm not sure if it's possible with Eth contracts to know the exact reason of contract invocation failure because we use proxy contracts.

So option 1 looks a bit more reliable to me. But I'm also open to any other suggestions.

txhsl: Option 1 is definitely better, and I'd say it's practical.

We first upgrade the node, then let it check the DKG version by itself.

txhsl: Implemented.

txhsl:
Since we've added a new method `ZK_VERSION()` to monitor the contract version about DKG, we need to upgrade the `KeyManagement` contract on T4 before we upgrade anything else.

consensus/dbft/dkg.go
txhsl: https://github.com/Consensys/gnark/issues/1496.

Anyway, this issue can not be solved immediately. We need a new version of `bane-labs/zk-dkg` or `Consensys/gnark` to improve this part.

consensus/dbft/dkg.go
txhsl: > I think we can implement this suggestion after that, because we will have only 1 pair of pk and css file, and they will be in the smallest size we can expect.

No, it's not. I checked the implementation of recursion proof, we still need 3 pairs of these files to generate different proofs. So “read during startup” is still not a practical choice.

The optimization itself will save days in ZK setup, but won't help us in optimizing the memory usage.

AnnaShaleva (approved):
Much cleaner this way, LGTM.

> Since we've added a new method ZK_VERSION() to monitor the contract version about DKG, we need to upgrade the KeyManagement contract on T4 before we upgrade anything else.

An alternative is to check if `getversion` call failed, and if so then ignore error and return version 0 from `getZKVersion`. But your suggestion looks more reliable to me.

PR #449, opened 2025-04-10
dbft: configure max blocks count for dbft_status RPC API

### LINKED ISSUE #102: Allow to configure max blocks count for `dbft_status` RPC API
# Rationale

Since #93 we have a `dbft_status` RPC API exposed with maximum number of blocks to take statistics from restricted by 1000 (`maxStatisticsPeriod` constant in the RPC dBFT API implementation). Ref. https://github.com/bane-labs/go-ethereum/pull/93#issuecomment-1897730374.

# Implementation

We need `maxStatisticsPeriod` to be customized via node config.

qianhh:
Close #102. @AnnaShaleva, please review.

### DIFF
--- a/consensus/dbft/api.go
+++ b/consensus/dbft/api.go
@@ -28,20 +28,28 @@ import (
 	"github.com/ethereum/go-ethereum/rpc"
 )
 
-const (
-	// defaultStatisticsPeriod is the default number of blocks to be taken into
+// StatisticsConfig is the configuration for dBFT statistics calculation.
+type StatisticsConfig struct {
+	// DefaultStatisticsPeriod is the default number of blocks to be taken into
 	// account during dBFT statistics calculation.
-	defaultStatisticsPeriod = uint64(64)
-	// maxStatisticsPeriod is the maximum number of blocks to be taken into
+	DefaultStatisticsPeriod uint64
+	// MaxStatisticsPeriod is the maximum number of blocks to be taken into
 	// account during dBFT statistics calculation.
-	maxStatisticsPeriod = uint64(1000)
-)
+	MaxStatisticsPeriod uint64
+}
+
+// DefaultStatistics is the default dBFT statistics calculation configuration.
+var DefaultStatistics = StatisticsConfig{
+	DefaultStatisticsPeriod: 64,
+	MaxStatisticsPeriod:     1000,
+}
 
 // API is a user facing RPC API to allow retrieve some information from the DBFT
 // consensus engine.
 type API struct {
-	chain consensus.ChainHeaderReader
-	bft   *DBFT
+	chain  consensus.ChainHeaderReader
+	bft    *DBFT
+	config StatisticsConfig
 }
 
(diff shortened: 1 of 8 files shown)


consensus/dbft/api.go
```diff
@@ -28,20 +28,27 @@ import (
 	"github.com/ethereum/go-ethereum/rpc"
 )
 
-const (
-	// defaultStatisticsPeriod is the default number of blocks to be taken into
+type StatisticsConfig struct {
```
AnnaShaleva, line 32: Exported type needs a comment.

consensus/dbft/dbft.go
```diff
@@ -236,22 +236,27 @@ type config struct {
...
 	}
 	if cfg.SecondsPerBlock == 0 {
 		return nil, errors.New("zero-period dBFT chain is not supported")
 	}
 	if cfg.Coinbase == (common.Address{}) {
 		return nil, errors.New("empty dBFT Coinbase is not allowed, need to specify mining rewards receiver")
 	}
+	if cfg.statisticsConfig.MaxStatisticsPeriod < cfg.statisticsConfig.DefaultStatisticsPeriod {
+		return nil, fmt.Errorf("max statistics period %d is less than default %d", cfg.statisticsConfig.MaxStatisticsPeriod, cfg.statisticsConfig.DefaultStatisticsPeriod)
+	}
```
AnnaShaleva: We also need to sanitize this config at the dBFT level: set the default values if not provided.

AnnaShaleva (commented):
Please, add `Close #102` to the commit message or PR description.

node/config.go
```diff
@@ -220,6 +221,9 @@ type Config struct {
 	EnablePersonal bool `toml:"-"`
 
 	DBEngine string `toml:",omitempty"`
+
+	// DBFTStatistics is the dBFT statistics calculation configuration.
+	DBFTStatistics dbft.StatisticsConfig `toml:",omitempty"`
```
AnnaShaleva, line 226: Is there any documentation that should be upgraded wrt this node config extension?

txhsl: There seems no particular document about the node configuration setup, we only provide an example about how to run a node.

Or is there any document about dBFT we need to update?

consensus/dbft/dbft.go
txhsl: https://github.com/bane-labs/go-ethereum/blob/a33e8443f6d0fc3965fdcf90065a18e591a64976/node/defaults.go#L77

AnnaShaleva: I saw this line. I'm talking about additional check at the dBFT constructor level: if MaxStatisticsPeriod oe DefaultStatisticsPeriod is not specified at the dBFT constructor, then set the default value.

node/config.go
AnnaShaleva: There's no such document on the dBFT side as far. I'm asking because I wasn't sure. So OK, there's just nothing to update.

txhsl (approved):

PR #451, opened 2025-04-14
crypto/tpke: consensus/dbft: fix signature computation

txhsl:
We need an independent hard fork to fix it, since we have deployed the code on T4.

Somehow ugly, yes.


txhsl:
#462.

PR #458, opened 2025-04-22
dbft: optimize CN index -> DKG index conversion

### LINKED ISSUE #357: Optimize CN index -> DKG index conversion
# Rationale

https://github.com/bane-labs/go-ethereum/blob/3b5367f2b3f6fba5e645f1aceba903d4f4e16f6a/consensus/dbft/dbft.go#L2226

This API introduced in #332 as a solution to #323 works fine but creates a lot of validator slice copies (this API is a frequently used code) and also performs a lot of sort operations. We need to optimize it since it's a hot-path.

# Implementation

Calculate CNIndex -> DKGIndex once per height, cache the mapping for a couple of latest blocks.

qianhh:
Close #357. 

I've optimised it for the requirements in #357, and we'll need to test it to see if it works or if there's room for improvement.

### DIFF
--- a/consensus/dbft/dbft.go
+++ b/consensus/dbft/dbft.go
@@ -26,6 +26,7 @@ import (
 	"fmt"
 	"io"
 	"math/big"
+	"sort"
 	"sync"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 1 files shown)


txhsl:
It works, and LGTM.

consensus/dbft/dbft.go
```diff
@@ -227,6 +228,9 @@ type DBFT struct {
 	dkgSnapshot  *Snapshot
 	txWatchList  []*TxWatchRetry
 	loopTaskChan chan *TxWatchList
+
+	// dkgIndexCache is a cache for storing the index array of the ordered validators
+	dkgIndexCache *lru.Cache[uint64, []int]
```
txhsl: I prefer to put this after `validatorsCache` definition, as well as its initialization, how do you think?

qianhh: Indeed, the two are related and make more sense when put together.

PR #461, opened 2025-04-24
dbft: fix sending data on closed channels

### LINKED ISSUE #452: `OnPayload` send on closed channel
#### System information

Geth version: `v0.3.2`

#### Expected behaviour

We always expect the geth code exits without panic.

#### Actual behaviour

@qianhh met a case that dBFT `OnPayload` tried to send a dBFT message on a closed channel from `eth/protocols/dbft`.

1. He used a wrong antimev keystore to start up a miner node;
2. After `handleDKG` returned an error about the DKG round mismatch, dBFT tried to exit automatically;
3. After dBFT engine exited, a dBFT message came.

#### Backtrace

Sorry, the text has gone, so only a screenshot.

Image

~~@AnnaShaleva It's not a normal use case, so I'm not sure if this should be categorized as a bug, also if we should deal with it inside dBFT or relate it to #405.~~

qianhh:
Close #452. 

I can't reproduce this bug now, except to change from `break events` to `Close` in the goroutine where the error occurs, and to listen to the exit signal of the `quit` channel when the channel sends data.

### DIFF
--- a/consensus/dbft/dkg.go
+++ b/consensus/dbft/dkg.go
@@ -174,7 +174,10 @@ func (c *DBFT) handleDKG(snapshot *Snapshot, keystore *antimev.KeyStore, h *type
 	if !suspended && currentHeight >= shareStartHeight && currentHeight < targetHeight {
 		// Send watch task list to loopTaskChan when handleDKG finished
 		defer func() {
-			c.loopTaskChan <- watchList
+			select {
+			case <-c.quit:
+			case c.loopTaskChan <- watchList:
+			}
 		}()
 	}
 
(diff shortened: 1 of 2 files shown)


qianhh:
Currently the process does not exit after `dbft` is `Close`ed off, is this the expected behaviour?

consensus/dbft/dkg.go
```diff
@@ -174,7 +174,10 @@ func (c *DBFT) handleDKG(snapshot *Snapshot, keystore *antimev.KeyStore, h *type
 	if !suspended && currentHeight >= shareStartHeight && currentHeight < targetHeight {
 		// Send watch task list to loopTaskChan when handleDKG finished
 		defer func() {
-			c.loopTaskChan <- watchList
+			select {
```
txhsl, line 177: The modifications about DKG channel handling may be conflict with #444.

I prefer leaving them aside and let's focus on the dBFT P2P messages.

AnnaShaleva (commented):
Unfortunately, I can't agree with the fix presented in this PR for two reasons:
1. `(*DBFT).Close` is not designated for internal usage. This method may be called only by the external consensus service manager and `go c.Close()` is not something I'd like to see in our code.
2. dBFT's eventLoop initially supposed to have the only possible reason to stop its operation: the call of `(*DBFT).Close`. But due to the fact that some callbacks may return error (`postBlock` and `waitForNewSealingProposal`), this statement was violated, and it's the root reason that should be fixed.

I refactored dBFT's eventLoop management in #467.

AnnaShaleva:
Replaced by #467.

Issue #453, opened 2025-04-15
Light client protocol


txhsl:
# Rationale

A light client protocol will be helpful if we are going to have a ZK-based cross-chain message protocol. I think this issue is now for discussion, since we haven't put it on the roadmap.

We always refer to [this article](https://medium.com/@ingonyama/bridging-the-multichain-universe-with-zero-knowledge-proofs-6157464fbc86) when talking about ZK messaging. It lists several projects about this topic, Succinct, Electron and zkBridge. They aim to build a trustless message protocol based on ZK, and the LC protocol is exactly the core of it.

When most of the other cross-chain protocols are using multi-sigs or centralized trust to guarantee the safety and correctness. These projects choose to rely on the trust of LC protocols. By proving a correct LC block verification from the sender chain, a continuos state change can be synchronized to the receiver chain.

IMO, we need at least a document specification of LC protocol, or an LC verifier like https://github.com/cometbft/tendermint-rs/blob/main/light-client-verifier/src/verifier.rs.

# Implementation

Geth has deleted its internal light client implement, ref https://github.com/ethereum/go-ethereum/pull/28586, because it was broken after Merge, the consensus was moved to CL. So we don't have such an implementation either.

Now for the Ethereum PoS, a light client protocol should be implemented in consensus layer clients, e.g. https://github.com/OffchainLabs/prysm/issues/12991, https://github.com/status-im/nimbus-eth2/issues/2337.

Although there are different LC client implementations, they stick to the same protocol definition, ref https://github.com/ethereum/annotated-spec/blob/master/altair/sync-protocol.md.

![](https://github.com/ethereum/annotated-spec/blob/master/altair/lightsync.png?raw=true)

This protocol is important for ZK messaging, because it determines what should an LC client verify when listening to an Ethereum full node. Then we can use ZK to prove this procedure and synchronize the state root to a smart contract on the receiver chain.

There is a simple practice about this synchronization, ref https://github.com/succinctlabs/sp1-helios or https://github.com/succinctlabs/sp1-tendermint-example.

Neo X now has a similar rule about block verification, and it's even much easier when we use the BLS.

1. Ethereum CL needs 512 public keys to compute the global public key for verification, and needs a Merkel hash to identify the block miner;
2. Neo X has an on-chain DKG to maintain the global public key for verification, CNs share the same secret, so there is no need for public key aggregation, and no need to verify the block miner;
3. Ethereum CL doesn't verify the election of sync committee;
[...]

PR #462, opened 2025-04-27
crypto/tpke: consensus/dbft: fix signature computation and remove support for ECDSA sigs

txhsl:
Conflict #451. Relate #453.


txhsl:
Seems going worse with a lot of redundant codes, closing.

PR #463, opened 2025-04-28
crypto/tpke: consensus/dbft: fix signature computation

### LINKED ISSUE #450: TPKE is using the negative signature
#### System information

Geth version: `v0.3.2`

#### Expected behaviour

After implementing the `crypto/tpke` library, we can use BLS signatures in Neo X. Although secret keys are generated through on-chain DKG, the scheme itself is expected to equal Ethereum CL.

#### Actual behaviour

I'm verifying the computation result of https://github.com/ChainSafe/bls and our `crypto/tpke`, for a possible ZK bridge in the future, and I notice the result is different.

In short word, the algorithm itself is correct, but a `Neg` is applied on the signature, not in verification. This bug was imported when I was using `github.com/consensys/gnark-crypto/ecc/bls12-381` to replace `github.com/kilic/bls12-381`, there was a difference in the paring options.

There is a similar Go implementation by https://github.com/Consensys, ref https://github.com/Consensys/gnark-crypto/blob/feat/bls-sig/ecc/bls12-381/signature/bls/bls.go#L205.

The fix itself is simple, ref #451.

txhsl:
Close #450. Reopened #451.

### DIFF
--- a/consensus/dbft/dbft.go
+++ b/consensus/dbft/dbft.go
@@ -738,7 +738,7 @@ func (c *DBFT) verifyCommitCb(p dbft.ConsensusPayload[common.Hash]) error {
 	switch v {
 	case dbftutil.ExtraV0:
 		expectedLen = crypto.SignatureLength
-	case dbftutil.ExtraV1:
+	case dbftutil.ExtraV1, dbftutil.ExtraV2:
 		if c.config.enforceECDSASignatures || (!isAMEV && c.chain.Config().IsNeoXAMEV(new(big.Int).Add(h, bigOne))) {
 			expectedLen = crypto.SignatureLength
 		} else {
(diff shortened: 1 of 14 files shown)


consensus/dbft/dbft.go
```diff
@@ -1379,13 +1379,19 @@ func (c *DBFT) getBlockWitness(pub *tpke.PublicKey, block *Block) ([]byte, error
...
 			}
 		}
 		msg := dbftRLP(block.header)
 		c.lock.RLock()
 		ks := c.amevKeystore
 		c.lock.RUnlock()
 		sig, err := ks.AggregateAndVerifySig(msg, shares)
+		if v == dbftutil.ExtraV1 {
+			sig.Neg()
+		}
```
AnnaShaleva: I'd like to abstract dBFT from math as much as possible. What do you think about:
```suggestion
		sig, err := ks.AggregateAndVerifySig(msg, shares, v == dbftutil.ExtraV1)
```

and then just something like:
```
func (ks *KeyStore) AggregateAndVerifySig(msg []byte, inputs map[int]*tpke.SignatureShare, negateResult bool) (*tpke.Signature, error) {
```

consensus/dbft/dbftutil/util.go
```diff
@@ -23,6 +23,9 @@ const (
 	// ExtraV1 is the 1-st version of block's Extra. Extra of this version includes global
 	// TPKE public key followed by aggregated validators' threshold signature.
 	ExtraV1 ExtraVersion = 0x01
+	// ExtraV1Fix is the fix of the 1-st version of block's Extra. Extra of this version
+	// includes global TPKE public key followed by a fixed V1 threshold signature.
+	ExtraV1Fix ExtraVersion = 0x02
```
AnnaShaleva: Just use `ExtraV2` everywhere, because it's the second version of Extra in its essence.

consensus/dbft/dbft.go
AnnaShaleva: And the same approach may be applied to `SignShare`.

AnnaShaleva (commented):
Agree, hardfork is inevitable.

consensus/dbft/dbft.go
txhsl: Good and possible.

txhsl:
@songb2 Need a verification on the compatibility with current running dBFT ExtraV1, including both synchronization and participation.

songb2:
No issues found.

PR #464, opened 2025-05-06
dbft: optimize mempool-based proposed transactions verification

### LINKED ISSUE #341: Optimize mempool-based proposed transactions verification
# Rationale

#339 utilizes local LegacyPool to verify proposed in-block transactions. Every time clear mempool is created. It may be inefficient.

# Implementation

Allocate local LegacyPool once, bind it to `*DBFT`, reuse it for in-block transactions verification and clear it every time after subsequent verification. The caveat is mempool reset, because Geth's mempool is not designed for multiple resets at the same height.

We need to benchmark the propose solution to see if it works faster than creation of mempool every time when it's needed.

qianhh:
Close #341.

After the discussion in #341, I implemented a SimpleLegacyPool specifically for verifying transactions and removed other unnecessary functions.

### DIFF
--- a/consensus/dbft/dbft.go
+++ b/consensus/dbft/dbft.go
@@ -1237,8 +1237,8 @@ func (c *DBFT) processPreBlockCb(b dbft.PreBlock[common.Hash]) error {
 
 // newLocalPool returns an initialized instance of LegacyPool with default config
 // except that locals are prohibited and journal is not stored.
-func (c *DBFT) newLocalPool(parent *types.Header) *legacypool.LegacyPool {
-	p := legacypool.New(legacypool.Config{
+func (c *DBFT) newLocalPool(parent *types.Header) *legacypool.SimpleLegacyPool {
+	p := legacypool.NewSimple(legacypool.Config{
 		Locals:                  nil,
 		NoLocals:                true,
 		Journal:                 "",
--- a/core/txpool/legacypool/simplelegacypool.go
+++ b/core/txpool/legacypool/simplelegacypool.go
@@ -1,0 +1,865 @@
+package legacypool
+
+import (
+	"fmt"
+	"math"
+	"math/big"
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"github.com/ethereum/go-ethereum/antimev"
+	"github.com/ethereum/go-ethereum/common"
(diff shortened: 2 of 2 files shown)


txhsl (commented):
There is no need for these copy-pastes, and meanwhile this is not the simplest implementation to meet our requirements.

I prefer only adding simplified `NewSimple()` and `InitSimple()` to `LegacyPool` to share most of its logic, e.g. https://github.com/bane-labs/go-ethereum/commit/ed60f3a44fa134ce429c46184e6ad3da44dee40a.

qianhh:
> There is no need for these copy-pastes, and meanwhile this is not the simplest implementation to meet our requirements.
> 
> I prefer only adding simplified `NewSimple()` and `InitSimple()` to `LegacyPool` to share most of its logic, e.g. [ed60f3a](https://github.com/bane-labs/go-ethereum/commit/ed60f3a44fa134ce429c46184e6ad3da44dee40a).

Putting them together is a good approach, but in that case, the Add method will be blocked when it is called. Therefore, it is necessary to write another function to remove this part of the code and replace the Add method.
https://github.com/bane-labs/go-ethereum/blob/22ba65d6234d06ee8818e470e8c78ae28e8af2bc/core/txpool/legacypool/legacypool.go#L1188-L1192

txhsl:
> Therefore, it is necessary to write another function to remove this part of the code and replace the Add method.
> 
> https://github.com/bane-labs/go-ethereum/blob/22ba65d6234d06ee8818e470e8c78ae28e8af2bc/core/txpool/legacypool/legacypool.go#L1188-L1192

Yep, but if we check `runReorg()` after this calling, I think no reset really happens after this request. Only `promoteExecutables()` is executed, as well as truncate and broadcast. Neither of them is necessary in our check logic.

We can either change `add()` itself or change the condition check in `requestPromoteExecutables()`.

Ref #466.

AnnaShaleva (commented):
I'm joining to https://github.com/bane-labs/go-ethereum/pull/464#pullrequestreview-2820164035, if there's a chance, we need to reuse existing code as much as possible.

I presume this PR may be closed in favour of #466 since I see the same logic is implemented in #466 but with less changes.

PR #466, opened 2025-05-07
core/txpool: consensus/dbft: use static legacy pool instance in dbft

### LINKED ISSUE #341: Optimize mempool-based proposed transactions verification

txhsl:
Close #341. Replace #464. Ref https://github.com/bane-labs/go-ethereum/pull/464#pullrequestreview-2820164035.

Since we only add new transactions to an empty `LegacyPool` instance in dBFT verification and check its errors, this PR simplify the pool construction and initialization, and remove unused maintenance loops.

### DIFF
--- a/consensus/dbft/dbft.go
+++ b/consensus/dbft/dbft.go
@@ -223,6 +223,10 @@ type DBFT struct {
 	validatorsCache *lru.Cache[uint64, []common.Address]
 	// dkgIndexCache is a cache for storing the index array of the ordered validators
 	dkgIndexCache *lru.Cache[uint64, []int]
+	// staticPool is a legacy pool instance for decrypted transaction verification,
+	// which is initialized once per height at postBlock callback. It should be reset
+	// before any reusage at the same height.
+	staticPool *legacypool.LegacyPool
 
 	// The fields below are for testing only
 	fakeDiff bool // Skip difficulty verifications
@@ -910,9 +914,11 @@ func (c *DBFT) verifyPreBlockCb(b dbft.PreBlock[common.Hash]) bool {
 	}
 	ethBlock := dbftBlock.ToEthBlock()
 
-	localPool := c.newLocalPool(parent)
-	defer localPool.CloseSilently()
-	errs := localPool.Add(dbftBlock.transactions, false, false)
+	// If is the first view of the block, then initialize static pool with fresh parent.
+	if c.dbft.Context.ViewNumber == 0 {
+		c.initStaticPool(parent)
+	}
+	errs := c.staticPool.Add(dbftBlock.transactions, false, false)
 	for i, err := range errs {
 		if err != nil {
 			log.Warn("proposed PreBlock has invalid transaction",
@@ -923,6 +929,7 @@ func (c *DBFT) verifyPreBlockCb(b dbft.PreBlock[common.Hash]) bool {
 			return false
 		}
 	}
+	c.staticPool.ResetStatic()
 
 	state, receipts, gasUsed, err := c.chain.VerifyBlock(ethBlock, true)
 	if err != nil {
@@ -1122,7 +1129,6 @@ func (c *DBFT) processPreBlockCb(b dbft.PreBlock[common.Hash]) error {
 			j                  int
 			txx                = make([]*types.Transaction, len(pre.transactions))
 			parent             = c.chain.GetHeader(pre.header.ParentHash, pre.header.Number.Uint64()-1)
-			localPool          = c.newLocalPool(parent)
 			fallbackToEnvelope = func(i int, incrementJ bool, reason string) bool {
 				if len(reason) != 0 {
 					log.Info("Falling back to envelope",
@@ -1130,7 +1136,7 @@ func (c *DBFT) processPreBlockCb(b dbft.PreBlock[common.Hash]) error {
 						"envelope index", i,
 						"reason", reason)
 				}
-				errs := localPool.Add([]*types.Transaction{pre.transactions[i]}, false, false)
+				errs := c.staticPool.Add([]*types.Transaction{pre.transactions[i]}, false, false)
 				if errs[0] != nil {
 					log.Info("Falling back to original set of transactions",
 						"envelope hash", pre.transactions[i].Hash(),
@@ -1147,7 +1153,6 @@ func (c *DBFT) processPreBlockCb(b dbft.PreBlock[common.Hash]) error {
 				return true
 			}
 		)
-		defer localPool.CloseSilently()
 
 		// If we're backup, then pre.finalReceipts are filled in during PreBlock verification;
 		// if we're primary, then reuse receipts got after PrepareRequest construction since
@@ -1156,7 +1161,8 @@ func (c *DBFT) processPreBlockCb(b dbft.PreBlock[common.Hash]) error {
 		if ctx.IsPrimary() && pre.finalState == nil {
 			receipts = c.sealingReceipts
 		}
-
+		// No need to reinitialize static pool since it was already initialized in verifyPreBlockCb.
+		// Reset the static pool after processing all transactions.
 		for i := range pre.transactions {
 			var isEnvelope = j < len(pre.envelopesData) && pre.envelopesData[j].index == i
 			if !isEnvelope || // pre.transactions[i] is not an envelope, use it as-is.
@@ -1192,7 +1198,7 @@ func (c *DBFT) processPreBlockCb(b dbft.PreBlock[common.Hash]) error {
 					break
 				}
 			}
-			errs := localPool.Add([]*types.Transaction{decryptedTx}, false, false)
+			errs := c.staticPool.Add([]*types.Transaction{decryptedTx}, false, false)
 			if errs[0] != nil {
 				if fallbackToEnvelope(i, true, fmt.Sprintf("decrypted transaction has pool conflicts: %s", errs[0].Error())) {
 					continue
@@ -1205,6 +1211,7 @@ func (c *DBFT) processPreBlockCb(b dbft.PreBlock[common.Hash]) error {
 			hasDecryptedTxs = true
 		}
 		pre.finalTransactions = txx
+		c.staticPool.ResetStatic()
 	}
 
 	// Use cached processing results if no new transactions were included into
@@ -1250,10 +1257,10 @@ func (c *DBFT) processPreBlockCb(b dbft.PreBlock[common.Hash]) error {
 	return nil
 }
 
-// newLocalPool returns an initialized instance of LegacyPool with default config
+// newStaticPool returns an initialized instance of LegacyPool with default config
 // except that locals are prohibited and journal is not stored.
-func (c *DBFT) newLocalPool(parent *types.Header) *legacypool.LegacyPool {
-	p := legacypool.New(legacypool.Config{
+func newStaticPool(chain ChainHeaderReader) *legacypool.LegacyPool {
+	return legacypool.NewStatic(legacypool.Config{
 		Locals:                  nil,
 		NoLocals:                true,
 		Journal:                 "",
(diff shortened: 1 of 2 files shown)


core/txpool/legacypool/legacypool.go
```diff
@@ -309,6 +309,32 @@ func New(config Config, chain BlockChain) *LegacyPool {
...
+		beats:       make(map[common.Address]time.Time),
+		all:         newLookup(),
+	}
+	pool.locals = newAccountSet(pool.signer)
+	for _, addr := range config.Locals {
+		log.Info("Setting new local account", "address", addr)
+		pool.locals.add(addr)
+	}
+	pool.priced = newPricedList(pool.all)
+	return pool
```
AnnaShaleva, line 308: `config.Journal` is empty in the dBFT's call to `New`, so removing journal initialization code from `NewSimple` doesn't help a lot.

core/txpool/legacypool/legacypool.go
```diff
@@ -309,6 +309,32 @@ func New(config Config, chain BlockChain) *LegacyPool {
...
+	pool := &LegacyPool{
+		config:      config,
+		chain:       chain,
+		chainconfig: chain.Config(),
+		signer:      types.LatestSigner(chain.Config()),
+		pending:     make(map[common.Address]*list),
+		cached:      make(map[common.Address]*list),
+		queue:       make(map[common.Address]*list),
+		beats:       make(map[common.Address]time.Time),
+		all:         newLookup(),
```
AnnaShaleva: `NewSimple` constructor differs from `New` in that it doesn't allocate channels in fact. `NewSimple` should be reused from `New`, and it sounds to me more like `NewStatic` (and `InitStatic`) or something like that.

core/txpool/legacypool/legacypool.go
```diff
@@ -364,6 +390,33 @@ func (pool *LegacyPool) Init(gasTip uint64, head *types.Header, reserve txpool.A
 	return nil
 }
 
+// InitSimple sets the gas price needed to keep a transaction in the pool and
+// the chain head to allow balance / nonce checks. This method doesn't start
+// loops.
+func (pool *LegacyPool) InitSimple(gasTip uint64, head *types.Header, reserve txpool.AddressReserver) error {
+	// Set the address reserver to request exclusive access to pooled accounts
+	pool.reserve = reserve
```
AnnaShaleva, line 354: Please, don't copy-paste things, it complicates further code maintenance. If you're moving some part of the code to a separate method, then try to reuse it as much as possible from the original code.

core/txpool/legacypool/legacypool.go
```diff
@@ -1232,6 +1232,10 @@ func (pool *LegacyPool) Add(txs []*types.Transaction, local, sync bool) []error
 		errs[nilSlot] = err
 		nilSlot++
 	}
+	// Return if no channel is initialized, ref NewSimple and InitSimple
+	if pool.reqPromoteCh == nil {
```
AnnaShaleva, line 1219: It should be a part of the previous commit, otherwise previous commit contains bug.

AnnaShaleva (commented):
> Close https://github.com/bane-labs/go-ethereum/issues/341

It doesn't, although it makes mempool structure more light and mempool initialization process faster. As you can see, we allocate mempool twice per block: during PrepareRequest verification and during PreBlock processing (if there are decrypted transactions). So the intention of #341 is to allocate mempool once during PrepareRequest verification and then reuse the same pool for further PreBlock processing (with preliminarily reset of the existing pool). The idea is to request the DB's state snapshot required for mempool initialization once per block only.

So we either need further adjustments and benchmarks for this PR (if you'd like to close #341) or I agree to merge this PR as a patch without closing #341 because this PR in fact simplifies local pool management.

I also need results of testing from @songb2: we need to setup privnet and try to load it with simple and encrypted transactions to see if mempool verification still works as expected and there's no unexpected interaction with nil channels.

txhsl:
> It doesn't, although it makes mempool structure more light and mempool initialization process faster. As you can see, we allocate mempool twice per block: during PrepareRequest verification and during PreBlock processing (if there are decrypted transactions). So the intention of #341 is to allocate mempool once during PrepareRequest verification and then reuse the same pool for further PreBlock processing (with preliminarily reset of the existing pool).

Seems like we misunderstood the proposal. I will check.

I think we can update this PR to close #341, the idea is not conflict.

core/txpool/legacypool/legacypool.go
txhsl: But in case the following codes are called unexpectedly, I think we'd better ensure it's empty here.

https://github.com/bane-labs/go-ethereum/blob/0b3f82e15b19a70df2631e574c53dc4e4468b474/core/txpool/legacypool/legacypool.go#L989

https://github.com/bane-labs/go-ethereum/blob/0b3f82e15b19a70df2631e574c53dc4e4468b474/core/txpool/legacypool/legacypool.go#L1011

https://github.com/bane-labs/go-ethereum/blob/0b3f82e15b19a70df2631e574c53dc4e4468b474/core/txpool/legacypool/legacypool.go#L1099-L1107

txhsl:
> The idea is to request the DB's state snapshot required for mempool initialization once per block only. So we either need further adjustments and benchmarks for this PR (if you'd like to close https://github.com/bane-labs/go-ethereum/issues/341) or I agree to merge this PR as a patch without closing https://github.com/bane-labs/go-ethereum/issues/341 because this PR in fact simplifies local pool management.

@AnnaShaleva I think I've implemented the code required by that issue, and the privnet works with these modifications. Please check if it is the thing you may expect, and then we will test it.

consensus/dbft/dbft.go
```diff
@@ -895,9 +897,8 @@ func (c *DBFT) verifyPreBlockCb(b dbft.PreBlock[common.Hash]) bool {
 	}
 	ethBlock := dbftBlock.ToEthBlock()
 
-	localPool := c.newLocalPool(parent)
-	defer localPool.CloseSilently()
-	errs := localPool.Add(dbftBlock.transactions, false, false)
+	c.staticPool = newStaticPool(parent, c.chain)
```
qianhh: Here I have some doubts: Since the new instance is only created in `verifyPreBlockCb` and then directly used in `processPreBlockCb`, does it always execute `verifyPreBlockCb` first and then `processPreBlockCb` each time? Also, after the verification here, there is no reset. Maybe in `processPreBlockCb`, when verifying, there might be an error due to the existence of the same transaction?

txhsl: > does it always execute verifyPreBlockCb first and then processPreBlockCb each time?

I think yes? The reset is before the second usage, maybe I move it here after the first usage as an optimization.

Moved. There was a reset, but in another place.

vncoelho:
> PreBlock

I think the PreBlock processing really needs a different pool

songb2:
I tested the pr compared with main by sending 10000 txs in batch. The performance is improved, main took 663s to handle while this change just took 430s.
main
fea80ac84ce84422978cce9aa0fb868
this pr
fd0872aca134eb5d5c78a73a2906d2d

txhsl:
> I tested the pr compared with main by sending 10000 txs in batch. The performance is improved, main took 663s to handle while this change just took 430s.

We may have some miscounted times in this table, but I can see performance gets improved.

core/txpool/legacypool/legacypool.go
```diff
@@ -381,6 +381,23 @@ func (pool *LegacyPool) InitStatic(gasTip uint64, head *types.Header, reserve tx
...
+	pool.cached = make(map[common.Address]*list)
+	pool.queue = make(map[common.Address]*list)
+	pool.beats = make(map[common.Address]time.Time)
+	pool.all = newLookup()
+	pool.locals = newAccountSet(pool.signer)
+	for _, addr := range pool.config.Locals {
+		log.Info("Setting new local account", "address", addr)
+		pool.locals.add(addr)
+	}
+	pool.priced = newPricedList(pool.all)
```
AnnaShaleva: Please, move this code to a helper and reuse it from both `ResetStatic` and `NewStatic`.

consensus/dbft/dbft.go
```diff
@@ -907,6 +909,7 @@ func (c *DBFT) verifyPreBlockCb(b dbft.PreBlock[common.Hash]) bool {
 			return false
 		}
 	}
+	c.staticPool.ResetStatic()
```
AnnaShaleva, line 932: Don't need to reset here because there's a high chance that there's no envelopes in the block and pool won't be reused anymore at this height. Instead, we need to reset the pool right before decrypted envelopes processing (if they exist in this block), see the comment below.

consensus/dbft/dbft.go
```diff
@@ -1106,15 +1109,14 @@ func (c *DBFT) processPreBlockCb(b dbft.PreBlock[common.Hash]) error {
 			j                  int
 			txx                = make([]*types.Transaction, len(pre.transactions))
 			parent             = c.chain.GetHeader(pre.header.ParentHash, pre.header.Number.Uint64()-1)
-			staticPool         = c.newStaticPool(parent)
```
AnnaShaleva: Reset the pool right here.

consensus/dbft/dbft.go
```diff
@@ -1188,6 +1190,7 @@ func (c *DBFT) processPreBlockCb(b dbft.PreBlock[common.Hash]) error {
 			hasDecryptedTxs = true
 		}
 		pre.finalTransactions = txx
+		c.staticPool.ResetStatic()
```
AnnaShaleva, line 1214: And then don't need to reset the pool here, because PreBlock is processed once per view.

consensus/dbft/dbft.go
```diff
@@ -222,6 +222,8 @@ type DBFT struct {
 	validatorsCache *lru.Cache[uint64, []common.Address]
 	// dkgIndexCache is a cache for storing the index array of the ordered validators
 	dkgIndexCache *lru.Cache[uint64, []int]
+	// staticPool is a legacy pool instance for decrypted transaction verification
+	staticPool *legacypool.LegacyPool
```
AnnaShaleva, line 229: We need to add a comment that this pool is initialized once per height at postBlock callback, so that users of this field should reset it before the usage if needed.

consensus/dbft/dbft.go
```diff
@@ -895,8 +897,8 @@ func (c *DBFT) verifyPreBlockCb(b dbft.PreBlock[common.Hash]) bool {
 	}
 	ethBlock := dbftBlock.ToEthBlock()
 
-	staticPool := c.newStaticPool(parent)
-	errs := staticPool.Add(dbftBlock.transactions, false, false)
+	c.staticPool = newStaticPool(parent, c.chain)
```
AnnaShaleva: Move this initialization to `postBlock` callback prior to `handleDKG`, chain state at the current height should already be available by this moment. And here we need to reset only if CV happened:
```suggestion
	if c.dbft.Context.ViewNumber > 0 {
		c.staticPool.ResetStatic()
	}
```

The overall idea is to request state snapshot once per block.

AnnaShaleva (commented):
Good, now the second commit actually fixes #341, we just need to optimize it a little bit more in order not to spend time on useless resets.

consensus/dbft/dbft.go
txhsl: > Move this initialization to `postBlock` callback prior to `handleDKG`, chain state at the current height should already be available by this moment.

It looks a bit confusing to me, since we don't need the current state but the parent state to verify transactions.

If we are going to judge based on view number, then we can move the initialization into the `else` case here.

> The overall idea is to request state snapshot once per block.

OK, then let's only init it when `ViewNumber` is 0.

consensus/dbft/dbft.go
txhsl: Well, implemented, but there is a potential issue. If we have the static pool not reset immediately after `verifyPreBlockCb` and `processPreBlockCb`, the transactions in preblock we add to the pool will be left there until the next block comes.

The pool size won't be very huge, up to the block limit, but it will be there.

txhsl: Reverted, the first implementation seems better to me, since I think spending some time to remove transactions in static pool is helpful in better memory management.

Let's
1. Create the pool instance with `chain` when starting dBFT engine;
2. Initialize the pool with the latest `parent` in `verifyPreBlockCb` when it is the first view of a new block;
3. Reset (clear all transactions) after every usage of this pool.

PR #467, opened 2025-05-13
dbft: refactor shutdown process

### LINKED ISSUE #452: `OnPayload` send on closed channel

AnnaShaleva:
Refactor dBFT eventloop and DKG task dispatcher shutdown.

Close #452, a replacement of #461.

### DIFF
--- a/consensus/dbft/dbft.go
+++ b/consensus/dbft/dbft.go
@@ -133,6 +133,9 @@ var (
 	errUnauthorizedSigner = errors.New("unauthorized signer")
 )
 
+// errShutdown represents an error caused by engine shutdown initiated by user.
+var errShutdown = errors.New("shutdown detected")
+
 // DBFT is the proof-of-authority consensus engine.
 type DBFT struct {
 	messages chan Payload
@@ -2077,10 +2080,12 @@ func (c *DBFT) Start(chain ChainHeaderWriter) {
 		// initialised by miner. Start consensus once we have new sealing work in Seal.
 		err := c.waitForNewSealingProposal(c.lastIndex+1, false)
 		if err != nil {
-			log.Warn("Failed to fetch latest sealing proposal",
+			if errors.Is(err, errShutdown) {
+				return
+			}
+			log.Crit("Failed to fetch latest sealing proposal",
 				"index", c.lastIndex+1,
 				"err", err.Error())
-			return
 		}
 
 		// Manually initialize static pool and DKG snapshot in case if waitForNewSealingProposal didn't call
@@ -2153,7 +2158,7 @@ func (c *DBFT) waitForNewSealingProposal(desiredHeight uint64, updateContext boo
 		}
 		select {
 		case <-c.quit:
-			return errors.New("shutdown detected")
+			return errShutdown
 		default:
 		}
 		time.Sleep(time.Second)
@@ -2228,11 +2233,6 @@ func (c *DBFT) eventLoop() {
 	// prevent a major security vuln where external parties can DOS you with blocks
 	// and halt your dBFT operation for as long as the DOS continues.
 	downloaderEvents := c.mux.Subscribe(downloader.DoneEvent{}, downloader.FailedEvent{})
-	defer func() {
-		if !downloaderEvents.Closed() {
-			downloaderEvents.Unsubscribe()
-		}
-	}()
 	dlEventCh := downloaderEvents.Chan()
 
 events:
@@ -2241,9 +2241,6 @@ events:
 		select {
 		case <-c.quit:
 			log.Info("shutting down dBFT event loop")
-			c.dbft.Timer.Stop()
-
-			c.chainHeadSub.Unsubscribe()
 			break events
 		case <-c.dbft.Timer.C():
 			h, v := c.dbft.Timer.Height(), c.dbft.Timer.View()
@@ -2279,13 +2276,7 @@ events:
 		case tx := <-c.txs:
 			c.dbft.OnTransaction(&Transaction{Tx: tx})
 		case b := <-c.chainHeadEvents:
-			err := c.handleChainBlock(b.Block.Header(), true)
-			if err != nil {
-				log.Warn("Failed to handle chain block",
-					"index", b.Block.NumberU64(),
-					"err", err.Error())
-				break events
-			}
+			c.handleChainBlock(b.Block.Header(), true)
 		case err := <-c.chainHeadSub.Err():
 			// System has stopped.
 			log.Info("Stopping dBFT service since block subscriptions are stopped")
@@ -2303,26 +2294,13 @@ events:
 			switch ev.Data.(type) {
 			case downloader.FailedEvent:
 				latest := c.chain.CurrentHeader()
-				err := c.handleChainBlock(latest, false)
-				if err != nil {
-					log.Warn("Failed to handle latest chain block",
-						"index", latest.Number.Uint64(),
-						"err", err.Error())
-					break events
-				}
-
+				c.handleChainBlock(latest, false)
 			case downloader.DoneEvent:
 				// Stop reacting to downloader events.
 				downloaderEvents.Unsubscribe()
 
 				latest := c.chain.CurrentHeader()
-				err := c.handleChainBlock(latest, false)
-				if err != nil {
-					log.Warn("Failed to handle latest chain block",
-						"index", latest.Number.Uint64(),
-						"err", err.Error())
-					break events
-				}
+				c.handleChainBlock(latest, false)
 			}
 		}
 		// Always process block event if there is any, we can add one above or external
@@ -2337,13 +2315,7 @@ events:
 			}
 		}
 		if latestBlock.Block != nil {
-			err := c.handleChainBlock(latestBlock.Block.Header(), true)
-			if err != nil {
-				log.Warn("Failed to handle latest chain block",
-					"index", latestBlock.Block.NumberU64(),
-					"err", err.Error())
-				break events
-			}
+			c.handleChainBlock(latestBlock.Block.Header(), true)
 		}
 		newView := c.dbft.ViewNumber
 		// If ChangeView has happened, we always need to wait for the new proposal
@@ -2352,16 +2324,23 @@ events:
 			log.Info("Change view detected, waiting for new sealing task to be submitted by miner", "old view", oldView, "new view", newView)
 			err := c.waitForNewSealingProposal(uint64(c.dbft.Context.BlockIndex), true)
 			if err != nil {
-				log.Warn("Failed to fetch latest sealing proposal",
+				if errors.Is(err, errShutdown) {
+					break events
+				}
+				log.Crit("Failed to fetch latest sealing proposal",
 					"index", c.dbft.Context.BlockIndex,
 					"err", err.Error())
-				break events
 			}
 			log.Info("Start dBFT process for updated sealing work",
 				"index", c.dbft.Context.BlockIndex,
 				"view", newView,
 			)
 		}
+	}
+	c.dbft.Timer.Stop()
+	c.chainHeadSub.Unsubscribe()
+	if !downloaderEvents.Closed() {
+		downloaderEvents.Unsubscribe()
 	}
 drainLoop:
 	for {
@@ -2500,7 +2479,7 @@ func (c *DBFT) newConsensusPayloadCb(ctx *dbft.Context[common.Hash], t dbft.Mess
 	return cp
 }
 
-func (c *DBFT) handleChainBlock(h *types.Header, checkForSync bool) error {
+func (c *DBFT) handleChainBlock(h *types.Header, checkForSync bool) {
 	// A short path if miner is not active and the node is in the process of block
 	// sync. In this case dBFT can't react properly on the newcoming blocks since no
 	// sealing task is expected from miner.
@@ -2509,7 +2488,7 @@ func (c *DBFT) handleChainBlock(h *types.Header, checkForSync bool) error {
 			"block index", h.Number.Int64(),
 			"dbft index", c.dbft.BlockIndex,
 		)
-		return nil
+		return
 	}
 
 	// We can get our own block here, so check for index.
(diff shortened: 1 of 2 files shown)


AnnaShaleva:
@txhsl, I see that this PR will conflict with #444 due to the last commit, so we may either finish and merge #444 firstly and then I'll rebase this PR or you may check the last commit and implement its idea in #444 PR. Either way is OK for me, but we should fix the DKG task dispatcher management because right now there's a bug.

@qianhh, @songb2 you're welcomed to review and test.

txhsl:
> Either way is OK for me, but we should fix the DKG task dispatcher management because right now there's a bug.

I prefer that we finish #444 first.

AnnaShaleva:
Ready for review&testing.

AnnaShaleva:
OK, there's one inaccuracy in dispatcher stop process, will fix:
```
INFO [05-27|20:20:13.214] New block in the chain                   "dbft index"=5 "chain index"=5 hash=0x1d777e5beab665dc2ab0d24aba864096535367415eabb4682858b96522fc137a "parent hash"=0x9a2bd21eec6c7e38645451a4337c66008f316d2c0a631e4b744174b85063105b primary=5 coinbase=0x1212000000000000000000000000000000000003 "mix digest"=0x3e48bc8f2eb502267b1eae6bded2ece31a29829cdf94afb44f37688702643e43
INFO [05-27|20:20:13.214] Fetching latest sealing proposal         "desired number"=6
INFO [05-27|20:20:13.214] DKG task dispatcher stopped
panic: send on closed channel

goroutine 4380 [running]:
github.com/ethereum/go-ethereum/consensus/dbft.(*DBFT).handleDKG(0xc000798908, 0xc000f33020, 0xc0002a6000, 0xc0126ed908, 0x0, 0x0)
	github.com/ethereum/go-ethereum/consensus/dbft/dkg.go:343 +0x153c
github.com/ethereum/go-ethereum/consensus/dbft.(*DBFT).waitForNewSealingProposal(0xc000798908, 0x6, 0x0)
	github.com/ethereum/go-ethereum/consensus/dbft/dbft.go:2158 +0x4c7
github.com/ethereum/go-ethereum/consensus/dbft.(*DBFT).handleChainBlock(0xc000798908, 0xc01293cf08, 0x90?)
	github.com/ethereum/go-ethereum/consensus/dbft/dbft.go:2491 +0x7d5
github.com/ethereum/go-ethereum/consensus/dbft.(*DBFT).eventLoop(0xc000798908)
	github.com/ethereum/go-ethereum/consensus/dbft/dbft.go:2305 +0xb65
created by github.com/ethereum/go-ethereum/consensus/dbft.(*DBFT).Start in goroutine 4373
	github.com/ethereum/go-ethereum/consensus/dbft/dbft.go:2093 +0x5a9
```

AnnaShaleva:
Fixed, ready for review.

consensus/dbft/dbft.go
```diff
@@ -2494,7 +2474,6 @@ func (c *DBFT) handleChainBlock(h *types.Header, checkForSync bool) error {
 			"block index", h.Number.Int64(),
 			"dbft index", c.dbft.BlockIndex,
 		)
-		return nil
```
txhsl, line 2512: This makes it no longer a short path here.

If we are sure that the following codes will not be triggered anyway, then it's OK, but I think they will.

Fixed, but leave this for comments, maybe I'm wrong.

consensus/dbft/dkg.go
```diff
@@ -340,275 +340,315 @@ func (c *DBFT) handleDKG(snapshot *snapshot, keystore *antimev.KeyStore, h *type
...
-			select {
-			case c.executeProofTaskChan <- &taskList:
-			default:
-				// Give up the tasks since the channel is going to block DKG and dBFT.
-				return fmt.Errorf("proof task channel is full")
-			}
+		select {
+		case <-c.quit:
+			return nil
+		case c.dkgTaskExecutorCh <- &taskList: // send even if taskList is empty since dkgTaskExecutorCh needs a heartbeat.
```
txhsl: Also related to the line https://github.com/bane-labs/go-ethereum/pull/467/files#diff-c80584fa2bebbf3162e0d5141bf4b091712fdba0776be26333fe2a0f72249f7eR556, I think we need the heartbeat sent from `handleDKG()` instead of inside `dkgTaskExecutor()`.

This is because `dkgTaskExecutor()` maybe time-consuming, but we require the `dkgTaskWatcher()` to be executed once per new height. So this change will delay the regular loop of `retryList` check.

It looks OK before we enable ZK generation, but after that, the `dkgTaskExecutor()` will become busy and empty `taskList` will be stuck in the `dkgTaskExecutorCh` and trigger the `proof task channel is full` error.

```
CRIT [06-03|11:36:29.401] Failed to fetch latest sealing proposal  index=12 err="failed to initialize DKG snapshot at height 11: proof task channel is full"
```

Update: Also encountered a `send on closed channel` panic, I'm going to refactor this part.

consensus/dbft/dkg.go
```diff
@@ -340,275 +340,315 @@ func (c *DBFT) handleDKG(snapshot *snapshot, keystore *antimev.KeyStore, h *type
...
-			default:
-				panic(fmt.Errorf("unknown DKG method to prove: %s", task.Method))
+				// Release the cache when the task list is fully handled.
+				sevenMsgR1CS = nil
+				sevenMsgPK = nil
+			}
+			select {
+			case <-c.quit:
+				break executeLoop
+			case c.dkgTaskWatcherCh <- watchList: // send even if taskList is empty since dkgTaskWatcherCh needs a heartbeat.
```
txhsl: As I've mentioned above, this may be blocked by the ZK proof generation.

txhsl (changes requested):

consensus/dbft/dkg.go
```diff
@@ -340,275 +340,315 @@ func (c *DBFT) handleDKG(snapshot *snapshot, keystore *antimev.KeyStore, h *type
...
-							log.Error("DKG get transaction receipt status error", "txHash", item.TxHash, "status", receipt["status"])
+
+						// Send successfully, wait 3 blocks to check transaction status.
+						if item.TxHash != nil && currentHeight-item.SendHeight == 3 {
+							receipt, err := c.txAPI.GetTransactionReceipt(context.Background(), *item.TxHash)
+							if err != nil {
+								needRetry = true
+								log.Error("DKG get transaction receipt failed", "err", err, "txHash", item.TxHash)
+							}
+							if receipt["status"] != types.ReceiptStatusSuccessful {
```
qianhh: If the call to `GetTransactionReceipt()` fails and returns an error, then the `receipt` variable will be nil. Attempting to access it directly will cause a panic.

```suggestion
							} else if receipt["status"] != types.ReceiptStatusSuccessful {
```

txhsl: Seems like an issue not related to this PR but the original #332, we may open another PR to fix this.

But the fix is simple, yes, we can commit your suggestion now as well.

qianhh: Yes, I just happened to see this code and noticed a minor issue here. So I commented on it right there.

txhsl:
Rebased to the main branch, otherwise may encounter static pool errors in testing.

consensus/dbft/dbft.go
AnnaShaleva: Good catch, this change was added accidentally. The fix should be a part of this commit, otherwise this commit is illegal. So I moved this fix to this commit, will push an update soon.

consensus/dbft/dbft.go
```diff
@@ -238,9 +238,10 @@ type DBFT struct {
 	fakeDiff bool // Skip difficulty verifications
 
 	// The fields for dkg
-	dkgSnapshot       *snapshot
-	dkgTaskExecutorCh chan *taskList
-	dkgTaskWatcherCh  chan *taskList
+	dkgSnapshot        *snapshot
+	dkgTaskExecutorCh  chan *taskList
+	dkgTaskWatcherCh   chan *taskList
+	dkgWatcherLooperCh chan struct{}
```
AnnaShaleva: And this new channel is useless and may be avoided even if we'd like to send heartbit irrespective from tasks processing.  Will push an update in a few hours.

consensus/dbft/dkg.go
```diff
@@ -340,275 +340,332 @@ func (c *DBFT) handleDKG(snapshot *snapshot, keystore *antimev.KeyStore, h *type
...
+		select {
+		case <-c.quit:
+			return nil
+		case c.dkgTaskWatcherCh <- nil:
+		default:
+			// Use non-blocking send since if dkgTaskWatcherCh is busy with processing
+			// some other tasks, then triggering it one more time is completely useless.
+			log.Info("failed to send heartbit signal to DKG task watcher: handler is busy, will retry after the next block",
+				"index", h.Number.Uint64())
+		}
```
AnnaShaleva, line 364: This part replaces `dkgWatcherLooperCh`.

consensus/dbft/dkg.go
```diff
@@ -340,275 +340,332 @@ func (c *DBFT) handleDKG(snapshot *snapshot, keystore *antimev.KeyStore, h *type
...
+			sevenMsgPK = nil
+
+			if len(watchList) > 0 {
+				select {
+				case <-c.quit:
+					break executeLoop
+				case c.dkgTaskWatcherCh <- &watchList:
+					// Blocking send since transactions have been constructed and sent, we need to
+					// notify dkgTaskWatcherCh anyway even if it will block dBFT operation.
 				}
```
AnnaShaleva: @txhsl pay attention to this part since the behaviour was different in your commit. I think it's important to wait until `dkgTaskWatcherCh` will pick up this new watchlist because the transactions were sent to the network, so we need to track their status anyway. So it's a valid reason to block even standard dBFT flow.

Also, one iteration of `dkgTaskWatcherCh` handler doesn't take a lot of time since this handler just tries to resend existing tasks. So we shouldn't spend an eternity in `case c.dkgTaskWatcherCh <- &watchList:`.

consensus/dbft/dbft.go
AnnaShaleva: Updated, see the reviewcomment.

consensus/dbft/dkg.go
```diff
@@ -340,275 +340,331 @@ func (c *DBFT) handleDKG(snapshot *snapshot, keystore *antimev.KeyStore, h *type
 		if err := keystore.Persist(); err != nil {
 			return fmt.Errorf("failed to persist keystore, err: %v", err)
 		}
+		// Notify dkgTaskExecutorCh iff there's new task.
 		if len(taskList) > 0 {
 			select {
-			case c.executeProofTaskChan <- &taskList:
+			case <-c.quit:
```
txhsl, line 346: I have a question about the usage of `select` here and several places else, outside `for {}` blocks.

AFAIK, if both of the two cases are executable, then Golang will randomly pick one of them.

I've encountered `send on closed channel` in the first implementation. Is it really safe to use as this?

consensus/dbft/dkg.go
txhsl: The `dkgTaskWatcherCh` won't be that busy, so this will make the watcher to run one more time than once per block, at several specific time points.

But if you insist that we shouldn't have one more channel to handle the heartbeats, I'm OK with it.

consensus/dbft/dkg.go
```diff
@@ -340,275 +340,331 @@ func (c *DBFT) handleDKG(snapshot *snapshot, keystore *antimev.KeyStore, h *type
...
-					log.Error("Unsupported ZK version", "version", task.ZKVersion)
-					continue
+					panic(fmt.Errorf("unknown DKG method to prove: %s", task.Method))
+				}
+			}
+			// Release the cache when the task list is fully handled.
+			sevenMsgR1CS = nil
+			sevenMsgPK = nil
+
+			if len(watchList) > 0 {
```
txhsl: IMO, the `watchList` won't be `nil` here, so I prefer to remove the check.

consensus/dbft/dkg.go
txhsl: OK then, I see your concern. But I'm worrying about the usage of `select`, which makes the blocking here dangerous, ref https://github.com/bane-labs/go-ethereum/pull/467#discussion_r2140673705.

consensus/dbft/dkg.go
AnnaShaleva: Agree, fixed.

consensus/dbft/dkg.go
AnnaShaleva: It's a valid question.

> AFAIK, if both of the two cases are executable, then Golang will randomly pick one of them.

That's true, except the default case. Default is executed only when no other options are available.

> I've encountered send on closed channel in the first implementation.

It happens due to the fact that `dkgTaskWatcherCh` may be closed in the end of `dkgTaskWatcher` earlier than subsequent `handleDKG` call finishes its execution. So technically, there was a race between `handleDKG` sending to `dkgTaskWatcherCh` and `dkgTaskWatcher` closing `dkgTaskWatcherCh`. It's a bug, and it should be solved by proper awaiting of all routines completion before closing `dkgTaskWatcherCh`. The same is relevant for `dkgTaskExecutorCh`.

Updated, but we need to test the updated version.

consensus/dbft/dkg.go
AnnaShaleva: For this case proper awaiting of eventLoop before closing `dkgTaskWatcherCh` should also solve the problem, see the updated version.

consensus/dbft/dkg.go
txhsl: I see the difference, will have a detailed check tomorrow.

txhsl (approved):

Issue #446, opened 2025-04-01
`subTotalcost` panics when trying to delete expired transaction in `encrypted` queue


txhsl:
#### System information

Geth version: `v0.3.1`
OS & Version: Linux

#### Expected behaviour

Any transactions in Geth queue (as well as the `encrypted` cache after #316) has a 3h-lifetime limit as default. After being added into mempool and get expired, these transactions should be removed safely.

#### Actual behaviour

Currently, on T4, we've set up such a node with `--txpool.signaturecache` enabled, and it panicked with `totalcost underflow` when `removeTx`. BTW, this happened when T4 stopped, which means no block was generated during that 3 hours.

#### Backtrace

````
INFO [03-20|07:56:24.943] encrypted transaction cached             txHash=ef0796..46dc97 sender=0x2e211B562507169E1C444bA39d77feD711a69A89 nonce=162
WARN [03-20|07:56:24.943] Served eth_sendRawTransaction            conn=127.0.0.1:44542 reqid=1579251176812223 duration=1.220517ms err="encrypted transaction cached"
WARN [03-20|07:56:37.833] Served eth_call                          conn=127.0.0.1:57634 reqid=0c755813-ea2e-4024-a4f4-84c164d5d37a duration="466.739µs" err="execution reverted" errdata=0x7352d91c
WARN [03-20|07:56:38.030] Served eth_call                          conn=127.0.0.1:57638 reqid=01d60508-5fa9-4cd8-b293-b745cb55b6f2 duration="450.358µs" err="execution reverted" errdata=0x7352d91c
WARN [03-20|07:56:38.237] Served eth_call                          conn=127.0.0.1:57648 reqid=d0d0eeb0-f58c-45e9-bc63-a4058673249f duration="447.251µs" err="execution reverted" errdata=0x7352d91c
WARN [03-20|07:56:38.432] Served eth_call                          conn=127.0.0.1:57662 reqid=2c241b8f-d87a-4ebf-ad01-1e1eef30f735 duration="507.322µs" err="execution reverted" errdata=0x7352d91c
INFO [03-20|08:48:35.799] Regenerated local transaction journal    transactions=2 accounts=1
INFO [03-20|09:48:35.799] Regenerated local transaction journal    transactions=2 accounts=1
INFO [03-20|10:48:35.799] Regenerated local transaction journal    transactions=2 accounts=1
panic: totalcost underflow

goroutine 4192 [running]:
[...]

PR #468, opened 2025-05-13
core/txpool: refactor amev transaction management inside mempool

### LINKED ISSUE #372: Refactor amev transaction management inside mempool

qianhh:
Close #372.

Refactor the amev transaction management within the memory pool and move it all to CachePool.


qianhh:
At present, the original implementation is still being used. Only one transaction of the current nonce is cached for each address.

https://github.com/bane-labs/go-ethereum/blob/22b9ca37e5ce10662698e4d4cef5aee1c2193524/core/txpool/cachepool/cachepool.go#L515-L526

qianhh:
In order to separate from LegacyPool, some code was copied from it. There are quite a few redundancies. I wonder if anyone has a better improvement plan?

@AnnaShaleva, @txhsl, @songb2 you're welcomed to review and test.

core/txpool/txpool.go
```diff
@@ -324,6 +327,11 @@ func (p *TxPool) Add(txs []*types.Transaction, local bool, sync bool) []error {
 
 		// Try to find a subpool that accepts the transaction
 		for j, subpool := range p.subpools {
+			// If the enableAMEVCache flag is set to true, then the first item in the subpools is cachepool.
+			// If the transaction is local, then the cache pool condition judgment is required.
+			if p.enableAMEVCache && j == 0 && !local {
```
txhsl: Can we judge based on pool type instead of its position, and remove `enableAMEVCache` from the pool structure as well as the input parameters?

Just for reference.

https://github.com/bane-labs/go-ethereum/blob/97683e25c0363c575355e1fa4f73bc6c7cc84141/eth/backend.go#L291-L304

core/txpool/cachepool/cachepool.go
```diff
@@ -0,0 +1,1065 @@
...
+		reorgDoneCh:     make(chan chan struct{}),
+		reorgShutdownCh: make(chan struct{}),
+		initDoneCh:      make(chan struct{}),
+	}
+
+	return pool
+}
+
+// Filter returns whether the given transaction can be consumed by the legacy
+// pool, specifically, whether it is a Legacy, AccessList or Dynamic transaction.
```
txhsl: The comments should be paid attention to, like the word `legacy` here. The method is returning whether the given transaction can be consumed by this amev cache pool.

core/txpool/cachepool/cachepool.go
```diff
@@ -0,0 +1,1065 @@
...
+func (pool *CachePool) Close() error {
+	// Terminate the pool reorger and return
+	close(pool.reorgShutdownCh)
+	pool.wg.Wait()
+
+	log.Info("Transaction pool stopped")
+	return nil
+}
+
+// Reset implements txpool.SubPool, allowing the legacy pool's internal state to be
```
txhsl: `legacy`. Maybe we can refer to the comment style of blob pool.

core/txpool/cachepool/cachepool.go
```diff
@@ -0,0 +1,1065 @@
...
+	}
+
+	// Check tx ready to process based pending list. The transaction's nonce should be
+	// equal to the pending nonce.
+	//
+	// NOTE: After adding this judgment, only the next nonce transaction can be added.
+	// The previous pending transactions cannot be replaced and an error will be
+	// returned directly. Should we make any improvements in the future?
+	pendingNonce := pool.pendingNonces.get(from)
+	if tx.Nonce() < pendingNonce {
```
txhsl: I think we can have a looser check here.

In the legacy pool, users are able to replace pending transactions and queued transactions, we shouldn't be more strict than that, this judgement can be checked based on the finalized nonce.

Within the pool limit, I'm expecting every possible future transaction can be cached or replaced here, until its nonce gets confirmed in a new block or the transaction itself expires.

core/txpool/cachepool/cachepool.go
```diff
@@ -0,0 +1,1065 @@
...
+	// Sanitize the input to ensure no vulnerable gas prices are set
+	config = (&config).sanitize()
+
+	// Create the transaction pool with its initial settings
+	pool := &CachePool{
+		config:          config,
+		chain:           chain,
+		chainconfig:     chain.Config(),
+		signer:          types.LatestSigner(chain.Config()),
+		pending:         make(map[common.Address]*list),
```
txhsl: I prefer a `cached` naming here, as well as for the noncer, because these transactions are not pending for confirmation, but for its sender to fetch.

Also for the stats, I prefer to count them in queued or a new number. How do you think?

core/txpool/cachepool/cachepool.go
```diff
@@ -0,0 +1,1065 @@
...
+	for i, tx := range txs {
+		_, err := pool.add(tx)
+		errs[i] = err
+	}
+	return errs
+}
+
+// Status returns the status (unknown/pending/queued) of a batch of transactions
+// identified by their hashes.
+func (pool *CachePool) Status(hash common.Hash) txpool.TxStatus {
```
txhsl: Also here, we may need a new `TxStatus` for this pool.

core/txpool/cachepool/cachepool.go
```diff
@@ -0,0 +1,1065 @@
...
+	var (
+		newTip = uint256.MustFromBig(tip)
+	)
+	pool.gasTip.Store(newTip)
+	log.Info("Cache pool tip threshold updated", "tip", newTip)
+}
+
+// Nonce returns the next nonce of an account, with all transactions executable
+// by the pool already applied on top.
+func (pool *CachePool) Nonce(addr common.Address) uint64 {
```
txhsl: This result should not cover the nonce from legacy pool, otherwise it causes issues in the Envelope sending from wallets.

core/txpool/cachepool/cachepool.go
```diff
@@ -0,0 +1,1065 @@
...
+// identified by their hashes.
+func (pool *CachePool) Status(hash common.Hash) txpool.TxStatus {
+	if pool.Has(hash) {
+		return txpool.TxStatusPending
+	}
+	return txpool.TxStatusUnknown
+}
+
+// Get returns a transaction if it is contained in the pool and nil otherwise.
+func (pool *CachePool) Get(hash common.Hash) *types.Transaction {
```
txhsl: I think this may should return `nil`, transactions from this pool should only be accessed by `GetCachedTransaction`.

I've mentioned the issue, this pool is only for cache, so it should not influence any RPC data, API data, or expose the transaction through any other ways expect `GetCachedTransaction` with nonce and sig.

core/txpool/cachepool/cachepool.go
```diff
@@ -0,0 +1,1065 @@
...
+}
+
+// get returns a transaction if it is contained in the pool and nil otherwise.
+func (pool *CachePool) get(hash common.Hash) *types.Transaction {
+	return pool.all.Get(hash)
+}
+
+// Has returns an indicator whether txpool has a transaction cached with the
+// given hash.
+func (pool *CachePool) Has(hash common.Hash) bool {
```
txhsl: Methods like this should also be reconsidered.

txhsl (commented):
The new file `noncer.go` and `list.go` seem to have the same content from legacy pool.

Maybe we should try to find a way to reuse them, if we are not going to change their content.

core/txpool/txpool.go
qianhh: I have tried this method, but in the txpool, it is not possible to make type assertions and compare which type it is. This would cause a circular reference.

txhsl: Yes, I see the problem.

But at least we need to untie the declaration about position `0` in `eth/backend` and the check condition here in `core/txpool`.

They are two different packages, and we are not requiring the cache pool to be the first subpool in method `New()`.

What about `reflect.TypeOf()`?

core/txpool/cachepool/cachepool.go
qianhh: Okay, I'll refactor the variable names.

core/txpool/cachepool/cachepool.go
qianhh: I added a new `TxStatus`.

https://github.com/bane-labs/go-ethereum/blob/fc709e9e5d18c6e2162bf0302d5b8db1eaec5deb/core/txpool/txpool.go#L37-L45
https://github.com/bane-labs/go-ethereum/blob/fc709e9e5d18c6e2162bf0302d5b8db1eaec5deb/core/txpool/cachepool/cachepool.go#L618-L625

core/txpool/cachepool/cachepool.go
qianhh: I've relaxed the conditions to allow caching as long as it's not smaller than the current nonce.

core/txpool/txpool.go
qianhh: The `CachePool` should be placed at the top. Otherwise, the `Filter()` method in `LegacyPool` needs to be used to filter out the transactions that need to be cached. However, the `Filter()` method lacks the `local` parameter, making it difficult to determine. Unless it is modified or a new method is added.

txhsl: I see the issue, remote transactions also use this way.

core/txpool/subpool.go
```diff
@@ -168,4 +168,8 @@ type SubPool interface {
 
 	// GetCachedTransaction returns the encrypted transaction cached in tx pool.
 	GetCachedTransaction(nonce uint64, sender common.Address) *types.Transaction
+
+	// FilterEx is a selector used to decide whether a transaction would be added
+	// to this particular subpool.
+	FilterEx(tx *types.Transaction, local bool) bool
```
txhsl: We need another name for this, and describe its difference from `Filter` and when should it be used.

To me, `FilterEx` is only used when adding a new transaction, while `Filter` is used in reset. We can use `FilterAdd` for this new method, or maybe you can explain its meaning so I can understand this naming.

Remember to put this declaration just behind `Filter`.

core/txpool/legacypool/legacypool.go
```diff
@@ -304,6 +306,19 @@ func (pool *LegacyPool) Filter(tx *types.Transaction) bool {
 	}
 }
 
+// FilterEx returns whether the given transaction can be consumed by the legacy
+// pool, specifically, whether it is a Legacy, AccessList or Dynamic transaction.
+func (pool *LegacyPool) FilterEx(tx *types.Transaction, local bool) bool {
```
txhsl: Updating comments is necessary. To me, the usage of two `Filter` are changed, so I prefer to update both comments to describe when should the method be used.

txhsl:
I need to review it again, and then we go for tests.

core/txpool/cachepool/cachepool.go
```diff
@@ -0,0 +1,1018 @@
...
+	}
+}
+
+// Close terminates the transaction pool.
+func (pool *CachePool) Close() error {
+	// Terminate the pool reorger and return
+	close(pool.reorgShutdownCh)
+	pool.wg.Wait()
+
+	log.Info("Transaction pool stopped")
```
txhsl: Maybe we need a message different from legacy pool.

core/txpool/cachepool/cachepool.go
```diff
@@ -0,0 +1,1018 @@
...
+			// by a return statement before running deferred methods. Take care with
+			// removing or subscoping err as it will break this clause.
+			//
+			// This error "ErrTxPoolCached" indicates successful caching, so it needs to be excluded.
+			if err != nil && !errors.Is(err, ErrTxPoolCached) {
+				pool.reserve(from, false)
+			}
+		}()
+	}
+	// If the transaction pool is full, discard underpriced transactions
```
txhsl: > If the transaction pool is full, discard underpriced transactions

Is it? I think this line is misleading.

txhsl (commented):
LGTM. @songb2 I think we can start to have some tests.

And @AnnaShaleva is welcome for any further review about this.

core/txpool/cachepool/cachepool.go
qianhh: I have revised this comment to make its meaning clearer and more understandable.

vncoelho (commented):
such a good refactor and improvement

vncoelho (commented):
I am just a little confused about the term blobpool here.
In the original client blob is now only accessed by CNs.

txhsl:
> I am just a little confused about the term blobpool here. In the original client blob is now only accessed by CNs.

It is the original design from Geth. Blob transactions are not in a permanent storage, and not every full node should download them.

AnnaShaleva (commented):
This implementation doesn't match the original proposal (https://github.com/bane-labs/go-ethereum/pull/316#discussion_r1873097905). The problem is still in the code duplication between legacy pool and cache pool. What I'd like to see in this PR is a compatible extension of `legacypool.LegacyPool` behaviour that is enabled/disabled depending on some `SignaturesCache` setting flag; without allocation of new CachePool structure.

The motivation for that is the same as discussed earlier: the basic pipeline of encrypted transactions validation at the mempool level is the same as for simple transactions (except for some amev-secific rules), by reusing LegacyPool code we can guarantee that. Also, code duplication makes it harder to sync with the original Geth code (i.e. who will be responsible for tracking all LegacyPool-related changes implemented in the original Geth repo and porting them to CachePool if we maintain CachePool as a separate structure in a separate file).

txhsl:
> What I'd like to see in this PR is a compatible extension of `legacypool.LegacyPool` behaviour that is enabled/disabled depending on some `SignaturesCache` setting flag; without allocation of new CachePool structure.

Ref https://github.com/bane-labs/go-ethereum/pull/468/files#diff-9cc7046826ce92d9cc5432cf41ed4fc14bb6907da44f771e3932302c40e8dbc2R243-R254, the new `CachePool` is enabled/disabled based on the node configurations.

Based on this, I want to explain the proposal of this PR.

The original #316 allocates 2 different queues `pending` and `encrypted` to handle normal transactions and transactions need cached separately. The `encrypted` queue is allocated anyway, and somehow breaks the `LegacyPool` rules, e.g. #446 and several UTs about `LegacyPool`. The UTs are hard to fix because we need to maintain a different logic in `LegacyPool` with transaction cache.

That's why we have #372, we need a separated subpool to realize the transaction cache.

So your comment is focusing on the code reusage and maintenance. About the code reusage, I'd say yes, I have the same idea. As you can see, the `list.go` file and `lookup` of `CachePool` are using almost the same code as them in `LegacyPool`.

If we are talking about the reusage of `list.go` and `lookup`, then I also have the same idea of merging these two packages.

> The motivation for that is the same as discussed earlier: the basic pipeline of encrypted transactions validation at the mempool level is the same as for simple transactions (except for some amev-secific rules), by reusing LegacyPool code we can guarantee that.

However, this is not 100% correct. We're already reusing the transaction validation as it in `/txpool/validation.go`, but I have to say it is they are the only thing we can reuse from `LegacyPool`. Because the `CachePool` has different and simple rules. It behaves quite different from `LegacyPool` so that the RPC returns correct data for user wallets, e.g. the nonce management, address reverse, and two pools are even conflict. So it's hard to reuse any logic of `LegacyPool`, also from the Golang language perspective, that's why we still need to create this `CachePool` with its own code.

Another point is that `LegacyPool` has been updated a lot of times since `v1.13.15`, the code is still alive. So changing the basic handling logic of `LegacyPool` is not good for further maintenance as well.

@AnnaShaleva How do you think?

qianhh:
I moved the CachePool to the legacypool package, which enables the sharing of more components and reduces redundant code.

txhsl (approved):

Issue #244, opened 2024-07-04
Add extensible whitelist cache to dBFT protocol


AnnaShaleva:
# Rationale

#240 adds extensible payloads verification at the dBFT protocol level for all (both CN and seed) nodes. This functionality covers our current needs, because currently we have the only type of extensible payloads (consensus payloads) and the only type of valid extensible senders (validators). However, in future we may wont to add more extensible types (not likely though) or support more extensible senders (more likely). Thus, we need to make extensible whitelist management more "smart" and add extensible whitelist cache.

Note that this issue is not so critical because we have cache managed by `getValidators`:
https://github.com/bane-labs/go-ethereum/blob/47d271238a77ddc36865ca697f3e2aff8e9d87e1/consensus/dbft/dbft.go#L1643-L1648
# Implementation

It should work similar to NeoGo extensible whitelist management:
https://github.com/nspcc-dev/neo-go/blob/5566e354d451fd14d3b2a856cfe4b1b654a7343e/pkg/core/blockchain.go#L1802

I tried to implement the same approach for NeoX (ref. 96f4a9cec60eeff7c481ec60f0d6bb1a99dda776, branch https://github.com/bane-labs/go-ethereum/tree/extensible-whitelist), but the tricky part is that this approach works only for CN nodes. Ideally, extensible whitelist should be managed by the blockchian, not by the dBFT engine, and thus it would probably be better to move `getValidators` call from dBFT to the Eth blockchain API.

AnnaShaleva:
Depends on #299.

PR #469, opened 2025-05-23
dbft: detach dBFT extensible payloads verifier from dBFT engine

### LINKED ISSUE #299: Detach dBFT extensible payloads verifier from dBFT engine
# Rationale

Extensible payloads verification does not depends on the dBFT engine in any way. The only common thing between them is the same algorithm for next block validators calculation. dBFT engine may be turned on/off (for non-mining/mining nodes), whereas extensible verification should always be enabled.

# Implementation

Extensible payloads verification (`(c *DBFT) IsExtensibleAllowed`) should be moved to a separate tiny service by its own. This service should track Downloader events to properly react on chain sync, but it should not be a part of dBFT, the current design isn't good, ref. #298. Also, extensible verification service should have its own validators cache (#244).

qianhh:
Close #299 .

The original extensible payloads verifier function has been separated from `dbft` and is now part of the `eth` module. It can be called through the Eth API.


txhsl:
I'm OK with these modifications, but I think @AnnaShaleva may expect more about the extensible payload, since we haven't changed the basic logic.

AnnaShaleva (changes requested):
> The original extensible payloads verifier function has been simply separated.

It doesn't solve neither https://github.com/bane-labs/go-ethereum/issues/244 nor https://github.com/bane-labs/go-ethereum/issues/299. Both these issues are not just about moving extensible-related code to a separate structure/file.

The purpose of https://github.com/bane-labs/go-ethereum/issues/299 is to enable extensible payloads verification not only for CNs, but also for non-mining nodes. That's why it is suggested to abstract extensible verification service, but the key issue is that his service should be properly integrated with dbft network protocol to be active even on seed nodes (right now extensible sender is not verified by non-mining nodes):
https://github.com/bane-labs/go-ethereum/blob/3ac8c58810a40a93706435e9b6e9f055814b88fc/eth/protocols/dbft/pool.go#L94
So probably it would be a good idea to make this service managed by blockchain itself, i.e. to make it a part of Eth blockchain API (right now it's bound to the activity of dBFT event loop).

The purpose of https://github.com/bane-labs/go-ethereum/issues/244 is to implement per-block cache for extensible whitelist members and update this cache only when it's necessary (right now - only for those blocks when committee may be changed). This cache should work exactly the same way as in NeoGo, refer to the original issue for link and for initial implementation. Also, in NeoGo this cache is bound to the blockchain's block processing routine, since it's quite native to update the whitelist exactly after the block is processed by the chain. The same logic should be suitable for Eth.

qianhh:
> It doesn't solve neither #244 nor #299. Both these issues are not just about moving extensible-related code to a separate structure/file.
> 
> The purpose of #299 is to enable extensible payloads verification not only for CNs, but also for non-mining nodes. That's why it is suggested to abstract extensible verification service, but the key issue is that his service should be properly integrated with dbft network protocol to be active even on seed nodes (right now extensible sender is not verified by non-mining nodes):
> 
> https://github.com/bane-labs/go-ethereum/blob/3ac8c58810a40a93706435e9b6e9f055814b88fc/eth/protocols/dbft/pool.go#L94
> 
> 
> So probably it would be a good idea to make this service managed by blockchain itself, i.e. to make it a part of Eth blockchain API (right now it's bound to the activity of dBFT event loop).
> The purpose of #244 is to implement per-block cache for extensible whitelist members and update this cache only when it's necessary (right now - only for those blocks when committee may be changed). This cache should work exactly the same way as in NeoGo, refer to the original issue for link and for initial implementation. Also, in NeoGo this cache is bound to the blockchain's block processing routine, since it's quite native to update the whitelist exactly after the block is processed by the chain. The same logic should be suitable for Eth.

Previously, I misunderstood the meaning. So, I'll first solve the #299 issue and incorporate it into the Eth blockchain API.

AnnaShaleva:
> incorporate it into the Eth blockchain API.

Note that it’s just an assumption, we need to check may be there’s a better place for this cache.

qianhh:
I have completed the integration work. Now let's see if there are any areas that can be improved? @AnnaShaleva

txhsl:
I'll review this if you want to get merged.

qianhh:
It's ready now.

consensus/dbft/dbft.go
```diff
@@ -1105,10 +1090,12 @@ func (c *DBFT) processPreBlockCb(b dbft.PreBlock[common.Hash]) error {
 			sharesCurr = make(map[int][]*tpke.DecryptionShare, ctx.M())
 			sharesPrev = make(map[int][]*tpke.DecryptionShare)
 			blockNum   = pre.header.Number.Uint64() - 1
+			dkgIndex   int
+			err        error
```
txhsl: I don't see the benefit to extend the lifecycle of `dkgIndex` and `err`, seems we can keep them as before?

consensus/dbft/dbft.go
```diff
@@ -1492,8 +1479,9 @@ func (c *DBFT) getBlockWitness(pub *tpke.PublicKey, block *Block) ([]byte, error
 		for i := 0; i < len(vals); i++ {
 			if p := dctx.CommitPayloads[i]; p != nil && p.ViewNumber() == dctx.ViewNumber {
 				var err error
+				var dkgIndex int
```
txhsl: As well as here, `dkgIndex` and `err` seems not necessary to be declared in `var` scope.

I prefer to remove L1481 instead of add L1482.

eth/protocols/dbft/pool_test.go
```diff
@@ -140,6 +141,8 @@ func (c *testChain) IsAddressAllowed(u common.Address) error {
 }
 func (c *testChain) BlockHeight() uint64 { return c.height }
```
txhsl: Can we remove this `BlockHeight`? It seems no where to be used, and should be replaced by `BlockNumber`.

eth/backend.go
```diff
@@ -467,12 +467,14 @@ func New(stack *node.Node, config *ethconfig.Config) (*Ethereum, error) {
 		}
 		return eth.beacon.GetTransaction(hash)
 	})
+
+	syncing := func() bool {
```
txhsl: This change works and I'm OK with it. But I wish this `syncing` can be renamed to a `bftSyncing` or something else, and we add a comment here to declare that this syncing check is only for dBFT.

The current beacon implementation is for dBFT only (so as this syncing check), but the ability to connect to a Ethereum PoS CL is also preserved. Our idea is to bring dBFT to Geth, instead of making Geth only work for dBFT (although we never test whether the Neo X node can still work in other EVM networks).

eth/backend.go
```diff
@@ -467,12 +467,14 @@ func New(stack *node.Node, config *ethconfig.Config) (*Ethereum, error) {
...
+		log.Debug("Syncing status", "initialized", eth.beacon.InitialSynced(), "beacon", eth.beacon.Syncing(), "eth", eth.Syncing())
+		// Ignore the beacon syncing status after the initial sync, to prevent blocking expected dBFT message handling.
+		return !eth.beacon.InitialSynced() || eth.Syncing()
+	}
+	eth.extensibleVerifier = verifier.New(eth.APIBackend, syncing)
```
txhsl: Is it better if it's `verifier.NewExtensibleVerifier(eth.APIBackend, syncing)`, or `NewExtensibleVerifier(eth.APIBackend, syncing)`?

I'm not sure if I fully get the original idea.
1. If we only need this extensible verifier, then we don't need to new the `verifier` folder, but put it under `eth`, so use it as `NewExtensibleVerifier(eth.APIBackend, syncing)`;
2. If we may want to extend the types of verifier, then we have the new `verifier` package, and use it as `verifier.NewExtensibleVerifier(eth.APIBackend, syncing)`.

Anyway, it should be one of the two.

qianhh: Yes, this place was indeed poorly planned. Let's go with the second option.

txhsl: It also depends on your understanding of #240. So far maybe we know few about this idea from N3.

txhsl (approved):
@songb2

eth/api_backend.go
```diff
@@ -563,3 +564,23 @@ func (b *EthAPIBackend) BlobSidecarByRoot(ctx context.Context, hash common.Hash,
 	}
 	return blockBlobs[index], nil
 }
+
+func (b *EthAPIBackend) ExtensiblePayloadsVerifier() *verifier.ExtensibleVerifier {
+	return b.eth.extensibleVerifier
+}
+
+func (b *EthAPIBackend) IsExtensibleAllowed(blockNum uint64, addr common.Address) error {
+	return b.eth.extensibleVerifier.IsExtensibleAllowed(blockNum, addr)
```
txhsl: Is it possible that a `dbft` message comes before the verifier is initialized, so causes a null pointer panic?

txhsl: If so, then we need to separate its initialization and the connection to beacon syncing check.

Return `ErrSyncing` or another error before the beacon instance is created and connected.

txhsl: > Is it possible that a `dbft` message comes before the verifier is initialized, so causes a null pointer panic?

Checked and seems not.

txhsl:
The circular dependency between backend and extensible verifier doesn't bring direct issues but looks somehow risky.

I tried to improve it in the same way as gasprice oracle. How do you think? @qianhh

qianhh:
> The circular dependency between backend and extensible verifier doesn't bring direct issues but looks somehow risky.
> 
> I tried to improve it in the same way as gasprice oracle. How do you think? 

OK, that's a good idea.

PR #473, opened 2025-05-28
dbft: refactor static pool and DKG snapshot initialization

AnnaShaleva:
Close #471, a replacement of #472.

>  I've seen the problem is CV related.

@txhsl is right, to reproduce problem from #471 the CN should be shutdown, then wait until it's her turn to accept block, wait for other CNs will accept CV and then turn on the missing CN (but before other nodes will accept this block at view 1).

So this PR needs testing.

PR #476, opened 2025-06-11
dbft: prohibit to initialize local pool with empty root

AnnaShaleva:
For the case of dBFT it's not correct to initialize local pool with empty root because it's critical for dBFT to have relevant state for proper proposal verification. dBFT must guarantee that by the moment of static pool initialization the node's state at the given root is present; if it's not true then it's a bug that should be fixed on the dBFT side.

Should be a part of #473. If this error ever happens during node lifetime, then we need to fix the way how dBFT handles node's synchronisation process.

### DIFF
--- a/consensus/dbft/dbft.go
+++ b/consensus/dbft/dbft.go
@@ -1278,7 +1278,7 @@ func newStaticPool(chain ChainHeaderReader) *legacypool.LegacyPool {
 
 // initStaticPool initializes the static pool with the provided parent header.
 func (c *DBFT) initStaticPool(parent *types.Header, state *state.StateDB) error {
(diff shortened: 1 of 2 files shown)


txhsl (approved):

txhsl:
> Should be a part of #473. If this error ever happens during node lifetime, then we need to fix the way how dBFT handles node's synchronisation process.

I think the issue of fetching an empty state is fixed in #473, by using the passed-in state directly in `postBlock()`.

We've not yet encountered the same error after that, but this switch is helpful to check and avoid it thoroughly.

AnnaShaleva:
> I think the issue of fetching an empty state is fixed in https://github.com/bane-labs/go-ethereum/pull/473,

Agree, it should be fixed by this change.

PR #480, opened 2025-06-24
antimev: allow startup with an empty backup keystore

### LINKED ISSUE #479: Allow node to startup with an empty anti-mev keystore at any time
# Rationale

Related to #405. Anti-mev keystore is the local storage of our on-chain DKG. It contains all the secrets generated locally and received from the network.

The latter can be synchronized from the KeyManagement contract at any moment, as long as the message decryption key is always known. But the former needs a `Recover` phase in DKG, otherwise will be lost forever if the keystore file becomes not accessible.

The current implementation of our DKG handling in dBFT supposes that the node works in a very safe environment, with only network disconnections and unexpected interruptions can happen, e.g. #332, #444 and #475. We suppose that the keystore only stops at a history time point, but no data lost happens.

But as @AnnaShaleva has suggested, we need to take different cases into consideration, to make this mechanism stronger. This time, we are going to allow CNs to use an empty backup keystore to startup at any time.

# Implementation

First, we are not going to change the threshold number of malicious nodes, using an empty backup keystore is only a local behaviour change.

Second, in #332 and #444, I have separated the secret handling and key generation. Since our epoch change only relies on the verification of KeyManagement contract, the BLS12381 secret key and public key required by dBFT and Envelope decryption should be easily recoverable from public data (as long as the message decryption key is consistent).

It means the hard-to-recover local secret doesn't affect our consensus and anti-mev directly. The lost of local secrets will prevent our CNs to send `Reshare` transactions, but `Recover` is not impacted.

~~Third, it's possible that node recover its local secret based on KeyManagement contract data, we can try to implement it.~~

### LINKED ISSUE #481: DKG synchronization access resharing key group when it doesn't exist
#### System information

Geth version: `v0.4.0`

#### Expected behaviour

`handleDKG` can synchronize from any time point with an out-of-date anti-mev keystore file.

#### Actual behaviour

When a keystore is out of at least 2 DKG epochs, `handleDKG` will call `syncLastRoundSecrets` to download the data of the epoch before the latest.

It also tries to insert the Reshare data, but the key group doesn't exist, so caused a `required keygroup not found` error.

### LINKED ISSUE #482: DKG epoch change uses an always-false `isShareReady` check
#### System information

Geth version: `v0.4.0`

#### Expected behaviour

#475 fixed the DKG behaviour under discontinue mining or synchronization. But also used an incorrect contract state during local epoch change.

#### Actual behaviour

The method `isShareReady` in KeyManagement contract returns if there are `n` shares have been accepted. But the at the time point of the calling in `handleDKG` epoch change, the contract itself has already started a new DKG round, so this method always returns `false`.

It results in an issue, that when there is a `Recover` stage happens, the result of recovered reshare will not be downloaded to local keystores, and causes a `secret share not enough` in `handleDKG`.

### LINKED ISSUE #483: Incorrect `ZKVersion` is sent to `dkgTaskExecutor`
#### System information

Geth version: `v0.4.0`

#### Expected behaviour

#444 add a new flag value `ZKVersion` to determine which kind of share should be generated and sent to KeyManagement contract.

#### Actual behaviour

But when tasking `reshareRecovered`, the `currentHeight` is put in a wrong position, so that used as `ZKVersion` in task executor. It will result in a `Unsupported ZK version` error, skip the message generation and this reshare can not be sent along with a transaction.

https://github.com/bane-labs/go-ethereum/blob/863cd5338f14f665322d44f1dbd4b66199f598fe/consensus/dbft/dkg.go#L335

txhsl:
Close #479. Close #481. Close #482. Close #483. Including some other fixes and optimizations found in DKG.

### DIFF
--- a/antimev/keystore.go
+++ b/antimev/keystore.go
@@ -13,6 +13,7 @@ import (
 	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/ethereum/go-ethereum/crypto/ecies"
 	"github.com/ethereum/go-ethereum/crypto/tpke"
+	"github.com/ethereum/go-ethereum/log"
 	"github.com/google/uuid"
 	keystorev4 "github.com/wealdtech/go-eth2-wallet-encryptor-keystorev4"
 )
@@ -200,10 +201,13 @@ func (ks *KeyStore) IsRecovering() bool {
 
 // OnValidatorList initializes sharing and resharing, should be called
 // when the key group members are determined. It initializes 1 or 2 key groups.
-func (ks *KeyStore) OnSharePeriodStart() {
+func (ks *KeyStore) OnSharePeriodStart(forceInitReshare bool) {
 	// Set up groups for dkg resharing
 	if ks.shared != nil {
 		ks.resharing = ks.shared.newTemplateForReshare(ks.size)
+	}
+	if ks.resharing == nil && forceInitReshare {
+		ks.resharing = newThresholdKeyGroup(ks.size)
 	}
 	// Set groups for dkg sharing
 	ks.sharing = newThresholdKeyGroup(ks.size)
@@ -276,7 +280,7 @@ func (ks *KeyStore) RevertRound() {
 
 // OnEpochChange checks and settles the results of sharing and resharing, will
 // revert both of them of any of them is not successful.
-func (ks *KeyStore) OnEpochChange(selfPvss []byte, aggregatedCmt []byte, isMemberOfNewGroup bool) error {
+func (ks *KeyStore) OnEpochChange(selfPvss []byte, aggregatedCmt []byte, lastRoundCmt []byte, isMemberOfNewGroup bool) error {
 	// Revert dkg if contract doesn't give a aggregated commitment
 	if len(aggregatedCmt) == 0 {
 		ks.RevertRound()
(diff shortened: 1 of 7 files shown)


txhsl:
@songb2

antimev/keystore.go
```diff
@@ -282,10 +282,7 @@ func (ks *KeyStore) OnEpochChange(selfPvss []byte, aggregatedCmt []byte, isMembe
 		ks.RevertRound()
 		return nil
 	}
-	if isMemberOfNewGroup && len(selfPvss) == 0 {
-		return ErrLengthMismatch
-	}
-	if !isMemberOfNewGroup && len(selfPvss) > 0 {
+	if isMemberOfNewGroup && len(selfPvss) == 0 || !isMemberOfNewGroup && len(selfPvss) > 0 {
```
qianhh: The two conditions mentioned here need to be enclosed in parentheses, right?
`if (isMemberOfNewGroup && len(selfPvss) == 0) || (!isMemberOfNewGroup && len(selfPvss) > 0) {`

qianhh (changes requested):

antimev/keystore.go
txhsl: Fixed.

qianhh (approved):

PR #490, opened 2025-07-10
crypto/tpke: antimev: enforce none-zero random scalar

txhsl:
In the current crypto scheme, we use `rand.Int()` to generate random numbers for DKG local secret generation, crypto commitment and encryption.

Even though the possibility is very low, but this function may output `0`, which weakens the security of our protocol in several perspectives.

This PR prevents zero values from appearing in `crypto/tpke` and `antimev`.


qianhh (approved):
LGTM

PR #503, opened 2025-08-28
dbft: check that DKG-related files are available on node startup

AnnaShaleva:
Avoid runtime problems that may occur if the node is started with improperly specified DKG files:
```
2025-08-28 01:16:30.023    ERROR[08-28|03:16:30.023] invalid r1cs file                        file=r1cs/seven_message.ccs err="open r1cs/seven_message.ccs: no such file or directory"
```

### DIFF
--- a/consensus/dbft/dbft.go
+++ b/consensus/dbft/dbft.go
@@ -26,6 +26,7 @@ import (
 	"fmt"
 	"io"
 	"math/big"
+	"os"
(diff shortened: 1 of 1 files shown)


txhsl (approved):
It verifies the file existence, it's good.

But we don't have a method to verify the file itself. Maybe we can find one in the future.

PR #508, opened 2025-09-01
params: migrate `NeoXDKGBlock` to time-based fork

### LINKED ISSUE #500: Refactor fork `NeoXDKG` from height-based to timestamp-based
# Rationale

Since `shanghai`, the fork definition has changed from `XXXBlock` to `XXXTime`.

Currently, we have the upgrade sequence as `shanghaiTime`=>`neoXDKGBlock`=>`neoXAMEVBlock`=>`neoXEthSigBlock`, and we will have `cancunTime`=>`pragueTime`=>`osakaTime` in the future.

The fork definitions need a unification, to only timestamp-based ones.

Also, the none-zero `neoXDKGBlock` and `neoXAMEVBlock` cause problems for maintains, if we change to timestamp-based, they can only be `0`.

# Implementation

1. Change fork definitions to only timestamp-based ones;
2. Migrate the configuration files of Testnet and Mainnet.

Update:

Since only `NeoXDKG` defines a different EVM behaviour, and the other two are just dBFT-related, we'll only refactor `NeoXDKG`.

txhsl:
Close #500. @AnnaShaleva This is my proposal of #500. It is not a must, but an idea.

Ethereum started to define upgrades only in timestamps since `Shanghai`, e.g. `Cancun`, `Prague` and `Osaka`. But we have `NeoXDKG`, `NeoXAMEV` and `NeoXEthSig` defined in block number in this sequence. 

Especially, `NeoXDKG` (number based) defines an EVM version between `Shanghai` (time based) and `Cancun` (time based). Although the EVM behaviour of our `NeoXDKG` and `Cancun` is almost the same, but the configuration is still confusing to node runners. If we don't have `Shanghai` enabled from `0`, then we will see the value of `NeoXDKG` is smaller than `Shanghai`.

I know `NeoXAMEV` and `NeoXEthSig` highly depend on the block number for judgement, and they make no difference to the EVM. So we can leave them defined in block numbers, but `NeoXDKG` can be easily migrated into a timestamp.

So how do you think about the upgrades, considering we may turn to a dynamic-time dBFT in the future?


roman-khimov:
> So how do you think about the upgrades, considering we may turn to a dynamic-time dBFT in the future?

I'm not sure if we can change existing HFs without affecting state, but otherwise :+1:. If we're to update contracts as well to not have block number dependencies we'd be ready for dynamic time blocks completely. And to me it's a very nice feature to have, we're very happy with how it works for NeoFS networks.

params/config.go
```diff
@@ -591,8 +591,8 @@ func (c *ChainConfig) IsShanghai(num *big.Int, time uint64) bool {
 }
 
 // IsNeoXDKG returns whether num is either equal to the NeoXDKG fork block or greater.
-func (c ChainConfig) IsNeoXDKG(num *big.Int) bool {
-	return c.IsLondon(num) && isBlockForked(c.NeoXDKGBlock, num)
+func (c ChainConfig) IsNeoXDKG(num *big.Int, time uint64) bool {
+	return c.IsLondon(num) && isTimestampForked(c.NeoXDKGTime, time)
```
AnnaShaleva, line 595: Then how about NeoXAMEVBlock and NeoXEthSigBlock?

Time-based forks should be consecutive, we can't get back to block-based forks once time-based is enabled.

AnnaShaleva:
> if we can change existing HFs without affecting state

We can.

params/config.go
txhsl: > Then how about NeoXAMEVBlock and NeoXEthSigBlock?

It's hard to change them. DBFT use heights in judgement, and there are both `Add` and `Sub` computation. And only block number is transferred in the dBFT library. Honestly, I have no idea about how to migrate them, as well as how to enable them from `0`.

But it's OK to leave them block-based. If we start a new network, e.g. a new Testnet, NeoXAMEVBlock and NeoXEthSigBlock won't be enabled at `0`, and they have to be arranged behind everything in the new genesis, to wait for at least a round of DKG.

> Time-based forks should be consecutive, we can't get back to block-based forks once time-based is enabled.

But forks from the Ethereum upstream will be all time-based.

AnnaShaleva: > And only block number is transferred in the dBFT library. 

I may change it and adapt on the dBFT side if we decide to migrate these forks to timestamp-based behaviour, it's possible. It will require @roman-khimov's ACK.

> But forks from the Ethereum upstream will be all time-based.

The problem is that both Shanghai and Cancun are time-based as post-merge forks. So if we enable NeoXAMEV and NeoXEthSig in between Shanghai and Cancun, then I'd expect both of them to be time-based, following post-merge behaviour for consistency. From what I see in the code, it's not a critical requirement, but still it would be nice to have some consistency if it doesn't cost a lot.

> as well as how to enable them from 0

I'm still digging into this issue.

txhsl: > I may change it and adapt on the dBFT side if we decide to migrate these forks to timestamp-based behaviour, it's possible.

Will be nice to have them then, otherwise we need to keep consistency in the block-based style, if a consistency is required.

> The problem is that both Shanghai and Cancun are time-based as post-merge forks. So if we enable NeoXAMEV and NeoXEthSig in between Shanghai and Cancun, then I'd expect both of them to be time-based, following post-merge behaviour for consistency.

Cancun is the next fork we aim to, so they are in between.

AnnaShaleva: We discussed it with Roman, the final agreement is: it's possible to do the trick with time-based transition for NeoXAMEV/NeoXEthSig (although it's not that trivial as for NeoXDKG). But it seems that we don't really need NeoXAMEV/NeoXEthSig be time-based, especially given the fact that they already passed on Testnet and almost passed on Mainnet.

So let's keep NeoXAMEV and NeoXEthSig in the index-based format and only convert NeoXDKG to the time-based format.

txhsl: OK, reasonable.

txhsl:
@songb2 There are two things need to get tested.

1. Privnets work with a future `neoXDKGTime` after genesis (e.g. 3 mins after startup, with `neoXAMEV` and `neoXEthSig` enabled at Block `100`), you can get one from https://www.epochconverter.com/;
2. The new code with new `genesis_testnet.json`/`genesis_mainnet.json` can sync Testnet/Mainnet blocks to the latest correctly.

txhsl:
The new implementation can’t synchronize our Testnet in practice, should dig further or close directly.

txhsl:
https://github.com/bane-labs/go-ethereum/blob/eff7c3bda09635a3e4d4ab31025111e13f199886/core/forkid/forkid.go#L75-L98

The change of past forks changes the `ForkID` inside Eth protocol, which means a split of network.

It's OK to change the future, but not the history. I'd prefer to close this proposal.

PR #510, opened 2025-09-02
core: privnet: allow and enable `NeoXAMEVBlock` and `NeoXEthSigBlock` from 0

### LINKED ISSUE #509: Enable Privnet `NeoXAMEV` and `NeoXEthSig` from 0
# Rationale

Relate to #500. Follow #501. This is required for future updates about Privnet genesis. If we turn to timestamp-definition, then these forks have to be enabled from block `0` on Privnets.

# Implementation

Add custom V2 Extra data to genesis files and enable them from block `0`.

txhsl:
Close #509.


consensus/dbft/dbft.go
```diff
@@ -1749,7 +1749,7 @@ func (c *DBFT) verifyHeader(chain consensus.ChainHeaderReader, header *types.Hea
 		}
 
 		// Ensure that the mix digest is not zero.
-		if !isV1Extra && header.MixDigest == (common.Hash{}) {
+		if !isV1Extra && !isV2Extra && header.MixDigest == (common.Hash{}) {
```
AnnaShaleva, line 1752: Good catch.

core/genesis.go
```diff
@@ -510,10 +511,38 @@ func (g *Genesis) Commit(db ethdb.Database, triedb *triedb.Database) (*types.Blo
 			if block.MixDigest() != expected {
 				return nil, fmt.Errorf("inconsistent MixHash (NextConsensus) genesis block setting: expected %s, got %s", expected, block.MixDigest())
 			}
+		case byte(dbftutil.ExtraV1), byte(dbftutil.ExtraV2):
```
AnnaShaleva, line 508: This verification code is correct, but I have a question: valid genesis Extra should include aggregated signature built from a set of shares. Do you provide it in the new genesis config?

core/genesis.go
```diff
@@ -510,10 +511,38 @@ func (g *Genesis) Commit(db ethdb.Database, triedb *triedb.Database) (*types.Blo
...
+			switch extra[1] {
+			case byte(dbftutil.ExtraV1ECDSAScheme):
+				if len(extra) < dbftutil.HashableExtraV1Len+n*common.AddressLength+m*crypto.SignatureLength {
+					return nil, errors.New("can't start dBFT chain without validators addresses/signatures set in the genesis")
+				}
+				vals := make([]common.Address, n)
+				for i := range vals {
+					offset := dbftutil.HashableExtraV1Len + i*common.AddressLength
+					vals[i] = common.BytesToAddress(extra[offset : offset+common.AddressLength])
+				}
```
AnnaShaleva: Can we use `(Extra).ECDSASigners` for it?

core/genesis.go
```diff
@@ -510,10 +511,38 @@ func (g *Genesis) Commit(db ethdb.Database, triedb *triedb.Database) (*types.Blo
...
+			case byte(dbftutil.ExtraV1ThresholdScheme):
+				if len(extra) < dbftutil.HashableExtraV1Len+tpke.PublicKeyLen+crypto.SignatureLength {
+					return nil, errors.New("can't start dBFT chain without tpke publicky/signature set in the genesis")
+				}
+				offset := dbftutil.HashableExtraV1Len
+				pub, err := tpke.NewPublicKeyFromBytes(extra[offset : offset+tpke.PublicKeyLen])
+				if err != nil {
+					return nil, fmt.Errorf("failed to decode public key: %w", err)
+				}
```
AnnaShaleva: And `(Extra).ThresholdSigners` for it?

core/genesis.go
AnnaShaleva: My bad, we can't do that, empty signatures are OK.

We actually use empty signatures in genesis block for Meinnet/Testnet.

core/genesis.go
txhsl: Just for record, when using `(Extra).ThresholdSigners`, we need an infinite BLS12381 G2 point `0xc00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000` as the signature, instead of a zero array, otherwise it returns an error.

privnet/four/genesis_privnet.json
```diff
@@ -31,6 +31,8 @@
   },
   "difficulty": "1",
   "gasLimit": "30000000",
+  "extraData": "0x0201ce2c47e402f65838a2251f779218a3e69debb1236a1099d6fb7e65c04a7065f7995f92eb4a60bd7425546c255a720638c455ded5ac03fb933bfee41d638404d4e1834f160c51b6c3e036210bd241018dc00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
+  "mixHash": "0x953fe70afae6653830cde4febd89dfbfc6b4965fc0e24df4ed8a588bfd11b78d",
```
AnnaShaleva, line 35: Did you test it, does it work? I'm still having troubles with privnet run.

AnnaShaleva (approved):
If tested, then LGTM.

core/genesis.go
AnnaShaleva: It's OK, because Extra format matches exactly the one that is used for post-NeoXAMEV blocks.

privnet/four/genesis_privnet.json
txhsl: It works to me. If you still have problems, please let me know. 

Also, I will let @songb2 have a try on this as well.

Issue #517, opened 2025-10-09
Failed chain freezer on Privnet


txhsl:
#### System information

Geth version: `0.5.0-unstable`
OS & Version: Linux
Commit hash : f20d5b71ddc24d770a682509debe836eb0c9c318

#### Expected behaviour

We expect the `v1.14.13` code from #487 to work well.

#### Actual behaviour

Ref https://github.com/bane-labs/go-ethereum/pull/499#issuecomment-3337991203.

#### Steps to reproduce the behaviour

1. Make `geth`;
2. Start up a local Privnet, e.g. the `four-node` one;
3. Check the log of `node5`.

#### Backtrace

```
INFO [09-26|18:11:40.555] Rebuilt trie database                    root=e9f7e9..bb34fc
INFO [09-26|18:11:40.556] Rebuilding state snapshot
INFO [09-26|18:11:40.558] Committed new head block                 number=1 hash=3c8e71..ca5f41
INFO [09-26|18:11:40.568] Resuming state snapshot generation       root=e9f7e9..bb34fc accounts=0          slots=0             storage=0.00B    dangling=0 elapsed=4.975ms
WARN [09-26|18:11:40.655] Inserted known block                     number=3 hash=6328b2..17b21e uncles=0 txs=0 gas=0 root=e9f7e9..bb34fc
WARN [09-26|18:11:40.659] Inserted known block                     number=4 hash=f61a81..1966b0 uncles=0 txs=0 gas=0 root=e9f7e9..bb34fc
WARN [09-26|18:11:40.662] Inserted known block                     number=5 hash=c1472f..9a9c26 uncles=0 txs=0 gas=0 root=e9f7e9..bb34fc
WARN [09-26|18:11:40.669] Inserted known block                     number=6 hash=0f0c5a..7faac9 uncles=0 txs=0 gas=0 root=e9f7e9..bb34fc
WARN [09-26|18:11:40.679] Inserted known block                     number=7 hash=6e5248..a9a351 uncles=0 txs=0 gas=0 root=e9f7e9..bb34fc
WARN [09-26|18:11:40.691] Inserted known block                     number=8 hash=097928..ebbcbb uncles=0 txs=0 gas=0 root=e9f7e9..bb34fc
WARN [09-26|18:11:40.701] Inserted known block                     number=9 hash=5ac134..5929d9 uncles=0 txs=0 gas=0 root=e9f7e9..bb34fc
[...]

Issue #507, opened 2025-09-01
Refactor to properly support post-merge behaviour


qianhh:
# Rationale

During v1.14, Ethereum perform “The Merge” to split the original Geth into two parts, a Consensus Layer (CL) and an Execution Layer (EL). Since then, the Geth removed its functionality in PoW/PoA maintenance step by step. Instead, it needs a connection to CL to download new blocks, and the Geth client itself works more like a decentralized database without any possibility of inconsistency.

In v1.14, they implemented the following things and performed “The Merge”.

1. Add configurations about `TerminalTotalDifficulty` and `MergeNetsplitBlock` to switch from PoW to PoS block handling;
2. Implement `eth/catalyst` for eth1/eth2 RPC integration;
3. Disable the automatic block sealing of internal miner worker;
4. Separate the blob data storage and pruning to CL.

After the Merge, they implemented the following things to simplify block handling.

1. Remove reorg handling in local chain data maintenance, as well as the total difficulty data from database;
2. Simplify the check of PoW/PoS switch;
3. Remove the block mining broadcast and total difficulty information from network protocol;
4. Support pruned-able chain data and cutoff operations with configurable checkpoints.

Now in v1.16, they implemented the following things to reduce EL resource usage.

1. Use `ETH69` range packets instead of `ETH67`/`ETH68` single block packets to simplify network communication.

# Implementation

Ref: https://github.com/bane-labs/go-ethereum/pull/487#discussion_r2273467865.

We need to refactor miner/dBFT to properly support post-merge behaviour, otherwise we have to face a road division from Geth upstream, their new features are becoming more unsuitable in our use case (a pre-merge blockchain).

[1940bd3](https://github.com/bane-labs/go-ethereum/commit/1940bd3a48c1bfc62a319169b1e2cdab68ec51ae) uses dBFT as the CL to receive finialized block information, but maybe no enough.

txhsl:
#520

txhsl:
Well, the [structure](https://github.com/protolambda/zrnt/blob/9eed4d074b95d955dff7490ccbde1ba1909c1619/eth2/beacon/electra/block.go#L115-L135) of electra beacon blocks is far more than we need. IMO, it's better to keep using Eth1 blocks.

PR #487, opened 2025-06-27
Merge Geth [v1.14.0...v1.14.13]

qianhh:
This PR merges [v1.14.0...v1.14.13](https://github.com/ethereum/go-ethereum/compare/v1.13.15...v1.14.13) of Geth into bane-main and aimed to close https://github.com/bane-labs/go-ethereum/issues/233.

Progress

- [X] v1.14.0
- [X] v1.14.1...v1.14.11
- [X] v1.14.12...v1.14.13
- [X]  The node functions normally and consensus runs fine (1/4/7 privnet).
- [X]  Mainnet/T4 sync is OK.

The list of important PRs that were merged. @AnnaShaleva, @txhsl, please take a look at these PRs before the review.

- https://github.com/ethereum/go-ethereum/pull/28623
- https://github.com/ethereum/go-ethereum/pull/29169
- https://github.com/ethereum/go-ethereum/pull/29482
- https://github.com/ethereum/go-ethereum/pull/29338
- https://github.com/ethereum/go-ethereum/pull/29465
- https://github.com/ethereum/go-ethereum/pull/29431
- https://github.com/ethereum/go-ethereum/pull/27720
- https://github.com/ethereum/go-ethereum/pull/30069
- https://github.com/ethereum/go-ethereum/pull/29711
- https://github.com/ethereum/go-ethereum/pull/30425
- https://github.com/ethereum/go-ethereum/pull/30571
- https://github.com/ethereum/go-ethereum/pull/30601
- https://github.com/ethereum/go-ethereum/pull/30645
- https://github.com/ethereum/go-ethereum/pull/30534
- https://github.com/ethereum/go-ethereum/pull/29441
- https://github.com/ethereum/go-ethereum/pull/30609

Merged but Reverted

- https://github.com/ethereum/go-ethereum/pull/28623 I reverted a portion.
- https://github.com/ethereum/go-ethereum/pull/29169 I reverted a portion.
- https://github.com/ethereum/go-ethereum/pull/29381
- https://github.com/ethereum/go-ethereum/pull/29281
- https://github.com/ethereum/go-ethereum/pull/29179
- https://github.com/ethereum/go-ethereum/pull/30488
- https://github.com/ethereum/go-ethereum/pull/30642
- https://github.com/ethereum/go-ethereum/pull/30737 I reverted a portion.
- https://github.com/ethereum/go-ethereum/pull/29895

Ref https://github.com/bnb-chain/bsc/pull/2761.

PR #518, opened 2025-10-10
*: revert customized behaviors about finalized block

txhsl:
Related to #507 and #517.

#487 introduced a consensus engine change to publish finalized blocks locally to allow the blob finalization and avoid the usage of `terminalTotalDifficulty`. This enabled the chain freezer before the merge block and the TTD.

Considering the influence to future work on #507, and the possibility of enabling other post-merge behaviours unexpectedly, we've discussed and decided to move the refactor to later upgrades, e.g. a scheduled TTD with #507, or for further discussion.

@qianhh Let's first focus on the code merge from upstream.

### DIFF
--- a/eth/ethconfig/config.go
+++ b/eth/ethconfig/config.go
@@ -197,6 +197,7 @@ type Config struct {
 // Clique is allowed for now to live standalone, but ethash is forbidden and can
 // only exist on already merged networks.
 func CreateConsensusEngine(config *params.ChainConfig, db ethdb.Database, statisticsCfg dbft.StatisticsConfig) (consensus.Engine, error) {
+	// TODO: config TTD correctly and move this check back after Clique
 	if config.DBFT != nil {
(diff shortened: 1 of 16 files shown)


eth/ethconfig/config.go
```diff
@@ -187,6 +187,7 @@ type Config struct {
 // Clique is allowed for now to live standalone, but ethash is forbidden and can
 // only exist on already merged networks.
 func CreateConsensusEngine(config *params.ChainConfig, db ethdb.Database, statisticsCfg dbft.StatisticsConfig) (consensus.Engine, error) {
+	// TODO: config TTD correctly and move this check back after Clique
 	if config.DBFT != nil {
```
txhsl, line 201: I preserved this change from 1940bd3a48c1bfc62a319169b1e2cdab68ec51ae, so that we don't have to set TTD for now. But IMO, we need to handle this correctly in the future.

qianhh (approved):

PR #526, opened 2025-10-28
core: implement EL file system for blob storage

### LINKED ISSUE #520: Persist and prune blob data correctly
# Rationale

Ref https://github.com/bane-labs/go-ethereum/pull/499#issuecomment-3410502078, the current Geth implementation only handles the metadata of blob transaction and leaves the persistence of blob data to consensus layer.

The lifecycle of EIP-4844 should be like:
1. Blob transactions are sent to EL blob pool;
2. Blob metadata and sidecars are committed to CL;
3. CL confirms the blob transaction and store the data;
4. EL deletes the temporary storage for blob transactions after CL finalized (about 12.8 mins, 2 epochs of 32 blocks);
5. CL prunes the file storage for blob data after expiration (about 18.2 days, 4096 epochs of 32 blocks).

It is supposed that the CL maintains the final storage. Since we don't have a corresponding consensus layer (it's a separated program definition), our client only packs the blob transactions into blocks but evicts the blob data after a short period of time (the billy database is not designed for long-term storage).

# Implementation

Relate to #233, #496 and #507.

The blob transaction is handled, yes. But we need to store the blob data as well, if we want to support the functionality of EIP-4844 correctly.

qianhh:
Close #520.
I will proceed with the following 3 steps to achieve it one by one.
- [x] Refer to the [BSC](https://github.com/bnb-chain/bsc) implementation of storing and delayed release of blob data.
- [x] An independent protocol to transfer blobs optionally.
- [x] A new database to store the blobs, refer [prysm](https://github.com/OffchainLabs/prysm).


qianhh:
Rebased to bane-main.

eth/protocols/neox/protocol.go
```diff
@@ -0,0 +1,119 @@
...
+)
+
+// Constants to match up protocol versions and messages
+const (
+	Neox1 = 1
+)
+
+// ProtocolName is the official short name of the `neox` protocol used during
+// devp2p capability negotiation.
+const ProtocolName = "neox"
```
txhsl: Name it to `blob` or something similar, at least `neox` is not a good choice.

qianhh: Since the block data message distribution will also be included in the `neox` protocol later, the name `neox` was chosen. It would be even better if there were other more preferable options.

core/chain_makers.go
```diff
@@ -616,6 +647,13 @@ func (cm *chainMaker) makeHeader(parent *types.Block, state *state.StateDB, engi
 		header.ExcessBlobGas = &excessBlobGas
 		header.BlobGasUsed = new(uint64)
 		header.ParentBeaconRoot = new(common.Hash)
+
+		if cm.config.DBFT != nil {
```
txhsl: Can we not limit this to dBFT? I mean, a general handling will be better.

core/data_availability.go
```diff
@@ -0,0 +1,160 @@
...
+		return fmt.Errorf("invalid number of %d blob commitments compared to %d blob hashes", len(sidecar.Commitments), len(hashes))
+	}
+	if len(sidecar.Proofs) != len(hashes) {
+		return fmt.Errorf("invalid number of %d blob proofs compared to %d blob hashes", len(sidecar.Proofs), len(hashes))
+	}
+	// Blob quantities match up, validate that the provers match with the
+	// transaction hash before getting to the cryptography
+	hasher := sha256.New()
+	for i, vhash := range hashes {
+		computed := kzg4844.CalcBlobHashV1(hasher, &sidecar.Commitments[i])
```
txhsl: > // validateBlobSidecar it is same as validateBlobSidecar in core/txpool/validation.go

A bit different, maybe use `sidecar.ValidateBlobCommitmentHashes()` instead.

Or, since this is in the `core` package as well, we can make `txpool.validateBlobSidecar()` public and reuse the code.

core/data_availability.go
```diff
@@ -0,0 +1,160 @@
...
+	for i := range sidecar.Blobs {
+		if err := kzg4844.VerifyBlobProof(&sidecar.Blobs[i], sidecar.Commitments[i], sidecar.Proofs[i]); err != nil {
+			return fmt.Errorf("invalid blob %d: %v", i, err)
+		}
+	}
+	return nil
+}
+
+// IsDataAvailable it checks that the blobTx block has available blob data
+func IsDataAvailable(chain consensus.ChainHeaderReader, block *types.Block) (err error) {
```
txhsl: A method named by `IsXXX()` needs to return `bool` or `(bool, error)`. I prefer `verifyBlockBlobSidecars()` or `validateBlockBlobSidecars()` here.

core/data_availability.go
```diff
@@ -0,0 +1,160 @@
...
+	// blobs themselves via KZG
+	for i := range sidecar.Blobs {
+		if err := kzg4844.VerifyBlobProof(&sidecar.Blobs[i], sidecar.Commitments[i], sidecar.Proofs[i]); err != nil {
+			return fmt.Errorf("invalid blob %d: %v", i, err)
+		}
+	}
+	return nil
+}
+
+// IsDataAvailable it checks that the blobTx block has available blob data
```
txhsl: The word `it` needs to be removed from comments, and please end with `.`

core/data_availability.go
```diff
@@ -0,0 +1,160 @@
...
+func IsDataAvailable(chain consensus.ChainHeaderReader, block *types.Block) (err error) {
+	defer func(start time.Time) {
+		daCheckTimer.Update(time.Since(start))
+	}(time.Now())
+
+	// refer logic in ValidateBody
```
txhsl:
```suggestion
	// Refer logic in ValidateBody.
```

Start with capital letters and end with `.`

core/data_availability.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+	if block.Time()+params.MinTimeDurationForBlobRequests < highest.Time {
+		// if we needn't check DA of this block, just clean it
+		block.CleanSidecars()
+		return nil
+	}
+
+	// if sidecar is nil, just clean it. And it will be used for saving in ancient.
+	if block.Sidecars() == nil {
+		block.CleanSidecars()
```
txhsl: Well, here if `block.Sidecars() == nil`, then we set `block.sidecars = make(BlobSidecars, 0)`. IMO it's not a clean but an initialization.

core/data_availability.go
```diff
@@ -0,0 +1,160 @@
...
+
+	// only required to check within MinTimeDurationForBlobRequests seconds's DA
+	highest := chain.ChasingHead()
+	current := chain.CurrentHeader()
+	if highest == nil || highest.Number.Cmp(current.Number) < 0 {
+		highest = current
+	}
+	if block.Time()+params.MinTimeDurationForBlobRequests < highest.Time {
+		// if we needn't check DA of this block, just clean it
+		block.CleanSidecars()
```
txhsl: The `CleanSidecars()` operation seems not preferred to happen in a validation method. If the blob sidecar has no need to check, then we only return `nil`.

Can we move it to a separated pre-handling method if `CleanSidecars()` is necessary?

core/data_availability.go
```diff
@@ -0,0 +1,160 @@
...
+		}
+		if err := validateBlobSidecar(tx.BlobHashes(), sidecars[i]); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func CheckDataAvailableInBatch(chainReader consensus.ChainHeaderReader, chain types.Blocks) (int, error) {
```
txhsl: When calling this by `core.CheckDataAvailableInBatch()`, it's really hard to understand what is the method doing.

This file name should be something like `validation.go`, since these new methods doesn't serve any struct, or move them to `blockchain.go`.

core/types/block.go
```diff
@@ -237,6 +237,9 @@ type Block struct {
 	// inter-peer block relay.
 	ReceivedAt   time.Time
 	ReceivedFrom interface{}
+
+	// sidecars provides DA check
+	sidecars BlobSidecars
```
txhsl: We have `Payload` defined in `miner` to carry this.

core/data_availability.go
qianhh: `CleanSidecars()` has been deleted.

core/data_availability.go
qianhh: `CheckDataAvailableInBatch()` has been removed.

txhsl:
@qianhh I've uploaded my proposal to https://github.com/bane-labs/go-ethereum/tree/blob-protocol. Just an idea, we need further development, but try to rebase onto it if possible.

txhsl:
@qianhh Pushed some fixes, and you may need to check them.

Besides, I always get `TempDir RemoveAll cleanup: remove XXXX: The process cannot access the file because it is being used by another process.` errors in new UTs, e.g. `TestFreezer_ResetTailMeta_WithAdditionTable`, `TestFreezer_ResetTailMeta_EmptyTable` and `TestDBFTBlobFeeCost`. It seems only happen on Windows, but I wish we can pay attention to potential risks.

qianhh:
All functions have been completed. The default retention time for blob data is 8192 epochs, with each epoch consisting of 32 blocks, which amounts to 262,144 blocks.

Four new flags have been added:

1. `--blob-path` Location for blob storage. Default location will be a 'blobs' directory next to the datadir.
2. `--blob-storage-layout` Default is 'by-epoch'. Only support 'by-epoch' layout now.
3. `--blob-retention-epochs` Override the default blob retention period (measured in epochs). The node will exit with an error at startup if the value is less than the default of 8192 epochs.
4. `--blob-save-fsync` Forces new blob files to be fysnc'd before continuing, ensuring durable blob writes.
5. `--blob-sync` Enable blob sidecar synchronization alongside block synchronization.

Two new RPC APIs have been added:

1.`eth_getBlobSidecars` Obtain all the BlobSidecars of a certain block.
2.`eth_getBlobSidecarByTxHash` Obtain the BlobSidecar of a certain transaction.


@songb2 Testing can now begin. @txhsl If you have time, please take a moment to review the code and check if there are any omissions.

txhsl:
> @qianhh I've uploaded my proposal to https://github.com/bane-labs/go-ethereum/tree/blob-protocol. Just an idea, we need further development, but try to rebase onto it if possible.

Moved to https://github.com/bane-labs/go-ethereum/tree/beacon-protocol, currently working on the downloader.

cmd/utils/storage_options.go
```diff
@@ -0,0 +1,169 @@
...
+
+var (
+	BlobStoragePathFlag = &cli.PathFlag{
+		Name:     "blob-path",
+		Usage:    "Location for blob storage. Default location will be a 'blobs' directory next to the datadir.",
+		Category: flags.BlobStorageCategory,
+	}
+	BlobStorageLayout = &cli.StringFlag{
+		Name:        "blob-storage-layout",
+		Usage:       layoutFlagUsage(),
```
txhsl: Flat text is preferred. Consider replacing the `layoutFlagUsage()` with its direct content if the method is never reused somewhere else.

cmd/utils/storage_options.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+func validateLayoutFlag(_ *cli.Context, v string) error {
+	if slices.Contains(filesystem.LayoutNames, v) {
+		return nil
+	}
+	return errors.Errorf("invalid value '%s' for flag --%s, %s", v, BlobStorageLayout.Name, layoutOptions())
+}
+
+// SetBlobStorageConfig sets the blob storage configuration options in the node config
```
txhsl: Comment ends with `.`

cmd/utils/flags_blobstorage.go
```diff
@@ -0,0 +1,169 @@
...
+func blobStoragePath(c *cli.Context) string {
+	blobsPath := c.Path(BlobStoragePathFlag.Name)
+	if blobsPath == "" {
+		// append a "blobs" subdir to the end of the data dir path
+		blobsPath = filepath.Join(c.String(DataDirFlag.Name), "blobs")
+	}
+	return blobsPath
+}
+
+var errInvalidBlobRetentionEpochs = errors.New("value is smaller than spec minimum")
```
txhsl, line 151: It's rare to see an independent error definition here in flags and configuration.

Only about the style, consider using the direct error text as well.

cmd/utils/flags_blobstorage.go
```diff
@@ -0,0 +1,169 @@
...
+	// BlobSaveFsync enforces durable filesystem writes for use cases where blob availability is critical.
+	BlobSaveFsync = &cli.BoolFlag{
+		Name:     "blob-save-fsync",
+		Usage:    "Forces new blob files to be fysnc'd before continuing, ensuring durable blob writes.",
+		Category: flags.BlobStorageCategory,
+	}
+)
+
+// BlobStorageFlags is the list of CLI flags for configuring blob storage.
+var BlobStorageFlags = []cli.Flag{
```
txhsl, line 55: Since we define `BlobStorageFlags` here, can we rename this file to `flag_blobstorage`?

cmd/utils/flags_blobstorage_test.go
```diff
@@ -0,0 +1,163 @@
...
+
+	"github.com/ethereum/go-ethereum/core/filesystem"
+	"github.com/ethereum/go-ethereum/core/filesystem/primitives"
+	"github.com/ethereum/go-ethereum/node"
+	"github.com/ethereum/go-ethereum/params"
+	"github.com/stretchr/testify/require"
+	"github.com/urfave/cli/v2"
+)
+
+func TestBlobStoragePath_NoFlagSpecified(t *testing.T) {
```
txhsl, line 21: This fails on Windows.
```
=== RUN   TestBlobStoragePath_NoFlagSpecified
    storage_options_test.go:28:
                Error Trace:    D:/Bane/go-ethereum/cmd/utils/storage_options_test.go:28
                Error:          Not equal:
                                expected: "C:\\Users\\79949\\AppData\\Local\\Ethereum/blobs"
                                actual  : "C:\\Users\\79949\\AppData\\Local\\Ethereum\\blobs"

                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1 +1 @@
                                -C:\Users\79949\AppData\Local\Ethereum/blobs
                                +C:\Users\79949\AppData\Local\Ethereum\blobs
                Test:           TestBlobStoragePath_NoFlagSpecified
--- FAIL: TestBlobStoragePath_NoFlagSpecified (0.00s)
```

consensus/dbft/blockqueue.go
```diff
@@ -87,5 +93,12 @@ func (bq *blockQueue) PutBlock(b *types.Block, state *state.StateDB, receipts []
 	// Broadcast the block and announce chain insertion event
 	bq.mux.Post(core.NewMinedBlockEvent{Block: b})
 
+	sealHash := WorkerSealHash(b.Header())
```
txhsl: Seems like my code, but we'd better ensure the local insertion before broadcast.

So move this ahead of `bq.mux.Post`, and let it return the error if not nil.

consensus/dbft/blockqueue.go
```diff
@@ -87,5 +93,12 @@ func (bq *blockQueue) PutBlock(b *types.Block, state *state.StateDB, receipts []
...
+	err = bq.fs.CommitSealBlockHash(sealHash, b)
+	if err != nil {
+		// Due to the internal synchronization issue of dbft, it may receive and adopt block from other nodes,
+		// thereby changing the sealHash. Therefore, it is possible to return the error "errSealingHashNotFound".
```
txhsl: No, here we can commit the block/header/hash without its seal hash. Now I'm sure that reorg never happens in dBFT itself, and the block here is a signed full block.

The reorg is caused by block broadcast/announcement, from the subprotocol or block fetcher, sent by other nodes.

If we can handle reorgs in FS, then commit the final hash here.

core/file_system.go
```diff
@@ -0,0 +1,290 @@
...
+func (fs *FileSystem) loop() {
+	// Subscribe to chain head events to trigger subpool resets
+	var (
+		newHeadCh  = make(chan ChainHeadEvent)
+		newHeadSub = fs.bc.SubscribeChainHeadEvent(newHeadCh)
+	)
+	defer newHeadSub.Unsubscribe()
+
+	for range newHeadCh {
+		err := fs.finalize()
```
txhsl: This works, absolutely, but it's better if we pass in a finalized height or a finalized header as the parameter (to make the time cost predictable and stable), not from the chain head subscription. The finalize job seems quite light, maybe we don't need an independent routine to do this.

The ideal implementation is to decouple the `BlockChain` (L34) from the FS, if possible.

core/file_system.go
```diff
@@ -0,0 +1,290 @@
...
+
+// FileSystem manages blob sidecars associated with blocks.
+type FileSystem struct {
+	bc          *BlockChain
+	blobStorage *filesystem.BlobStorage
+
+	pendingBlobs map[common.Hash]*SealBlobs
+	pendingMu    sync.Mutex
+
+	blobFeed       event.Feed
```
txhsl: Let's make it a pull-on-demond protocol, which means requesters only try to fetch and download after receiving related block announcement, and never broadcast the full blob data in the network.

dBFT ensures that the validators' blob pool have synced all the blob sidecars. So when dBFT commits the block, blobs are also ready for access, which meets the above rule.

core/file_system.go
```diff
@@ -0,0 +1,290 @@
...
+	reannoBlobFeed event.Feed
```
txhsl: A reannouncement maybe not necessary. I checked the CL spec, the requirement is only accepting and announcing a block after the related blob sidecars are all downloaded and verified.

It means, when a node announced that it has reached the latest height, then it should have synced all blob data and can be used as the target to fetch from.

core/file_system.go
```diff
@@ -0,0 +1,290 @@
...
+	log.Trace("Seal blobs pending", "number", blockNumber, "sealHash", sealHash, "blobs", len(blobs))
+	return nil
+}
+
+// CommitSealBlockHash commits the sealed block hash associated with committed blobs.
+func (fs *FileSystem) CommitSealBlockHash(sealHash common.Hash, block *types.Block) error {
+	log.Trace("Commit seal blobs...", "number", block.Number(), "sealHash", sealHash, "hash", block.Hash(), "txs", len(block.Transactions()))
+
+	fs.pendingMu.Lock()
+	sealBlobs, ok := fs.pendingBlobs[sealHash]
```
txhsl: 🤔OK, the seal hash helps.

params/protocol_params.go
```diff
@@ -175,6 +175,13 @@ const (
 	BlobTxPointEvaluationPrecompileGas = 50000   // Gas price for the point evaluation precompile.
 
 	HistoryServeWindow = 8192 // Number of blocks to serve historical block hashes for, EIP-2935.
+
+	BlocksPerEpoch = 32 // Number of blocks in each epoch.
```
txhsl: We have another `epoch` definition in our Governance (60480 blocks), ref https://xdocs.ngd.network/governance/governance-in-neo-x.

It's necessary to add a comment here to clarify the difference (this is a mock parameter simulate Eth PoS), or change the expression.

consensus/dbft/blockqueue.go
txhsl: I think I forgot my original proposal as well. The seal hash is used to keep blob data out of the consensus, they should be committed by the miner, and get confirmed here.

Seal hash is a consist value before and after consensus, while the honest hash changes when the block signature is filled in the block extra. So if we use the honest hash, we will not be able to find the committed blobs.

Nothing about the reorg.

consensus/dbft/blockqueue.go
qianhh: It is placed after block broadcast/announcement because the function `bq.fs.CommitSealBlockHash(sealHash, b)` may return `errSealingHashNotFound`, but the block itself is not problematic, so there is no need to return an error, right? The missing sidecars can be obtained from other peers, so this error can actually be ignored. In 7-CNs, this error occurs more frequently.

consensus/dbft/blockqueue.go
qianhh: Yes, using the seal hash is necessary to distinguish between blocks.

core/file_system.go
qianhh: There is indeed room for improvement here. Currently, there are quite a few sidecars in the cache.

consensus/dbft/blockqueue.go
txhsl: Yeh... Only the primary node will 100% find the cache here, because it is its proposal been sealed as a block. For other CNs, they may also propose blocks but the hash will be different (due to the block nonce difference). So you will always hit this error.

The key point for other CNs is to find blobs from the local mempool and rebuild the sidecar (so that they can insert blobs to local fs before announcing the block), because currently we don't transfer blobs directly in dBFT.

Since our CL can still access the EL state, our consensus can still believe in the txpool verification. When dBFT enforces the blob tx synchronization, it also contains a trust that related blobs are included in the blob pool for further insertion.

Or do you think we need some info about blobs to be included in the block header? `versionedHashes` is something that may helps above local rebuild, but IMO it doesn't enhance trust.

txhsl: The cache is ... a short cut, but if we can't hit it frequently, we can remove it. E.g. I've removed the sealing task cache in our miner, there's also a block insertion.

qianhh: Good idea! If we directly obtain the transactions from the txpool and then rebuild the sidecars, there will be no need to perform the cache first. I'll try to implement it.

qianhh:
There are still some issues with synchronization, and they are currently being addressed.

qianhh:
It's ready now.

txhsl:
Maybe we need more consensus verifications about EIP-4844 in dBFT, let me check.

OK, seems no need, the maximum blob number per block is verified in header indirectly, through maximum blob gas per block.

eth/protocols/beacon/protocol.go
```diff
@@ -25,15 +25,21 @@ var ProtocolVersions = []uint{BEACON1}
...
 const maxMessageSize = 10 * 1024 * 1024
 
 const (
-	StatusMsg         = 0x00
-	NewBlockHashesMsg = 0x01
-	NewBlockMsg       = 0x02
+	StatusMsg         = 0x00 // Status message
+	NewBlockHashesMsg = 0x01 // New block hashes message
+	NewBlockMsg       = 0x02 // New block message
+	NewBlobsMsg       = 0x03 // New blobs message
```
txhsl: So there's still the raw blob broadcast? Can we only rely on the block hash announcement and the sidecar fetcher?

eth/protocols/beacon/protocol.go
```diff
@@ -25,15 +25,21 @@ var ProtocolVersions = []uint{BEACON1}
...
+	NewBlobsRootMsg   = 0x04 // New blobs root message
```
txhsl: Seems redundant with the block hash announcement `NewBlockHashesMsg`. Is this message special, e.g. only used when there is a blob in the new block?

And if we can include a batch of block hashes in the same message, why not extend it?

eth/protocols/beacon/protocol.go
qianhh: The `NewBlobsMsg` message is not being broadcasted now, and I remove it later.

eth/protocols/beacon/protocol.go
qianhh: Yes, this block hash will only be broadcasted when the block contains blob data.

beacon/impl/fetcher/sidecar_fetcher.go
```diff
@@ -0,0 +1,333 @@
...
+func (f *SidecarFetcher) loop() {
+	// Iterate the sidecar fetching until a quit is requested
+	var fetchTimer = time.NewTimer(0)
+	<-fetchTimer.C // clear out the channel
+	defer fetchTimer.Stop()
+	timeoutTicker := time.NewTicker(fetchTimeout)
+	defer timeoutTicker.Stop()
+
+	height := f.chain.CurrentBlock().Number.Uint64()
+	fetchHeight := min(0, height-uint64(params.MinEthEpochsForBlobsSidecarsRequest*params.BlocksPerEthEpoch))
```
txhsl: `min` or `max`? And this is an `uint64` decrease, care the underflow, the result seems always `0`.

beacon/impl/fetcher/sidecar_fetcher.go
```diff
@@ -0,0 +1,333 @@
...
+		case hash := <-f.done:
+			// A pending import finished, remove all traces of the notification
+			f.forgetHash(hash)
+
+		case <-timeoutTicker.C:
+			// Periodically clean out any abandoned fetches
+			earliestHeight := fetchHeight
+			for hash, task := range f.fetching {
+				cost := time.Since(task.time)
+				if cost > 3*fetchTimeout {
```
txhsl: Why is `3*fetchTimeout`, maybe need a comment.

beacon/impl/fetcher/sidecar_fetcher.go
```diff
@@ -0,0 +1,333 @@
...
+					f.forgetHash(hash)
+					if f.chain.GetBlockByHash(hash) == nil {
+						earliestHeight = min(earliestHeight, task.block.NumberU64()-1)
```
txhsl, line 212: Still `min`? Maybe I misunderstand this.

beacon/impl/fetcher/sidecar_fetcher.go
```diff
@@ -0,0 +1,333 @@
...
+						continue
+					} else if !f.fs.ShouldRetain(task.block.Number()) {
+						continue
+					}
+					// Deep-fetch sidecars if normal fetch failed multiple times
+					go func() {
+						if sidecars, err := f.deepFetchSidecars(hash); err != nil {
```
txhsl: If a node announces the blob but can not been fetched, then maybe drop the peer with `dropPeer`? Any other reason to retry with the deep fetch?

beacon/impl/fetcher/sidecar_fetcher.go
```diff
@@ -0,0 +1,333 @@
...
+			request := make(map[string][]common.Hash)
+			reqTaskMap := make(map[string][]*fetchingTask)
+
+			peers := f.filterPeers()
+			if len(peers) > 0 {
+				for hash, task := range f.scheduled {
+					timeout := arriveTimeout - gatherSlack
+					if time.Since(task.time) > timeout {
+						// Pick a random peer to retrieve from, reset all others
+						peer := peers[rand.Intn(len(peers))]
```
txhsl: The question is, maybe there is a node doesn't enable dBFT or enable the blob storage. It should never announce new blobs, and IMO the fetcher should not try to fetch from it.

But I see there‘s not an enforcement to fetch all the blobs otherwise reject the block. It seems not avoidable.

beacon/impl/fetcher/sidecar_fetcher.go
```diff
@@ -0,0 +1,333 @@
...
+
+		defer func() { f.done <- block.Hash() }()
+
+		// Run the actual import and log any issues
+		if err := f.fs.InsertBlobsWithoutValidation(block.Header(), sidecars); err != nil {
+			log.Debug("Propagated sidecars import failed", "peer", peer, "number", block.Number(), "hash", block.Hash(), "err", err)
+			return
+		}
+		// If import succeeded, broadcast the sidecars
+		go f.broadcastSidecars(block.Hash(), sidecars, false)
```
txhsl: Maybe announce instead?

beacon/impl/fetcher/sidecar_fetcher.go
```diff
@@ -0,0 +1,333 @@
...
+
+// filterPeers filters out peers that have not been able to provide blob sidecars recently.
+func (f *SidecarFetcher) filterPeers() []string {
+	peers := f.allBeacons()
+
+	//Filter out peers that have not been able to obtain the blob recently.
+	now := time.Now()
+	filtered := make([]string, 0, len(peers))
+	for _, peer := range peers {
+		if timeout, ok := f.noBlobPeers[peer]; ok {
```
txhsl: The peer maintaining is better to be handled in the protocol handler, which is `eth/handler_beacon.go`.

core/events.go
```diff
@@ -36,3 +37,8 @@ type ChainEvent struct {
 type ChainHeadEvent struct {
 	Header *types.Header
 }
+
+type BlobEvent struct {
+	BlockHash common.Hash
+	Sidecars  types.BlobSidecars
```
txhsl: Still looks like a full broadcast. I have to remind the sidecar is heavy, and we need to avoid passing them around the system if not necessary.

miner/miner.go
```diff
@@ -74,6 +75,7 @@ type Miner struct {
 	chain       *core.BlockChain
 	pending     *pending
 	pendingMu   sync.Mutex // Lock protects the pending block
+	fs          *core.FileSystem
```
txhsl: Seems not used.

beacon/impl/fetcher/sidecar_fetcher.go
qianhh: Well, there's a mistake here. I will fix it.

beacon/impl/fetcher/sidecar_fetcher.go
qianhh: Because the other nodes connected to the target node might not have enabled blob data synchronization, there is no guarantee that the blob announcement will be received. Therefore, I adopted the approach of synchronizing with all the connected nodes. If the retry attempt fails twice, the last retry will use deep fetch to relay and obtain the data, as this may cause a significant amount of network load. Hence, it should not be used frequently.

When a node that has enabled blob data synchronization receives a blob notification message and fails to retrieve the blob locally, it will actively initiate a data fetch operation from that node and save it locally.

beacon/impl/fetcher/sidecar_fetcher.go
qianhh: If multiple blocks are missing, the minimum height among them should be taken.

beacon/impl/fetcher/sidecar_fetcher.go
qianhh: Don't worry. If that node disables the blob data synchronization, it will return an empty result. On my end, it will be added to the "noBlobPeers" list for a certain period of isolation, and failed tasks will be retried.

beacon/impl/fetcher/sidecar_fetcher.go
qianhh: In fact, this is just "announce". I will rename it to make it more readable.

beacon/impl/fetcher/sidecar_fetcher.go
txhsl: > Because the other nodes connected to the target node might not have enabled blob data synchronization, there is no guarantee that the blob announcement will be received.

Well, this is annoying.

txhsl: > Therefore, I adopted the approach of synchronizing with all the connected nodes.

Will it be better if we make blob protocol as a sub protocol of `beacon`, like the `snap` protocol of `eth`, and also enforce the available time of blobs in the network as Ethereum 2.0 as a protocol parameter, instead of a customizable configuration.

Then nodes in this sub protocol can be sure that other nodes should receive blobs and have all the blobs, and can drop the bad node with an easy judgement. The network traffic can be reduced as well.

After some discussion, let's first stick to the current implementation.

beacon/impl/fetcher/sidecar_fetcher.go
qianhh: It has been migrated.

core/events.go
qianhh: Renamed.

beacon/impl/fetcher/sidecar_fetcher.go
qianhh: Relevant explanations have been added.

beacon/impl/fetcher/sidecar_fetcher.go
txhsl: > Well, there's a mistake here. I will fix it.

Still strange, no need for a `else {0}`?

qianhh: The default value is 0.

txhsl: OK, I see.

eth/protocols/beacon/broadcast.go
```diff
@@ -37,3 +37,20 @@ func (p *Peer) broadcastBlocks() {
 		}
 	}
 }
+
+// broadcastBlockBlob is a write loop that schedules blob broadcasts
+// to the remote peer. The goal is to have an async writer that does not lock up
+// node internals and at the same time rate limits queued data.
+func (p *Peer) broadcastBlockBlob() {
```
txhsl: Can we rename related things to `announce`, including the method and channel? Or rename `blockBlob` to `blobRoot`, maybe the later.

miner/miner.go
```diff
@@ -37,6 +37,7 @@ import (
 // to offer all the functions here.
 type Backend interface {
 	BlockChain() *core.BlockChain
+	FileSystem() *core.FileSystem
```
txhsl: Maybe not necessary? I mean the miner doesn't use this interface,  although the backend has it.

So as the `mockBackend` in the testing file, it seems we don't have to mock the `fs`.

We can keep the `miner` package clean with no modification.

qianhh:
The sidecar module in the "download" process might not be necessary. If a new node is started to initiate synchronization on a very long chain, problems may arise because some sidecars have already been outdated and destroyed. If the task times out, the peer will be dropped, which will affect the synchronization of the entire chain. 

Fortunately, the `sidecar_fetcher` component itself checks the blob data during the retention period. Failure to synchronize the blob data will not result in dropping the peer, and the blob data synchronization with the block is decoupled.

cmd/utils/flags_blobstorage.go
qianhh: Here, the `errInvalidBlobRetentionEpochs` is defined for the purpose of testing. Although it may seem a bit strange, let's keep it as it is for now.

eth/protocols/beacon/broadcast.go
qianhh: Renamed.

beacon/impl/fetcher/sidecar_fetcher.go
```diff
@@ -0,0 +1,325 @@
...
+	defer timeoutTicker.Stop()
+
+	fetchHeight := uint64(0)
+	height := f.chain.CurrentBlock().Number.Uint64()
+	minRetainBlockNumbers := uint64(params.MinEthEpochsForBlobsSidecarsRequest * params.BlocksPerEthEpoch)
+	if height > minRetainBlockNumbers {
+		fetchHeight = height - minRetainBlockNumbers
+	}
+
+	for {
```
txhsl, line 169: This is a downloader-like job, since the original responsibility of a fetcher is to react to some broadcast/announcement, but this one is a self-loop task, isolated from beacon handler.

To make details clear, let me introduce a bit about the block fetcher. It is responsible for receiving and handling new block messages after the downloader finishes.

The beacon handler will notify the fetcher that there's a new block in the network once receive a block message, and let the fetcher schedule the task to fetch the full block. The block fetcher loops in the background and download block independently.

However, the current sidecar fetcher is, checking the local block height by itself and send requests independently. This is a bit redundant with the beacon handler, since the handler is also requesting and downloading based on new blob announcement.

Although it's rare to happen, the beacon handler may take too long on performing the blob fetch. Can this be improved?

eth/handler.go
```diff
@@ -826,44 +843,28 @@ func (h *handler) BroadcastRequestTxs(txHashes []common.Hash) {
 	}
 }
 
-func (h *handler) handleBlobsEvents() {
+func (h *handler) handleAnnoBlobsEvents() {
```
txhsl: Prefer `blobsAnnounceLoop`, align with other namings e.g. `txBroadcastLoop`.

beacon/impl/fetcher/sidecar_fetcher.go
qianhh: Well, there is indeed a situation of duplicate downloads. I'm thinking about how to incorporate the download task for new blob announcement into it.

eth/handler.go
qianhh: Renamed.

txhsl (approved):
LGTM

PR #528, opened 2025-10-28
consensus: eth: refactor PreBlock handling and support Envelope event subscription

### LINKED ISSUE #524: Not-isolated execution of Envelope handling in PreBlock may cause a DOS
#### System information

Geth version: `v0.5.0-stable`

#### Expected behaviour

We expect every valid Envelope to be decrypted and executed if the decrypted data is valid for the blockchain state.

#### Actual behaviour

However, if a valid decrypted transaction changes the account balance (e.g. transferring all its GAS out) and makes its following decrypted transactions and Envelopes unexecutable, then the whole PreBlock will be reverted to the original proposal. If someone keeps on doing so, then there is a DOS of AMEV.

https://github.com/bane-labs/go-ethereum/blob/db10518a68898211da1e6c574a7bee5d102d8f81/consensus/dbft/dbft.go#L1146-L1170

The above `Add` can figure out the insufficient balance and return an error about insufficient balance, so that the PreBlock is reverted.

https://github.com/bane-labs/go-ethereum/blob/db10518a68898211da1e6c574a7bee5d102d8f81/consensus/dbft/dbft.go#L1242-L1273

And also, here, if one of the transactions in the list failed in state processing, the PreBlock is reverted.

IMO, these kinds of reverts are more like emergency handling instead of normal use cases. If any of them happens, dBFT is saved, but the guarantee of a transaction execution is broken. Because a transaction can be affected by its followers, which is good in a blockchain.

#### Implementation

We need to manually construct the final block by `applyTransaction`, so we can drop a transaction if it fails, then the constructed PreBlock will never revert.

### LINKED ISSUE #527: Add support for Envelope-related event subscription
# Rationale

Currently, our client doesn't provide any service to let the users know what happens to their Envelope transactions. So they feel hard to construct their Envelopes and use the Anti-MEV service.

# Implementation

Since we don't include Envelopes in any of our finalized/persisted data structure, an event for real-time subscription seems the most reasonable choice.

What we can provide is a data flow about Envelope handling instead of a continuous history.

txhsl:
Close #524. Close #527.

- [x] Add real-time event subscription for Envelope handling;
- [x] Eliminate the result influence between succeeded Envelopes and failed Envelopes.

@songb2 The new Envelope event can be subscribed through socket with the following request:

```json
{"id":1,"jsonrpc":"2.0","method":"eth_subscribe","params":["envelopes"]}
```

### DIFF
--- a/eth/filters/api.go
+++ b/eth/filters/api.go
@@ -26,6 +26,7 @@ import (
 	"time"
 
 	"github.com/ethereum/go-ethereum"
+	"github.com/ethereum/go-ethereum/antimev"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/consensus/misc/eip1559"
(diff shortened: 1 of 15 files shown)


eth/filters/api.go
```diff
@@ -308,6 +309,35 @@ func (api *FilterAPI) Logs(ctx context.Context, crit FilterCriteria) (*rpc.Subsc
 	return rpcSub, nil
 }
 
+// Envelopes creates a subscription that fires for all new envelopes that enter the consensus engine.
+func (api *FilterAPI) Envelopes(ctx context.Context) (*rpc.Subscription, error) {
```
vncoelho, line 313: good change!

qianhh (approved):

txhsl:
Rebased to `bane-main`.

PR #538, opened 2025-12-23
beacon: miner: add minimized beacon implementation and reuse post-merge mining process

### LINKED ISSUE #539: Refact miner to reuse the CL-EL pipeline
# Rationale

The EL miner only builds blocks/payloads on demand since v1.14, so it's no longer a forever-running background goroutine, and the burden of seed nodes is significantly reduced.

# Implementation

We need to merge this miner change, and implement a simple beacon client as the block building trigger.

This CL client communicates with EL through in-process API requests, and reuses the CL-EL message pipeline. Nodes without `--mine` should never prepare blocks in the background.

txhsl:
Close #539.

The miner code now catches up with the Geth v1.15.11, no longer a busy goroutine for both validator node and seed node. I move a simplified version to the new CL goroutine.

1. The support for Clique is removed, now dBFT is the only consensus can seal;
2. `Withdrawals` and `ParentBeaconRoot` are now published by CL, instead of set in dBFT;
3. Other implementations e.g. downloader, blockchain, dBFT remains the same, with TTD still not configured, so this is still a pre-merge chain;
4. Further work is still needed to fit to ETH69 and other post-merge behaviors, e.g. there is a redundancy between downloader and `engine_forkchoiceUpdated` about head setting.

The current mining flow becomes the following:

1. A new chain head is set to local, or there's a manual start input;
5. The mini CL sends `engine_forkchoiceUpdated` to EL miner through in-process RPC API;
6. The EL builds a payload locally (if validator);
7. The mini CL gets the payload ID and requests the payload data through `engine_getPayload` from EL;
8. The mini CL rebuilds the block and sends it to dBFT for sealing.

Some steps of the [CL-EL communication](https://deepwiki.com/ethereum/go-ethereum/9-consensus-layer-integration) are removed, e.g. the background running of payload building and the execution payload validation with `engine_newPayload`. That means:

1. EL may take more time to respond to `engine_forkchoiceUpdated`, and no empty block is available for a quick readiness;
2. CL won't request EL to verify a block.

Since dBFT is still allowed to access EL data and can ensure the network liveness, I think we are OK with above changes.

Besides, flag `--miner.coinbase` is changed to `--miner.pending.feeRecipient`, but its meaning remains the same, and we still overwrite the real fee recipient to GovReward in miner.

There are also several incompatibilities imported:
1. The TTD check of `engine_forkchoiceUpdated` is currently disabled, for pre-merge situation;
4. A field `Difficulty` is added to payload data, but since we use in-process communication, it can be removed easily when possible;
5. We still don't have finalized blocks, but like Point 2, we can add when possible.

@songb2 Testing is needed. @AnnaShaleva I invite you to have a review on this, and maybe you have ideas about whether we can remove the `difficulty` in our block (by setting it to 0).


txhsl:
Currently the Privnet Node 0 is not working correctly, will fix.

> Validators inside the consensus list for the next block build their payloads locally, others only set head;

Seems due to this, reverted this change.

beacon/impl/worker.go
```diff
@@ -0,0 +1,495 @@
...
+		payload, err := w.getPayload(resp.PayloadID, timestamp)
+		if err != nil {
+			log.Error("Failed to fetch payload", "err", err)
+			return
+		}
+
+		// Rebuild the block from execution payload.
+		versionedHashes := make([]common.Hash, 0)
+		for i, commitment := range payload.BlobsBundle.Commitments {
+			versionedHashes[i] = convertKzgCommitmentToVersionedHash(commitment)
```
qianhh: The versionedHashes array should be allocated with a certain length.

qianhh:
The following test cases failed. Please refer to the Geth v1.15.11 version for the repair.
`TestAccountUpdate`
`TestConsoleWelcome`
`TestAttachWelcome`
`TestWelcome`

Issue #541, opened 2025-12-25
Apply v1.13.15 post-merge updates as much as possible after implementing the mock CL


txhsl:
# Rationale

The current downloader is still pre-merge. Since we haven't set the TTD, the node synchronization still depends on the head broadcast, which starts from the genesis.

After TTD, Geth uses a reversed synchronization sequence, which starts from the latest.

# Implementation

The new pipeline of block proposing and downloading will be enabled by #539 and #540.

So in this issue, we focus on the downloader update. Only after this we can have an independent beacon network.

txhsl:
After some investigation, I decide to change this proposal to another topic.

The post-merge downloader of EL provides a quick state availability and sync efficiency,  by applying a reversed order of synchronization. However, this relies on the trust of its CL client, which is still built on top of a chained trust.

Since we don't have a real beacon chain, an order-reversed synchronization is no reliable enough to be performed.

PR #542, opened 2026-01-08
beacon: consensus/dbft: eth: split eth protocol, and move consensus and block broadcast fully on top of beacon

### LINKED ISSUE #540: Add a beacon protocol to support removed header handling in ETH68/ETH69
# Rationale

Geth v1.16 removes the header data exchange among EL. This means that the downloader can only know about the lastest block/head information from CL `engine_newPayload`/`engine_forkchoiceUpdated`.

`engine_newPayload` tells the EL what blocks (named execution payload) appear in the CL network, and `engine_forkchoiceUpdated` tells the EL which block is selected as the head block.

# Implementation

The `engine_forkchoiceUpdated` communication will be enabled in #539, so in this issue, we need something to replace/reuse the CL functionality about `engine_newPayload`.

txhsl:
Close #540, and is also an enhancement of #539.

By moving the block broadcasting and fetching to beacon layer, now the dBFT doesn't write to the EL blockchain directly.

TODO:
1. The downloader is still the old, so that it still needs both the `beacon` and `eth` protocols to work. We need to decouple it, since there is a post-merge way of synchronization. Let's include it in #541;
2. The blob-related message is not really functioning, needs some code from #526;
3. The dBFT consensus should execute a DA check on all blob sidecars in verification.

### DIFF
--- a/beacon/engine/gen_ed.go
+++ b/beacon/engine/gen_ed.go
@@ -28,6 +28,7 @@ func (e ExecutableData) MarshalJSON() ([]byte, error) {
 		GasLimit         hexutil.Uint64          `json:"gasLimit"      gencodec:"required"`
 		GasUsed          hexutil.Uint64          `json:"gasUsed"       gencodec:"required"`
 		Timestamp        hexutil.Uint64          `json:"timestamp"     gencodec:"required"`
+		Nonce            hexutil.Uint64          `json:"nonce"         gencodec:"required"`
 		ExtraData        hexutil.Bytes           `json:"extraData"     gencodec:"required"`
 		BaseFeePerGas    *hexutil.Big            `json:"baseFeePerGas" gencodec:"required"`
 		BlockHash        common.Hash             `json:"blockHash"     gencodec:"required"`
@@ -49,6 +50,7 @@ func (e ExecutableData) MarshalJSON() ([]byte, error) {
 	enc.GasLimit = hexutil.Uint64(e.GasLimit)
 	enc.GasUsed = hexutil.Uint64(e.GasUsed)
 	enc.Timestamp = hexutil.Uint64(e.Timestamp)
+	enc.Nonce = hexutil.Uint64(e.Nonce)
 	enc.ExtraData = e.ExtraData
 	enc.BaseFeePerGas = (*hexutil.Big)(e.BaseFeePerGas)
 	enc.BlockHash = e.BlockHash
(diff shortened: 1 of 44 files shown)


txhsl:
There‘s a possible `header not found` error caused by multi-sig scheme reorgs, ~~trying to fix~~ should have been solved, need further testing.

qianhh:
I submitted a blobtx transaction, but it couldn't be included in the block and the following error occurred:

`
WARN [01-15|18:16:08.459] Invalid NewPayload params                params.Number=16 params.ParentHash=f07d45..ec9259 params.BlockHash=c3db11..cd8926 params.StateRoot=909c83..940f1a params.FeeRecipient=0x1212000000000000000000000000000000000003 params.LogsBloom="0x000000...000000 (256B)" params.Random=a0af6a..b4ddb2 params.GasLimit=30,472,174 params.GasUsed=21000 params.Timestamp=1,768,472,163 params.Nonce=0 params.ExtraData="0x0201e4...2e896c (178B)" params.BaseFeePerGas=20,000,000,000 params.BlobGasUsed=131072 params.ExcessBlobGas=0 len(params.Transactions)=1 len(params.Withdrawals)=0 beaconRoot=56e81f..63b421 len(requests)=0 error="invalid number of versionedHashes: [] blobHashes: [0x01a38e7f9b929e41c547fa000c70617ce078234a73cdfaeb6c888e128e54eb27]"
WARN [01-15|18:16:08.459] error on enqueue block                   error="block rejected by EL, err: invalid number of versionedHashes: [] blobHashes: [0x01a38e7f9b929e41c547fa000c70617ce078234a73cdfaeb6c888e128e54eb27]"
`

txhsl:
> I submitted a blobtx transaction, but it couldn't be included in the block and the following error occurred:

The block is built with `tx.WithoutBlobTxSidecar()`, so there is no versioned hash can be collected in `feedback`. Let me check if there is an easy solution, for temporary usage.

txhsl:
Simple rebase, no critical change or fix.

consensus/ethash/ethash.go
```diff
@@ -84,7 +81,7 @@ func (ethash *Ethash) APIs(chain consensus.ChainHeaderReader) []rpc.API {
 
 // SubscribeEnvelopeEvent creates a subscription that fires for all new envelopes that enter the consensus engine.
 func (ethash *Ethash) SubscribeEnvelopeEvent(ch chan<- []*antimev.EnvelopeInfo) event.Subscription {
-	return ethash.scope.Track(ethash.envelopeFeed.Subscribe(ch))
```
qianhh, line 87: After making these modifications, some test cases failed.

txhsl: Seems revert is the only choice, reverted.

And this make me find a goroutine leak, will fix from #538.

PR #595, opened 2026-04-24
beacon/impl: eth: add proposal cache for missing tx searching

txhsl:
Close #576. Close #604.

This PR moves the missing tx protocol to the CL, and adds the block proposal into the searching boundary, to solve the tx replacement issue on the Testnet.

I adopt the minor change here as the first step. If this works on the Testnet, then we may push the refactor forward.

PR #593, opened 2026-04-14
beacon/impl: eth: refactor synchronization to post-merge way

txhsl:
Close #588. Close #594.

This PR implement #453 in our CL, so that we can do the synchronization in a post-merge way.

We'll have a CL synchronizer to manage and verify block headers, without processing all of the blocks. After finding the latest block, it will send trustful signals to the EL, without reorg which is not supported by the EL downloader.

- [x] Merge the post-merge downloader;
- [x] Implement the LC synchronizer with unit tests;
- [x] Merge updates after #582 is merged;
- [x] Implement a storage for the LC progress;
- [x] Test and make all things work as expected.

Known issues:

- [x] The v1.15.11 codebase seems not working well with its PathDB, so currently please use HashDB instead;
- [x] The synchronization suffers from #594;
- [x] There seems an issue when seed node starts along with other CNs from the genesis block, the sync doesn't exit correctly.

PR #598, opened 2026-04-24
beacon/impl: eth: improve the initial fetch height in sidecar fetcher

### LINKED ISSUE #597: Improve the initial fetch height in sidecar fetcher
# Rationale

Currently, the initial synchronization position calculation of the sidecar fetcher is not precise enough. If a complete re-synchronization is required, it will result in a lot of unnecessary requests. Therefore, the process needs to be improved by only synchronizing the sidecars that need to be saved.

# Implementation

By obtaining the latest block headers of other peer, the starting block number for synchronization can be calculated.

qianhh:
Close #597 .

By upgrading the status message of the beacon protocol, the latest block header of the peer can be quickly obtained, which is used to initialize the fetch height in the sidecar fetcher.

### DIFF
--- a/eth/protocols/beacon/handler.go
+++ b/eth/protocols/beacon/handler.go
@@ -94,6 +94,16 @@ var beacon1 = map[uint64]msgHandler{
 	BatchBlobsMsg:     handleBatchBlobs,
 }
 
+var beacon2 = map[uint64]msgHandler{
+	NewBlockHashesMsg: handleNewBlockhashes,
+	NewBlockMsg:       handleNewBlock,
+	NewBlobsRootMsg:   handleNewBlobsRoot,
+	GetBlobsMsg:       handleGetBlobs,
+	BlobsMsg:          handleBlobs,
+	GetBatchBlobsMsg:  handleGetBatchBlobs,
+	BatchBlobsMsg:     handleBatchBlobs,
+}
+
 // handleMessage is invoked whenever an inbound message is received from a
 // remote peer on the `beacon` protocol. The remote connection is torn down upon
 // returning any error.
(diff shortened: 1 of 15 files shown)


eth/protocols/beacon/handler.go
```diff
@@ -94,6 +94,16 @@ var beacon1 = map[uint64]msgHandler{
 	BatchBlobsMsg:     handleBatchBlobs,
 }
 
+var beacon2 = map[uint64]msgHandler{
```
txhsl, line 97: Seems the same as `beacon1`, maybe we don't need this.

eth/sync.go
```diff
@@ -174,7 +174,7 @@ func (cs *chainSyncer) nextSyncOp() *chainSyncOp {
 }
 
 func peerToSyncOp(mode downloader.SyncMode, b *beacon.Peer, p *eth.Peer) *chainSyncOp {
-	peerHead, peerTD := b.Head()
+	peerHead, peerTD, _ := b.Head()
```
txhsl: Need to rebase to #593.

eth/protocols/beacon/protocol.go
```diff
@@ -69,6 +70,18 @@ type StatusPacket struct {
 	BlobSync        bool
 }
 
+// StatusPacket2 is the network packet for the status message.
+type StatusPacket2 struct {
+	ProtocolVersion uint32
+	NetworkID       uint64
+	TD              *big.Int
+	Head            common.Hash
+	HeadNumber      uint64
```
txhsl, line 79: I'm not sure, but can't we add this new field without declaring a new packet and the related handler?

Since the `highestNumberOfBeacons` is injected to the sidecar fetcher, the node who runs the old protocol are set to 0 `headNumber` as well. The handshake itself is not so different to `beacon1`?

eth/protocols/beacon/handler.go
qianhh: We will need to clean up the code related to beacon1 in the next version.

eth/protocols/beacon/protocol.go
qianhh: This is done to make the upgrade smoother. When upgrading at some nodes, it should be able to be compatible with the previous version.

eth/sync.go
qianhh: Ok. I'm going to rebase.

txhsl: Not now, you can rebase this branch after that one is merged.

eth/protocols/beacon/handler.go
txhsl: OK, then I'll reuse them in #595.

txhsl (approved):

qianhh:
Rebased to `bane-main`.

PR #603, opened 2026-04-30
eth/handler.go: fix sidecar fetcher

### LINKED ISSUE #602: Fix sidecar fetcher
#### System information

Geth version: `0.6.0-unstable`
Commit hash : 0054e697ff02ddac9221574885be9be98c1c95fc

#### Expected behaviour
Correctly enable the event loop of the sidecar fetcher.

#### Actual behaviour
The event loop was not correctly enabled for the sidecar fetcher, resulting in network message blocking.

Run the network using `make privnet_start_seven`, and sending blob transaction will reproduce the result.

qianhh:
Close #602.

Correctly enable the event loop of the sidecar fetcher.

### DIFF
--- a/eth/handler.go
+++ b/eth/handler.go
@@ -578,6 +578,11 @@ func (h *handler) Start(maxPeers int) {
 	// start sync handlers
 	h.txFetcher.Start()
 
+	// start sidecar fetcher if blob sync is enabled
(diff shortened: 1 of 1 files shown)


txhsl (approved):
LGTM

PR #610, opened 2026-05-20
eth/downloader: refactor BeaconLightSync method

qianhh:
Close #609 .

Imitate the `skeleton` to implement a multi-task mode synchronizer.

PR #612, opened 2026-05-26
eth/downloader: prevent the cancelled request to trigger a revert process

qianhh:
Related to #610, and maybe the reason of https://github.com/bane-labs/go-ethereum/pull/610#issuecomment-4541539721.

- Fix the cancelled request will also trigger a revert process, which can lead to a panic situation
- Correct the judgment logic of the first task


txhsl (approved):
LGTM, @songb2 you can still test after this is merged

PR #606, opened 2026-05-08
eth: bind the syncing check to downloader

txhsl:
Close #605. Close #607.

PR #614, opened 2026-05-27
beacon, eth: enhance the overall synchronization status judgment

### LINKED ISSUE #611: CN may drop other nodes when resuming synchronization
#### System information

Geth version: `v0.6.0-stable`
Commit hash : 9d43dce6be9b43fc28f5f724a3f7f28ac82e6272

#### Expected behaviour

Related to #588. We expect the new downloader works well as the pre-merge one.

The new synchronization process has three steps instead of only one.

1. Beacon sync - CL performs a light header verification to find the latest head block produced by Neo X dBFT;
2. Header sync - EL downloads all the headers of the chain;
3. Block sync - EL downloads all necessary full blocks of the chain.

As a result, the syncing status depends on both of the CL and EL. We've fixed the status check of Step 2 and Step 3 in #606 and #608, so the EL process is continuous.

#### Actual behaviour

However, between Step 1 and Step 2, when the CL synchronizer stops but the EL downloader hasn't yet received the message, there is a short period of time that the following lines returns `false`.

https://github.com/bane-labs/go-ethereum/blob/9d43dce6be9b43fc28f5f724a3f7f28ac82e6272/eth/backend.go#L456-L458

If the CN receives dBFT messages in this period, it will still cause an unexpected peer drop, due to the same reason of #607. This can be solved by restart, but the status check needs a more continuous switch.

If this issue can't be solved perfectly in v0.6.0, we delay it to the next version.

qianhh:
Close #611 .

I added a timestamp for the completion of synchronization in both CL and EL. By comparing these two timestamps, we can determine the overall synchronization status where CL has completed synchronization but EL has not yet started.

### DIFF
--- a/eth/backend.go
+++ b/eth/backend.go
@@ -454,7 +454,8 @@ func New(stack *node.Node, config *ethconfig.Config) (*Ethereum, error) {
 	eth.handler.connectBeacon(eth.beacon)
 	if bft != nil {
 		syncing := func() bool {
-			return eth.beacon.Syncing() || !eth.Synced()
+			log.Debug("Syncing status", "beacon", eth.beacon.Syncing(), "eth", eth.Syncing())
+			return eth.beacon.Syncing() || eth.Syncing()
 		}
 		// Connect BFT to beacon protocol
 		bft.WithBeacon(eth.beacon.BlockBroadcaster(), eth.beacon.RefreshPendingPayload, eth.beacon.SubscribeSyncingEvents, syncing)
@@ -863,4 +864,11 @@ func (s *Ethereum) SyncMode() ethconfig.SyncMode {
 	}
 	// Nope, we're really full syncing
 	return ethconfig.FullSync
+}
+
+// Syncing returns true if the node downloader is still syncing.
+// NOTE: the method Synced() now returns if the node has the latest head,
+// so CL needs another method to check downloader state directly.
+func (s *Ethereum) Syncing() bool {
+	return s.handler.downloader.Syncing()
 }
--- a/eth/backend.go
+++ b/eth/backend.go
@@ -454,7 +454,7 @@ func New(stack *node.Node, config *ethconfig.Config) (*Ethereum, error) {
 	eth.handler.connectBeacon(eth.beacon)
 	if bft != nil {
 		syncing := func() bool {
-			return eth.beacon.Syncing() || eth.Syncing()
+			return eth.beacon.Syncing() || !eth.Synced()
 		}
 		// Connect BFT to beacon protocol
 		bft.WithBeacon(eth.beacon.BlockBroadcaster(), eth.beacon.RefreshPendingPayload, eth.beacon.SubscribeSyncingEvents, syncing)
@@ -863,11 +863,4 @@ func (s *Ethereum) SyncMode() ethconfig.SyncMode {
 	}
 	// Nope, we're really full syncing
 	return ethconfig.FullSync
-}
-
-// Syncing returns true if the node downloader is still syncing.
-// NOTE: the method Synced() now returns if the node has the latest head,
-// so CL needs another method to check downloader state directly.
-func (s *Ethereum) Syncing() bool {
-	return s.handler.downloader.Syncing()
 }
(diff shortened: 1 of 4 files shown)


beacon/impl/synchronizer/synchronizer.go
```diff
@@ -226,6 +229,9 @@ func (s *Synchronizer) BeaconExtend(verifiedHeaderHashes []common.Hash, finalize
 // Stop marks the synchronizer as stopped, and the next notification will trigger
 // the sync process again.
 func (s *Synchronizer) Stop() {
+	if s.startSyncTrustedHeadHash != s.trustedHead.Hash() {
+		s.latestSyncedMilliTimestamp.Store(time.Now().UnixMilli())
```
txhsl: What if the check returns `false` and skip the assignment?

I mean `eth.beacon.Syncing() || eth.Syncing() || eth.beacon.LatestSyncedMilliTimestamp() > eth.latestSyncedMilliTimestamp()` can be `false`, because `0 == 0`?

qianhh: If the Synchronizer triggers an update when it starts, but in fact it is already the latest head, there is no need to update the timestamp. This is because in such cases, the eth might not trigger an synchronization either. It would be incorrect without making this judgment.

txhsl: > If the Synchronizer triggers an update when it starts, but in fact it is already the latest head, there is no need to update the timestamp.

Yes, it's not necessary.

> This is because in such cases, the eth might not trigger an synchronization either.

But this maybe incorrect. The CL synchronizer is a tracer for the latest chain head, every time the node starts, it will find the latest trustful head block and send it to EL.

It means, if I shutdown the node when EL is syncing, the CL is synced. When I restart the node, the CL can be still synced (if there's no new block), but it will send the latest block to EL to resume its synchronization.

Since the `latestSyncedMilliTimestamp` is not in the DB, the CL and EL will both have `0`

qianhh: Yes, in this case, it becomes invalid.

eth/backend.go
```diff
@@ -454,7 +454,9 @@ func New(stack *node.Node, config *ethconfig.Config) (*Ethereum, error) {
 	eth.handler.connectBeacon(eth.beacon)
 	if bft != nil {
 		syncing := func() bool {
-			return eth.beacon.Syncing() || eth.Syncing()
+			middle := !eth.Synced() && eth.beacon.LatestSyncedMilliTimestamp() > eth.latestSyncMilliTimestamp()
```
txhsl: `middle` is a bit inaccurate, how about `downloading`?

eth/backend.go
```diff
@@ -454,7 +454,9 @@ func New(stack *node.Node, config *ethconfig.Config) (*Ethereum, error) {
...
+			log.Debug("Syncing status", "beacon", eth.beacon.Syncing(), "eth", eth.Syncing(), "middle", middle)
```
txhsl: The combination of `eth.Syncing()` and `middle` here seems redundant.

Maybe it's better to revert #606 and only use `eth.beacon.Syncing() || middle`/`eth.beacon.Syncing() || downloading`.

qianhh: Yes, `middle` already includes the synchronous state of `skeleton`.

qianhh:
I reorganized the logic and made some changes, which I think are more reasonable.

eth/downloader/beaconsync.go
```diff
@@ -186,6 +186,7 @@ func (d *Downloader) beaconSync(mode SyncMode, head *types.Header, final *types.
 	if err := d.skeleton.Sync(head, final, force); err != nil {
 		return err
 	}
+	d.syncing.Store(true)
```
txhsl, line 189: The idea is good, but this implementation seems not able to solve the original issue.

There's time gap between CL finishes syncing and the head signal reaches this line. The head block needs to be converted into payload, sent to the engine API and converted back to a block. 

But the previous time-based one can cover and prevent the mishandling. To be honest, that works more like we expect.

qianhh: I traced the entire call chain. The `Synchronizer.complete` callback function execution reaches here. At this point, the state in `Synchronizer` has not been set yet, so the entire state transfer is seamless and there is no gap.

txhsl: 🤔Well, looks like. OK, so the current workflow in CL is the light sync checks and ensures the head signal has been sent and returned with valid results, then stop the beacon synchronizer.

txhsl (approved):
LGTM, but I need to run a test before merging this.

PR #615, opened 2026-05-27
eth: notify block fetcher when handshake if necessary

### LINKED ISSUE #613: Make the CL synchronizer start even there's no new block in the network
# Rationale

The current CL synchronization requires a new block broadcast as the trigger. Which means, if there's no block, e.g. the consensus breaks down, then the new nodes in the network can't start synchronization, even though the chain data is in fact available.

This is kind of risky, although still protected by the 1/3 rule. If too many CNs delete their local chain data, they can't perform sync to restart the consensus.

# Implementation

The lastest head infomation carried by `beacon` handshake message, should be used to trigger CL synchronization as well.

txhsl:
Close #613. Based on #598.

Also contains a fix, ref https://github.com/bane-labs/go-ethereum/pull/615#issuecomment-4598653743.

### DIFF
--- a/beacon/impl/synchronizer/synchronizer.go
+++ b/beacon/impl/synchronizer/synchronizer.go
@@ -33,7 +33,9 @@ type LightVerifyFn func(headers []*types.Header) bool
 type LightSyncFn func(extend BeaconExtendFn, complete func(), start chan *types.Header) error
 
 // BeaconExtendFn is a callback type for extending the trusted header chain with headers received from the network.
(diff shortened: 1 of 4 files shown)


txhsl:
Works well as I expect.

txhsl:
#610 imported an issue that the beacon synchronizer won't stop after the light sync finishes, will fix after rebase.

The reason is it requires chain connection to stop, but since there's no new block broadcast in the network. there's no chain to merge with.

PR #629, opened 2026-06-11
beacon/impl: eth: fix and improve the CL synchronizer service and dBFT syncing check

### LINKED ISSUE #627: Fake chain attack
#### System information

Geth version: `v0.6.0-stable`

#### Expected behaviour

We expect our CNs handle new blocks correctly, if there's a new block, then respond to it.

#### Actual behaviour

The issue is, if someone broadcasts an independent new chain, which is not signed by the exact consensus, and heigher than the real one, our CNs may stop the dBFT procedure, fail the synchronization and keep waiting.

1. Build a private long chain and an independent short chain;
2. Merge the network, and only enable the consensus for the short chain.

txhsl:
Close #627.

- [x] Improve the syncing status check for smooth dBFT message handling;
- [x] Fix the status inconsistency that makes the node permanent waiting;
- [x] Implement a pending chain pruning mechanism for the CL synchronizer.

### DIFF
--- a/beacon/impl/beacon.go
+++ b/beacon/impl/beacon.go
@@ -142,6 +142,11 @@ func (b *Beacon) Syncing() bool {
 	return b.synchronizer.Syncing()
 }
 
+// InitialSynced returns whether the beacon client has ever been synced, which means a round of light sync has been completed.
+func (b *Beacon) InitialSynced() bool {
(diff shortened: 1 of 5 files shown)


txhsl:
@songb2 Ready for test

qianhh (approved):

PR #634, opened 2026-06-22
Release v0.6.1

txhsl:


AnnaShaleva:
@txhsl Please, wait before the merge and before publish the release. I'll push a PR with an automated GH Actions release job soon (today), we may try it out on v0.6.1 release.

txhsl:
> @txhsl Please, wait before the merge and before publish the release. I'll push a PR with an automated GH Actions release job soon (today), we may try it out on v0.6.1 release.

We aim for Wednesday, still waiting #629.

gsmachado:
@AnnaShaleva awesome!
don't forget to include arm64 binaries besides the usual amd64, please! 😄 
and let's try to use self-hosted runners 😄

AnnaShaleva:
> don't forget to include arm64 binaries besides the usual amd64

Will do.

> and let's try to use self-hosted runners

No. I'll setup GH Actions runners (that's what we use in nspcc-dev for production and what I have a strong experience in). Existing workflow doesn't work properly with self-hosted runner, you need to create an issue to properly migrate from GH Actions to self-hosted runners if you'd like to.

gsmachado:
> No. I'll setup GH Actions runners (that's what we use in nspcc-dev for production and what I have a strong experience in). Existing workflow doesn't work properly with self-hosted runner, you need to create an issue to properly migrate from GH Actions to self-hosted runners if you'd like to.

@AnnaShaleva Then, we risk not having arm64 binary, since GH Actions runners were failing for arm64 (as far as I remember).

What's the reason why "existing workflow" doesn't work properly with self-hosted runners? Isn't it just to add a label on the jobs and that's it?! It seems easy.

AnnaShaleva:
> since GH Actions runners were failing for arm64

Check out the https://github.com/bane-labs/go-ethereum/pull/635 please, it works well.

> What's the reason why "existing workflow" doesn't work properly

Default `build/ci.go` script has `--arch` flag, but arm64 binary cross-build doesn't work out-of-the-box on amd64 even with proper `GOARCH` and `GOOS` set, so I see two ways: we either need to dig a little bit deeper to use `ci.go` on amd64 for arm64 builds or use an arm64 runner:
https://github.com/bane-labs/go-ethereum/blob/cddf43c8c7001bd7469eb48d4ef995dfe76399df/build/ci.go#L227
Tested on amd64 GH runner, ref. https://github.com/AnnaShaleva/go-ethereum/actions/runs/27951605590/job/82709595526:
```
Run make all
go run build/ci.go install --arch arm64
go: downloading github.com/cespare/cp v0.1.0
go: downloading github.com/holiman/uint256 v1.3.2
go: downloading golang.org/x/crypto v0.39.0
go: downloading github.com/Azure/azure-sdk-for-go/sdk/storage/azblob v1.2.0
go: downloading golang.org/x/sys v0.36.0
go: downloading github.com/Azure/azure-sdk-for-go/sdk/azcore v1.7.0
go: downloading github.com/Azure/azure-sdk-for-go/sdk/internal v1.3.0
go: downloading golang.org/x/net v0.41.0
go: downloading golang.org/x/text v0.26.0
fork/exec /tmp/go-build583669073/b001/exe/ci: exec format error
```
Seting up our own arm64 runner looks excessive to me, as well as maintaining an amd64 runner whereas there are free GH Action runners available for us. So the most convenient way to me is just to migrate onto GH runners, both amd64 and arm64 work out-of-the-box.

gsmachado:
> Seting up our own arm64 runner looks excessive to me, as well as maintaining an amd64 runner whereas there are free GH Action runners available for us. So the most convenient way to me is just to migrate onto GH runners, both amd64 and arm64 work out-of-the-box.

@AnnaShaleva we already have our own arm64 runner. AxLabs set up for BaneLabs. Available to be used, since a long time. 👇 

image

I mentioned setting up using our own GH runners because we already went through similar problems like this in (several) other golang repos... and, in my opinion, having our own GH runners is faster and more predictable than relying on GH public stuff.

In addition to that, we set up our own runners (amd64 and arm64) because BaneLab has private repos, and GH runners have a limited free tier for the whole org.

So... I'm fine if you decide that we should rely on public ones for `go-ethereum` repo, but the resources (self-hosted runners) are available and will remain there, anyway. 🤷‍♂️

txhsl:
> In addition to that, we set up our own runners (amd64 and arm64) because BaneLab has private repos, and GH runners have a limited free tier for the whole org.

Yes, we quietly often hit the usage limit of free GH Actions runners, so we moved to self-hosted ones. So, it's better if we can keep this way.

> Default `build/ci.go` script has `--arch` flag, but arm64 binary cross-build doesn't work out-of-the-box on amd64 even with proper `GOARCH` and `GOOS` set, so I see two ways: we either need to dig a little bit deeper to use `ci.go` on amd64 for arm64 builds or use an arm64 runner:

Well, my experience tells me that the arm64 binary can't be built on amd64 machines, since I set `CGO_ENABLED=0` for every time.

But it's not a reason to block #635, we can do the migration later.

qianhh (approved):

PR #641, opened 2026-06-30
cmd: eth: internal: allow set head with timestamp and add an optional rewind

### LINKED ISSUE #638: Add an optional rewind step for `geth init`
# Rationale

Replace #515. Ref https://github.com/bane-labs/go-ethereum/issues/515#issuecomment-4798226977, it's possible that Ethereum imports new CL block fields, but keeps the EL one the same, like in Osaka. In this case, the EL code can't detect the fork change through the block signature. We don‘t have independent CL chains, so the CL can't detect either.

Geth has an automatic rewind handler for incompatible fork configurations. If the new config inside the code is different from which inside the DB, then rewind the head and wait for a new synchronization. However, this rewind is not enabled for the `geth init` command.

Since we take this different way for configuration, due to the content size of the genesis block. Our nodes get a dead lock if the database reinitialization is later than the fork upgrade.

# Implementation

We need to implement a rewind handler for the `geth init` command when a `ConfigCompatError` is detected, and give our user a choice to exit or rewind the local chain head.

txhsl:
Close #638.

After this change:
1. The rewind happens in `geth init` should write the correct chain config into the database, instead of just reporting an error;
2. The chain head should be reverted to the last block before the incompatible fork is enabled, and the blocks among this gap should be dropped;
3. The node then should be able to and start to sync from the new head to the latest head;
4. The node should enable dBFT after the local chain gets synced up to date;

### DIFF
--- a/cmd/geth/chaincmd.go
+++ b/cmd/geth/chaincmd.go
@@ -33,6 +33,7 @@ import (
 	"github.com/ethereum/go-ethereum/cmd/utils"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/ethereum/go-ethereum/console/prompt"
 	"github.com/ethereum/go-ethereum/core"
 	"github.com/ethereum/go-ethereum/core/history"
 	"github.com/ethereum/go-ethereum/core/rawdb"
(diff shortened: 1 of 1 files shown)


txhsl:
The rewinded chain doesn't start syncing, still trying to find the reason.

OK, seems like an incorrect step in my testing.

qianhh (approved):

PR #644, opened 2026-07-08
contracts: prevent sending rewards to validators with code delegation

### LINKED ISSUE #643: Update Governance reward distribution check according to EIP-7702
# Rationale

EIP-7702 brings a possibility that an EOA account can dynamically set and cancel the contract delegation.

It means that our validators and voters can use contract trigger in distribution, after registration or voting.

# Implementation

Make the account `code` check real-time before any transfer happens.

txhsl:
Close #643. With Hardhat and related dependencies upgraded.

The method `receive()` of Governance contract is called in dBFT handling by every block and any other direct transfer. So we don't expect to execute any third-party code which may cause the execution to fail.

@mialbu I'd be appreciate if you can review this and check if there's any other issues about GAS transfer.


txhsl:
Testnet deployed to `0x72ED2699E11821C910E29a7965D11ab2779cC395`
Mainnet deployed to `0xF2714AB189e95f6cB60c2624EC8b8a876abAB2Fb`

mialbu:
> @mialbu I'd be appreciate if you can review this and check if there's any other issues about GAS transfer.

It looks good to me. Afaik, it's only `receive()` that's being called during consensus (via `Governance.onPersist()` -> `GovReward.withdraw()` -> `Governance.receive()`). All other ETH transfer paths are fine to revert (and expected to) if the recipient has code imo.

One caveat I noticed – even though you’re clearly already aware of it, given the fix in `receive()` – is that even though we only accept EOAs when registering as a candidate, once registered, they can still delegate and use code underneath which ultimately would make them forfeit reward payments.

txhsl:
> One caveat I noticed – even though you’re clearly already aware of it, given the fix in `receive()` – is that even though we only accept EOAs when registering as a candidate, once registered, they can still delegate and use code underneath which ultimately would make them forfeit reward payments.

Yes, the fix seems rude. It refuses the candidate registration if the sender has a code delegation. And if a candidate delegates a code after register, he'll lose the validator reward.

To us, validator accounts are different from daily users - they are bind to the DKG transactions as well.

So far I don't see the benefit to support EIP-7702 in Governance. This is only a security patch to eliminate the potential risks.

If there's a necessity to implement, e.g. EIP-7702 is widely used/enforced by almost account and safe enough, then we may support it with a claimable map record.

mialbu:
> So far I don't see the benefit to support EIP-7702 in Governance. This is only a security patch to eliminate the potential risks.
>
> If there's a necessity to implement, e.g. EIP-7702 is widely used/enforced by almost account and safe enough, then we may support it with a claimable map record.

Yes, I agree. I think how it's handled in this implementation is fine, and as you said, we can solve it via claim decoupling that execution part from consensus once there should be actual demand for this.

PR #646, opened 2026-07-15
contracts: add GovPaymaster

### LINKED ISSUE #630: Gasless transactions
# Rationale

This is a feature proposed by AxLabs and we've already discussed before. The main goal is to redistribute X% of our governance reward to sponsor some transactions in the network, so our users can have a chance to send transactions without having any GAS in his account.

Since we have merged Geth v1.16 into Neo X with most of the post-merge features implemented, we are again in the same direction as Geth development. IMO it's time that we can add some new features on the current codebase.

The next v1.17 upgrade is huge as well, but won't cause any conflict to the architecture.

# Implementation

The basic principle of design is still changing as less as possible. Breaking the balance check of the stateful transaction validation may require different rules of txpool management, EVM execution, RPC estimation, and the miner behavior.

The only thing we can determine now is, our governance reward can be easily redistribute to a system address/contract.

My idea is to set an upper limit of the GAS sponsor (e.g. 0.1 GAS or 0.01 GAS, while a simplest GAS transfer requires only 0.00084 GAS in the network). The GovReward contract will check if the address/contract balance is lower than the limit, then send X% of the block fee every block until it again touches the limit.

@gsmachado How do you think about this rule of sponsorship? @mialbu And can we reuse the current x402 protocol for the payment? Or maybe that's not flexible enough for paying every type of transactions.

This topic is still open for discussion, so @roman-khimov @AnnaShaleva @vncoelho @qianhh your ideas are also welcome.

txhsl:
Close #630.

- [x] Implement the official Paymaster contract and reward distribution;
- [x] Add related UTs and docs;
- [x] Update the Privnet genesis files; 
- [x] Improve the validation rules.


txhsl:
@songb2

qianhh (approved):

PR #648, opened 2026-07-27
core/vm: apply the blacklist policy to EVM execution

### LINKED ISSUE #645: Blacklist policy check becomes insufficient after EIP-7702
#### System information

Geth version: `v0.6.1-stable`

#### Expected behaviour

We expect the blacklist defined in Neo X Policy can prevent specific addresses from sending transactions.

#### Actual behaviour

After EIP-7702, it's possible that a blacklisted account receives an UserOperation through the EntryPoint contract. Then the blacklisted address can still execute its code with its private key. E.g. with EIP-4337, the blacklisted address is still able to send contract calling as `msg.sender`, and transfer out GAS from its balance.

In EIP-4337, the user signature check doesn't happen in block execution. Instead, the EntryPoint contract calls the `validateUserOp`, and let the AA contract to check the signature on its own. So the Policy check out of EVM execution does't affect.

And the EIP-7702 delegation allows other senders to send the code authorization, which is a part of its sponsorship model. So the current blacklist can't prevent anything in this flow.

Ref https://github.com/eth-infinitism/account-abstraction/blob/v0.9.0/contracts/core/SenderCreator.sol. This needs to be further verified.

txhsl:
Close #645. The Governance blacklist has never been used before, so no hardfork is imported.

The old blacklist:
1. (EOA) Prevent address from sending transactions;
2. (EOA) Prevent address from becoming Governance validator.

The new blacklist:
1. (EOA) Prevent address from sending transactions;
2. (EOA) Prevent address from becoming Governance validator;
3. (EOA/contract) Prevent address from receiving GAS (except self destruct beneficiary, but it can't trigger any execution);
4. (EOA/contract) Prevent address from executing code (CALL);
5. (contract) Prevent address's code from being used (CALLCODE, DELEGATECALL and STATICCALL).

I think this now works more like as in N3.


txhsl:
@AnnaShaleva @roman-khimov I think I need NSPCC's opinion on #645 and this change, about whether we need/should extend the Policy power and change the EVM behavior.

qianhh (approved):

txhsl:
🤔There seems a possible case that if a validator gets blacklisted while being one of the CNs, then the system call will fail.

txhsl:
> 🤔There seems a possible case that if a validator gets blacklisted while being one of the CNs, then the system call will fail.

Fixed in contract.

AnnaShaleva (approved):
I think this fix is valid. Yes, it's a significant change of the EVM behaviour, but the concept of blacklisted addresses also changes EVM, and we have to support it properly with new EIPs.

> I think this now works more like as in N3.

Here are the constraints for blocked account on N3:
1. The blocked account (contract) can't be invoked directly (i.e. can't be the target of any instructions like CALL/CALLA/CALLT/dynamic calls):
https://github.com/nspcc-dev/neo-go/blob/433ffc7b4f608efd16f73a69ab8a6fee7259679c/pkg/core/interop/contract/call.go#L130
2. Blocked account can't be the signer of transaction:
https://github.com/nspcc-dev/neo-go/blob/433ffc7b4f608efd16f73a69ab8a6fee7259679c/pkg/core/blockchain.go#L2951
3. Blocked account can't be elected as a validator, can't register as a candidate and will be excluded from the list of candidates if already registered (track the usages of https://github.com/nspcc-dev/neo-go/blob/433ffc7b4f608efd16f73a69ab8a6fee7259679c/pkg/core/native/policy.go#L534).

However, it's still possible to send tokens (including GAS/NEO) to blocked account only if this account is not a contract. If it's a contract, then NEP11/NEP17 tokens can't be sent because `onNEP11Payment`/`onNEP17Payment` MUST be called on this contract to get an acknowledgement, and calls of blocked contract are prohibited. Here's GAS/NEO implementation reference:
https://github.com/nspcc-dev/neo-go/blob/433ffc7b4f608efd16f73a69ab8a6fee7259679c/pkg/core/native/native_nep17.go#L236

So the only difference between this PR and N3 behaviour is that in N3 blocked EOA accounts are still allowed to receive NEO/GAS. In that part I think that your PR behaves more correct than N3.

All in all, I agree with this fix. But we should document this behaviour for the end-users because otherwise it might be unexpected.

txhsl:
> But we should document this behaviour for the end-users because otherwise it might be unexpected.

OK, I will, both here and in the official doc.

Issue #576, opened 2026-03-11
dBFT stops with missing transactions


txhsl:
#### System information

Geth version: `v0.5.2-stable`
OS & Version: Windows/Linux/OSX

#### Expected behaviour

The network should keep alive even there is a missing transaction

#### Actual behaviour

Our Testnet once stopped due to a missing transaction.

#### Steps to reproduce the behaviour

Run a Privnet. This is a possible case without precondition.

#### Backtrace

````
1773222749428	2026-03-11T09:52:29.428Z	2026-03-11T09:52:29.428Z	INFO	received ChangeView	{"validator": 5, "reason": "TxNotFound", "new view": 5}
1773222749428	2026-03-11T09:52:29.428Z	2026-03-11T09:52:29.428Z	DEBUG	received message	{"type": "ChangeView", "from": 5, "height": 7045088, "view": 4, "my_height": 7045088, "my_view": 4}
1773222749239	2026-03-11T09:52:29.239Z	2026-03-11T09:52:29.239Z	DEBUG	received message	{"type": "ChangeView", "from": 5, "height": 7045088, "view": 4, "my_height": 7045088, "my_view": 4}
1773222749239	2026-03-11T09:52:29.239Z	2026-03-11T09:52:29.239Z	INFO	received ChangeView	{"validator": 5, "reason": "TxNotFound", "new view": 5}
1773222749238	2026-03-11T09:52:29.238Z	2026-03-11T09:52:29.238Z	INFO	received ChangeView	{"validator": 5, "reason": "TxNotFound", "new view": 5}
1773222749238	2026-03-11T09:52:29.238Z	2026-03-11T09:52:29.238Z	DEBUG	received message	{"type": "ChangeView", "from": 5, "height": 7045088, "view": 4, "my_height": 7045088, "my_view": 4}
1773222748881	2026-03-11T09:52:28.881Z	2026-03-11T09:52:28.880Z	DEBUG	received message	{"type": "ChangeView", "from": 0, "height": 7045088, "view": 4, "my_height": 7045088, "my_view": 4}
1773222748881	2026-03-11T09:52:28.881Z	2026-03-11T09:52:28.880Z	INFO	received ChangeView	{"validator": 0, "reason": "TxNotFound", "new view": 5}
1773222748880	2026-03-11T09:52:28.880Z	2026-03-11T09:52:28.880Z	INFO	received ChangeView	{"validator": 0, "reason": "TxNotFound", "new view": 5}
[...]

PR #655, opened 2026-09-11
consensus: migrate to extended PrepareRequest format

AnnaShaleva:
Implement https://github.com/nspcc-dev/neo-go/issues/4325.

Solves the problem described in https://github.com/bane-labs/go-ethereum/issues/576 at the consensus level by adding full list of serialized transactions to the PrepareRequest.

@txhsl please, review this PR. We'd like to introduce the same approach in both NeoX and N3 netwroks. For N3 networks we performed several benchmarks and this scheme does not lead to significant network performance degradation. At the same time, it solves the problem of "conflicting" transactions, see the results in https://github.com/nspcc-dev/neo-go/issues/4325#issuecomment-5481545518.

So from your side we need a testing assistance:
1. Setup a similar performance test and compare this PR with master to see if there's a significant TPS/TimePerBlock degradation on NeoX network.
2. Try to reproduce "conflicting" transactions scenario described in https://github.com/bane-labs/go-ethereum/issues/576 and check if this PR solves the problem.

Don't hesitate to share your thoughts/ideas. If everything works good, we'd like to enable this approach on NeoX in the next hardfork.


AnnaShaleva:
I checked some base scenarious on 4-nodes privnet with/without transactions, checked recoveries with a single CN killed, works as expected.

go.mod
```diff
@@ -53,7 +53,7 @@ require (
 	github.com/mattn/go-colorable v0.1.14
 	github.com/mattn/go-isatty v0.0.20
 	github.com/naoina/toml v0.1.2-0.20170918210437-9fafd6967416
-	github.com/nspcc-dev/dbft v0.3.2
+	github.com/nspcc-dev/dbft v0.4.1-0.20260910093827-22729f107405
```
AnnaShaleva, line 56: Note that this dBFT revision is not final, we're still polishing the https://github.com/nspcc-dev/dbft/pull/160.

roman-khimov: Which means it's a draft effectively. But something that can be tested.

consensus/dbft/dbft.go
txhsl, line 645: I'm OK with the idea, but IMO the Blob sidecar introduced by EIP-4844 (Blob transaction) brings some issue in this implementation.

@AnnaShaleva In the dBFT scope, we currently don't handle any Blob data, but depends on a trust of the CL beacon worker. If a Blob transaction is sent to dBFT, or can be fetched by `getTxCb`, then it's guaranteed that the Blob data of this transaction is cached locally, or in another word, exists in the network.

If the transaction is found in EL's txpool, then it's the `blobpool` verifies the blob data; If the transaction is found as a missing tx, then it's the CL beacon cache ensures the blob data.

https://github.com/bane-labs/go-ethereum/blob/f0e236838bb334c7c0d29eeca33533ed0cfda254/beacon/impl/miner/worker.go#L365-L395

This rule is verified parallel beyond dBFT, since it's heavy in computation. So the issue is, if dBFT gets the transations from `PreparePayload`, then the blob existence of EIP-4844 is not promised (we don't send this huge data through dBFT messages). E.g. the primary node can delete the sidecar of a Blob transaction. If there's no other CN knows it, then it gets removed from the network, but dBFT can't detect this behavior. 

And the following Blob sync may block the network, since all CNs can't provide the expected response.

txhsl: I'm OK with the idea, but IMO the Blob sidecar introduced by EIP-4844 (Blob transaction) brings some issue in this implementation.

@AnnaShaleva In the dBFT scope, we currently don't handle any Blob data, but depends on a trust of the CL beacon worker. If a Blob transaction is sent to dBFT, or can be fetched by `getTxCb`, then it's guaranteed that the Blob data of this transaction is cached locally, or in another word, exists in the network.

If the transaction is found in EL's txpool, then it's the `blobpool` verifies the blob data; If the transaction is found as a missing tx, then it's the CL beacon cache ensures the blob data.

https://github.com/bane-labs/go-ethereum/blob/f0e236838bb334c7c0d29eeca33533ed0cfda254/beacon/impl/miner/worker.go#L365-L395.

This rule is verified beyond dBFT. So the issue is, if dBFT gets the transations from `PreparePayload`, then the blob existence of EIP-4844 is not promised (we don't send this huge data through dBFT messages). E.g. the primary node can delete the sidecar of a Blob transaction. If there's no other CN knows it, then it gets removed from the network, but dBFT can't detect this behavior. 

And the following Blob sync may block the network, since all CNs can't provide the expected response.

txhsl: After some evaluation, my idea is to downgrade the current "find missing transactions" logic in `beacon` protocol to "find missing blobs", so that we can both ensure the EIP-4844 verification and remove the need to fetch transaction bodies.

I can open a PR based on this, and any other idea is still welcome.

txhsl:
Rebased to the latest

consensus/dbft/dbft.go
AnnaShaleva: Will check today. I also did some modifications on the dBFT side to be able to properly react on the missing blobs, will update once finished.

txhsl: > I also did some modifications on the dBFT side to be able to properly react on the missing blobs

🤔Then you need a callback function from the Geth node to check how many and which `versionedHash` carried by the blob transactions is missing, maybe it should just be a part of the prepare request verification in the node side?

@AnnaShaleva I've uploaded my idea to https://github.com/bane-labs/go-ethereum/tree/beacon-blobs, although not finalized yet. TBO, it seems to make the thing a step backward and leave the original issue unsolved, because it preserves the risk that some data not carried by prepare request can cause CVs, which could be solved by the full transaction extension. 

Ethereum's solution is to give an epoch window for blob gossips. The miner should broadcast the raw blob data as widely as possible, so that other miners may accept the block into the fork choice, so it's against our 1-block finality idea.

In the worst case, we may need to send the blob bundle along with the prepare request to solve it. Do you have any suggestion?

txhsl: Oh, I notice you're working on https://github.com/bane-labs/go-ethereum/tree/blobs-extension, I'm now looking into it.

AnnaShaleva: > it seems making the thing a step backward and leaving the original issue unsolved

With the approach from https://github.com/bane-labs/go-ethereum/tree/blobs-extension the original issue will be solved for non-blob transactions only, you're right. For blob transactions it's still relevant.

> we may need to send the blob bundle along with the prepare request to solve it.

Bundles are heavy, I think we just can't afford to include them to PrepareRequest because it will significantly reduce the network bandwidth and affect TPS. From the discussion with Roman we think that it's OK for blob transactions to cause the CVs: unlike Ethereum networks, for NeoX blob transactions are rare. So our suggestion is to preserve only simple transactions and let the blobs be requested by CN on demand, if not found locally.

> How about using GetBlobs instead of GetTx

Agree. So from the dbft library side no large changes are required. dBFT still will use `WithRequestTransactions` setting to setup tx request callback, it will be called for blob transactions only, and we'll slightly modify the callback itself to request the blob parts instead of the full transaction.

https://github.com/bane-labs/go-ethereum/tree/beacon-blobs has the right direction, it's just that PrepareRequest verification callback is not the most suitable place to request blobs. We'll combine your branch with changes from https://github.com/bane-labs/go-ethereum/tree/blobs-extension to send a proper blob request in `WithRequestTx` callback based on the versioned hashes from the extended PrepareRequest (it can be retrieved from the dBFT's context).

I'll finalize the dbft library part soon (we want to get rid of `TransactionHashas` context field and convert `Transactions` from map to slice) and push an update with proper `versionedHash` retrieval.

AnnaShaleva: > it's just that PrepareRequest verification callback is not the most suitable place to request blobs

Something like 05c92b21ae37d03cf959176fbb6d65f134fe331a. So dBFT library still marks the whole blob transactions as "missing" if its sidecar is missing. Then dBFT requests blob transaction by its has in `requestTxCb`, so we have to map blob tx hash to the list of `BlobHashes`.

The only thing that's left is to properly handle notification about the incoming (and verified) blob. Will check it.

PR #660, opened 2026-09-21
go.mod: fix keeper dependency

### LINKED ISSUE #659: `cmd/keeper` not working
#### System information

Geth version: `v0.7.0-unstable`
Commit hash : fdf0ad6e2952533003c608e7f20b6a761dff9372

#### Expected behaviour

We expect that the codebase can be built successfully.

#### Actual behaviour

The Geth binary and related tools are all good to build with `make all`, but we missed an independent directory `cmd/keeper`. Its dependencies always has something wrong since we first imported in v0.6.0.

Maybe we also need further TODOs to make it works for Neo X, not only the compilation, let's include them in this issue as well.

qianhh:
Close #659 .


qianhh:
It can be compiled through the following steps:
```sh
cd cmd/keeper
go build -tags "example" ./
```

cmd/keeper/go.mod
```diff
@@ -4,16 +4,19 @@ go 1.24.0
 
 require (
 	github.com/ProjectZKM/Ziren/crates/go-runtime/zkvm_runtime v0.0.0-20251001021608-1fe7b43fc4d6
-	github.com/ethereum/go-ethereum v0.0.0-00010101000000-000000000000
+	github.com/ethereum/go-ethereum v1.17.5
```
txhsl, line 7: Just for future reference, the mock version number works in the original codebase but doesn't work here. Because the `github.com/bane-labs/zk-dkg v0.3.0` requires a `github.com/ethereum/go-ethereum@v1.14.12`.

txhsl (approved):

